Issue #799, opened 2021-04-09
Redesign: Instances of %CLIENT_MAJOR_VERSION% in the text


anoadragon453:
There are instances of the `%CLIENT_MAJOR_VERSION%` placeholder appearing in the spec text, as seen [here](https://spec.matrix.org/unstable/application-service-api/#server-admin-style-permissions). I believe in the old spec this was replaced with the CS API major version number.

I'm not sure how this interacts with the new Matrix Spec Versioning stuff, but we probably want to write a Hugo template for this.

KitsuneRal:
With the transition to individual endpoint versions being done, should this issue be closed?

anoadragon453:
I no longer see this text anywhere in the rendered spec, so I'd say so!

aaronraimist:
@anoadragon453 I still see it on https://spec.matrix.org/unstable/application-service-api/#server-admin-style-permissions

PR #980, opened 2022-03-03
Switch `%CLIENT_MAJOR_VERSION%` in /login endpoint for current /login version

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/799

This was left over from the days of all CS API endpoints sharing the same major spec release version. `/login` now has its own endpoint version.

Preview: https://pr980--matrix-spec-previews.netlify.app


richvdh (approved):
thanks!

PR #987, opened 2022-03-08
Clarify that m.room.avatar url can be null

### LINKED ISSUE #471: Need to clarify what empty / unset state events look like
State events like `m.room.avatar` have required fields, however they are technically optional because removing the property does things. Or are clients expected to handle `"url": ""` (empty strings) as "no url" instead?

jplatte:
Might close #471 (that issue refers to "State events like `m.room.avatar`" but gives no other examples).


zecakeh:
I agree with this clarification but it should be noted that for now Element Web (and maybe others) actually omit the field.

turt2live:
This opens a bit of a pandora's box on what "required" means. We also say that `body` is required for room messages, but if a message is redacted then it doesn't exist. Likewise, we don't actually say what a client should do if it receives a `body`-less message.

I think for this change to be accepted we'd need to figure out if this is actually what we want, and adjust the associated wording too. I've left this comment as a not-review to hopefully get other SCT members to take a look and form an opinion.

jplatte:
IMHO redacted events are a red herring: They are already clearly marked as not the regular form of the event by having the `redacted_because` field which you can check for before looking at the event's content.

In multiple years of maintaining Ruma, this is the one and only event I've come across where the `Required` attribute is set but non-redacted events without the field are still frequently encountered by users. That's not to say there aren't any others where this could happen, but I don't see how this would be opening a pandora's box: I know of no other events I would care to apply the same kind of change to right now.

richvdh (commented):
I think I need more context here. If this is unrelated to redactions (which are indeed a separate case), can someone explain to me why clients are sending `m.room.avatar` without a url property? It seems like it wouldn't be much use.

jplatte:
That's just what happens when you "remove" the avatar in Element Web and other clients. It sends a new `m.room.avatar` event without a `url` field to have the latest avatar be 'none' without actually redacting the information which avatar was used up until that point (which setting a new avatar would also not do).

data/event-schemas/schema/m.room.avatar.yaml
```diff
@@ -12,8 +12,6 @@ properties:
       url:
         description: The URL to the image.
```
richvdh, line 13: It would be good to add some text here about what it means if `url` is omitted.

changelogs/client_server/987.clarification
```diff
@@ -0,0 +1 @@
+Clarify that the url field in `m.room.avatar` events is not required.
```
richvdh, line 1: this is in the wrong directory.

richvdh (changes requested):
Right, I see. I think there's an argument that that would be better handled by setting `url` to the empty string, but otoh if the ecosystem is already doing it by removing `url`, that horse has bolted.

Just needs some clarifying text, imho.

changelogs/client_server/987.clarification
richvdh: (if you merge latest `main` you'll get a CI check that checks this)

data/event-schemas/schema/m.room.avatar.yaml
```diff
@@ -10,10 +10,11 @@ properties:
           - $ref: core-event-schema/msgtype_infos/image_info.yaml
         description: Metadata about the image referred to in `url`.
       url:
-        description: The URL to the image.
+        description: |-
+          The URL to the image.
+
+          If this isn't set, that means a previous room avatar has been unset.
```
richvdh, line 16:
```suggestion
          The URL to the image. If this property is not present, the room has no avatar. This can be useful
          to remove a previous room avatar.
```

richvdh (approved):
this lgtm but I'm going to hold off from merging it for now in case any of the rest of @matrix-org/spec-core-team want to veto it

richvdh:
(and yes, my newsfragment CI check is broken. Fix in #993)

turt2live:
(please do not rebase after review has been given - instead prefer merge commits. Those merge commits will disappear when we merge the PR at the end)

jplatte:
Okay. Just noticed I force-pushed away the previous suggestion too. Will re-apply that.

KitsuneRal (approved):

jplatte:
> this lgtm but I'm going to hold off from merging it for now in case any of the rest of @matrix-org/spec-core-team want to veto it

How long is the veto timeframe? Can this be merged now?

ShadowJonathan:
> Might close https://github.com/matrix-org/matrix-spec/issues/471

this did close that issue due to GitHub’s PR keywords, should that issue be closed in the end?

richvdh:
think so

PR #990, opened 2022-03-08
GHA job to check newsfragments.

richvdh:
Having just found another *three* newsfragments in the wrong places, I cracked and wrote a CI job to check.

Ironically, this PR is failing CI, because it doesn't have a changelog. That doesn't happen very often, so I'm inclined to ignore it, but I'm open to suggestions for ways to flag/detect PRs that don't need a changelog.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,6 +41,18 @@ jobs:
         run: |
           /env/bin/python scripts/check-event-schema-examples.py
 
+  check-newsfragments:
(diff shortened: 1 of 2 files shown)


turt2live (approved):
thank you! A possible way of flagging internal changes is to set up a changelog directory that we clear out on release, but doesn't end up in the actual changelog. Alternatively, a label.

richvdh:
yeah I was thinking a label or somesuch. Not wild about having to make a newsfile that's going to get thrown away.

PR #992, opened 2022-03-08
Suppress scroll-anchoring in the sidebar

### LINKED ISSUE #965: Links to specific spec items (with URL fragments) "miss" their target when firefox navigates directly to them
* Open a new tab in firefox and navigate to https://spec.matrix.org/v1.1/client-server-api/#fallback-for-mtext-mnotice-and-unrecognised-message-types
* The browser loads the page and scrolls to the right place in the viewport. Some font faces are not loaded.
* After some time, fonts fail to load and we fall back to some default. (I noticed hundreds of content security policy warnings in the javascript console about not being able to load a remote resource from an inline style. What the heck are we trying to do?)
* Doing so recalculates layout. The main content of the page shifts. I am no longer looking at the anchor that I wanted to link to.
* Mild irritation, apply aloe vera and take deep breaths.

I make two conjectures:

1. I reckon that if we used a font that was available on the system, we wouldn't have this two-step loading which scrolls the viewport away from the intended destination fragment. See e.g. https://systemfontstack.com/
2. https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring refers to Firefox's "scroll anchoring" behaviour. Turning on the debug `layout.css.scroll-anchoring.highlight` tool option I see that the scroll anchor is the table of contents. Not sure why this happens; I guess firefox has some heuristic here. We can opt out of this behaviour with the css `.td-sidebar { overflow-anchor: none; }`. I reckon that doing this would also help avoid the jumpy behaviour.

richvdh:
Hopefully, fixes #965. Props to @DMRobertson for debugging this and pointing out the fix.

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -41,6 +41,11 @@ Custom SCSS for the Matrix spec
 }
 
 /* Styles for the sidebar nav */
+.td-sidebar {
+    /* don't attempt to use the sidebar (or things in it) as a scroll anchor. */
+    overflow-anchor: none;
+}
+
 .td-sidebar-nav {
   scroll-behavior: smooth;
   overscroll-behavior: contain;


richvdh:
look: a working link: https://pr992--matrix-spec-previews.netlify.app/client-server-api/#device-verification

turt2live (approved):
cryptic css is my favourite css

DMRobertson:
> look: a working link: https://pr992--matrix-spec-previews.netlify.app/client-server-api/#device-verification

Doesn't always work for me, but it seems to work more often. I think there's still some kind of race against a [FOIT](https://css-tricks.com/fout-foit-foft/) which breaks the linkage.

anoadragon453:
Also didn't work for me. Took me to the "Key verification framework" header. Firefox 98.0, Arch Linux.

richvdh:
LE GRAND SIGH

PR #997, opened 2022-03-16
Add missing object type to openapi client event

nico-famedly:
Signed-off-by: Nicolas Werner <n.werner@famedly.com>

Not sure if this needs a news fragment, since this is fairly trivial and just changes one line in the openapi stuff. This is also how it is done everywhere else in the openapi spec, so I assume it just was missed.


KitsuneRal (approved):
I don't think it needs a newsfragment either - there's no tangible difference even for any (well-behaved) code-generation tool, let alone readers of the spec.

Issue #374, opened 2018-09-06
Make transaction.origin and pdu.origin optional/go away


turt2live:

richvdh:
It seems that there already exist some implementations which do not include an `origin` on their outgoing PDUs, and Synapse does not insist on its presence on incoming events. I'm therefore leaning to the idea that the presence of the field in the spec should be considered a spec bug.

richvdh:
Possibly, but if there are any implementations that omit `transaction.origin`, there's more room to argue that such an implementation is buggy (since transactions aren't persisted to a room DAG)

jplatte:
Re.

> the body of a `send_join` request

I'm seeing an origin field in other places:

- The response fields of `GET /make_knock/{roomId}/{userId}`
- The response fields of `GET /make_leave/{roomId}/{userId}`
- The `Room State` object (used in the response of `PUT /send_join/{roomId}/{eventId}`)
- The `InviteEvent` object (used in request and response of `PUT /invite/{roomId}/{eventId}`)

Are some of these also redundant?

richvdh:
> Possibly, but if there are any implementations that omit `transaction.origin`, there's more room to argue that such an implementation is buggy (since transactions aren't persisted to a room DAG)

I'm having a hard time deciphering this. I *think* what I meant was: any implementation that does not send `transaction.origin` is buggy, and:

The difference between `transaction.origin` and `pdu.origin` is that the ecosystem already has to deal with missing `pdu.origin`. 

This is because Synapse does not enforce the presence of `pdu.origin`, and will therefore happily relay events that have no `origin` to other servers. In other words: if, when you receive a pdu, you insist on `origin`, then you cut yourself off from whole rooms; if you insist on `transaction.origin`, you only cut yourself off from the one server that is failing to send them.

richvdh:
> Re.
> 
> > the body of a `send_join` request
> 
> I'm seeing an origin field in other places:
> 
> * The response fields of `GET /make_knock/{roomId}/{userId}`
> 
> * The response fields of `GET /make_leave/{roomId}/{userId}`
> 
> * The `Room State` object (used in the response of `PUT /send_join/{roomId}/{eventId}`)
> 
> * The `InviteEvent` object (used in request and response of `PUT /invite/{roomId}/{eventId}`)
> 
> Are some of these also redundant?

Links to the relevant parts of the spec would be useful, but off the top of my head: they sound like they are just special cases of `pdu.origin`.

[...]

PR #998, opened 2022-03-17
Remove origin from unsigned_pdu_base.yaml

jplatte:
It doesn't serve a useful purpose, is not enforced to be present by Synapse and already being omitted by at least one homeserver implementation.

Fixes one half of #374.

Preview: https://pr998--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/server-server/definitions/unsigned_pdu_base.yaml
+++ b/data/api/server-server/definitions/unsigned_pdu_base.yaml
@@ -124,7 +124,6 @@ required:
   - event_id
   - room_id
   - sender
-  - origin
   - origin_server_ts
   - type
   - content


data/api/server-server/definitions/unsigned_pdu_base.yaml
```diff
@@ -25,10 +25,6 @@ properties:
     type: string
     description: The ID of the user sending the event.
     example: "@someone:matrix.org"
-  origin:
```
richvdh, line 28: you'll need to remove `origin` from the `required` list below too.

richvdh:
LGTM otherwise, but I'd like to check if anyone else in the SCT has an opinion.

PR #1002, opened 2022-03-25
Mark `from` parameter as optional for `/messages`

jplatte:
Continuation of https://github.com/matrix-org/matrix-spec-proposals/pull/3638 on the new spec repo.

Preview: https://pr1002--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/message_pagination.yaml
+++ b/data/api/client-server/message_pagination.yaml
@@ -57,7 +57,11 @@ paths:
             This endpoint can also accept a value returned as a `start` token
             by a previous request to this endpoint, though servers are not
             required to support this. Clients should not rely on the behaviour.
-          required: true
+
+            If it is not provided, the homeserver shall return a list of messages 
+            from the first or last (per the value of the `dir` parameter) visible
+            event in the room history for the requesting user.
+          required: false
           x-example: "s345_678_333"
         - in: query
           type: string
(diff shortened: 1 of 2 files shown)


data/api/client-server/message_pagination.yaml
```diff
@@ -49,6 +49,7 @@ paths:
         - in: query
           type: string
           name: from
+          x-changedInMatrixVersion: "1.3"
```
richvdh, line 52: this doesn't seem to be showing anything in the rendered spec.

Generally we need to describe *how* it changed, not just that it changed.

jplatte: So just prepend `**Since v1.3:**` to the new paragraph in the description, or something else?

richvdh: have a look at `data/api/client-server/definitions/client_event_without_room_id.yaml` for an example of how to use `x-changedInMatrixVersion` correctly

richvdh: hrm, it still doesn't seem to be rendering. We'll have to dig into what's wrong with the templates :/

richvdh: #1011 should fix this.

richvdh (approved):
lgtm

PR #1008, opened 2022-03-29
Fix syntax highlighting in rendered spec

richvdh:
Hugo's syntax highlighter currently generates inline `style`s; however thsed cause CSP errors and don't work. Instead, let's use proper CSS classes.


anoadragon453 (approved):
lgtm! tested and works locally

unrelated: I am thoroughly pleased that https://xyproto.github.io/splash/docs/rainbow_dash.html exists.

PR #1036, opened 2022-04-26
Get rid of the `proposal-in-review` label

richvdh:
Everything is in review. We may as well just use the draft state for WIP stuff.


turt2live (approved):
I think this is correct enough

:shipit:

PR #1037, opened 2022-04-29
Stateres v2: Use bold to introduce definitions.

DMRobertson:
Otherwise they are inline into the paragraph without punctuation, and
hard to read.

I'm guessing that pandoc did a poor job of converting a LaTeX
`begin{description}...\end{description}` or `\paragraph{...} ...` to
Markdown.

Preview: https://pr1037--matrix-spec-previews.netlify.app


DMRobertson:
Before:
Screenshot from 2022-04-29 15-42-48

After: 
Screenshot from 2022-04-29 15-42-37

turt2live (approved):
thanks!

PR #1038, opened 2022-04-29
Fix origin server name in S2S Request Authentication example.

dkasak:
The JSON payload listed a different origin HS name from the one mentioned in the Authorization header.

Preview: https://pr1038--matrix-spec-previews.netlify.app


turt2live (approved):
thanks!

Issue #774, opened 2021-03-04
prime characters (') are almost invisible


richvdh:
**Link to problem area**:

https://spec.matrix.org/unstable/rooms/v2/#state-resolution

**Issue**

The distinction between `S` and `S'` is absolutely key to this text, and it's almost impossible to distinguish.

![image](https://user-images.githubusercontent.com/1389908/109958528-380ea580-7cde-11eb-8f4d-6a72971467a8.png)

DMRobertson:
(One day, we'll have [MathML everywhere](https://caniuse.com/mathml) and we can trust browsers to lay things out sensibly.)

DMRobertson:
I tried including the prime in the italics. That looks better, but not great in Firefox:

```
sed -e "s/\*S\*′(/*S′*(/g" -i content/rooms/{fragments/v2-state-res.md,v1.md}
```
![Screenshot from 2022-05-03 13-24-25](https://user-images.githubusercontent.com/8614563/166452475-aa7f6883-1ccf-46ba-8893-ca9c4dc7720d.png)

PR #1040, opened 2022-04-29
Use A(E) and B(E) for state after/before events

DMRobertson:
A suggestion to mitigate #774.

Preview: https://pr1040--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1040.clarification
+++ b/changelogs/room_versions/newsfragments/1040.clarification
@@ -1,0 +1,1 @@
+Adjust state resolution notation to be easier to read in a browser.
--- a/content/rooms/fragments/v2-state-res.md
+++ b/content/rooms/fragments/v2-state-res.md
@@ -2,17 +2,17 @@
 toc_hide: true
 ---
 
-The room state *S*′(*E*) after an event *E* is defined in terms of the
-room state *S*(*E*) before *E*, and depends on whether *E* is a state
+The room state *A*(*E*) after an event *E* is defined in terms of the
+room state *B*(*E*) before *E*, and depends on whether *E* is a state
 event or a message event:
 
--   If *E* is a message event, then *S*′(*E*) = *S*(*E*).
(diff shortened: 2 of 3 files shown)


DMRobertson:
Created with

```shell
sed -e 's/\*S\*(/*B*(/g' -e "s/\*S\*′(/*A*(/g" -i content/rooms/{fragments/v2-state-res.md,v1.md}
```

uhoreg:
I like the idea in general, but I'm not sure about using "A" and "B", since "A" comes before "B" in the alphabet, so, at least for me, my brain gets confused about which one represents the earlier state, and which one is the later state.

DMRobertson:
Could try `I(E)` and`F(E)` for initial and final?

Could also try including the prime in the `<em>`. I can experiment a little next week.

DMRobertson:
Could also use the notation *P*(*E*) to mean "the state prior to *E*".

DMRobertson:
> Could also try including the prime in the `<em>`. I can experiment a little next week.

See #1043.

uhoreg:
"I" and "F" for "Initial" and "Final" would probably work.  It also suffers from the unfortunate fact that "I" comes after "F" in the alphabet, but they're far enough apart (and not at the beginning of the alphabet) that it doesn't trigger my "I don't know what order it's supposed to be" response.

turt2live:
@uhoreg as the sole SCT commenter on here, does https://github.com/matrix-org/matrix-spec/pull/1043 alleviate the concern enough or is there still work to be done here?

uhoreg:
While #1043 makes the primes more visible, primes can be somewhat foreign to many people, so I think there's still value in doing something along the lines of what this PR is trying to do.  On the other hand, this PR now has conflicts with the main branch (presumably because of 1043), so it may be easier to create a new PR, rather than trying to fix this one up.

I'd be inclined to let @DMRobertson decide how to proceed with this particular PR.

DMRobertson:
No strong opinions. #1043 makes things readable enough for me. If there's no obvious better notation, let's leave as-is.

PR #1043, opened 2022-05-03
Italicise primes and brackets in state res desc

### LINKED ISSUE #774: prime characters (') are almost invisible

DMRobertson:
I don't like it but it seems to make things look decent in Firefox.
This is probably good enough---we're not Knuth, after all.

To the reviewer: please check that this looks like an improvement for a Chromium-based browser too.

Additionally I have avoided the use of `*E*'s` as in "E's prev_events"
since the apostrophe looks poor here.

Resolves #774.
An alternative to #1040.

Preview: https://pr1043--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1043.clarification
+++ b/changelogs/room_versions/newsfragments/1043.clarification
@@ -1,0 +1,1 @@
+Adjust mathematical notation used in the description of state resolution to render better in browsers.
--- a/content/rooms/fragments/v2-state-res.md
+++ b/content/rooms/fragments/v2-state-res.md
@@ -2,20 +2,19 @@
 toc_hide: true
 ---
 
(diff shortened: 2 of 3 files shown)


DMRobertson:
Firefox before:

Screenshot from 2022-05-03 14-09-44


Firefox after:

Screenshot from 2022-05-03 14-09-03

turt2live (approved):
This looks closer to how it was probably intended, but a RST->MD transition long ago made Bad Decisions for us. Approving this while we determine if the alternative naming is better for the long run.

PR #1045, opened 2022-05-04
Attempt to define a valid event

### LINKED ISSUE #1044: Server-server API does not describe what it means for a PDU to be valid
> Whenever a server receives an event from a remote server, the receiving server must ensure that the event:
>
>   1.  Is a valid event, otherwise it is dropped.
>   2. ...

**Link to problem area**: https://spec.matrix.org/unstable/server-server-api/#checks-performed-on-receipt-of-a-pdu

**Issue**
Presumably "valid" means
- contains a `room_id` key
- the format of the event complies with the Event PDU Format in that room's room version.

DMRobertson:
Resolves #1044.

Well, maybe. See my comments on that issue for context.

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -355,7 +355,9 @@ specification](/rooms).
 Whenever a server receives an event from a remote server, the receiving
 server must ensure that the event:
 
-1.  Is a valid event, otherwise it is dropped.
+1.  Is a valid event, otherwise it is dropped. For an event to be valid, it
+    must contain a `room_id`, and it must comply with the event format of
+    that room version.
 2.  Passes signature checks, otherwise it is dropped.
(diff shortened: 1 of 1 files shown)


content/server-server-api.md
```diff
@@ -355,7 +355,9 @@ specification](/rooms).
 Whenever a server receives an event from a remote server, the receiving
 server must ensure that the event:
 
-1.  Is a valid event, otherwise it is dropped.
+1.  Is a valid event, otherwise it is dropped. For an event to be valid, it
+    must contain a `room_id`, and it must comply with the event format of
+    that room version.
```
turt2live, line 360: more links is more good, I think:
```suggestion
    that [room version](/rooms).
```

PR #1054, opened 2022-05-10
Drop lifetime from the call answer event example

### LINKED ISSUE #931: m.call.answer sample has an invalid field.
**Link to problem area**:
See the `lifetime` field here,
https://github.com/matrix-org/matrix-doc/blob/8d9a67d9f64c51be3e8465d7db1f1de5f47faa7a/data/event-schemas/examples/m.call.answer.yaml#L4-L12
it's not actually in the schema.
https://github.com/matrix-org/matrix-doc/blob/8d9a67d9f64c51be3e8465d7db1f1de5f47faa7a/data/event-schemas/schema/m.call.answer.yaml#L8-L38

**Issue**

Correct the sample or the schema.

KitsuneRal:
Signed-off: Alexey Rusakov <Kitsune-Ral@users.sf.net>

Closes #931.

### DIFF
--- a/changelogs/client_server/newsfragments/1054.clarification
+++ b/changelogs/client_server/newsfragments/1054.clarification
@@ -1,0 +1,1 @@
+Fix various typos throughout the specification.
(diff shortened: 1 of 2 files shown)


turt2live (approved):
manually merging because the CI is complaining about a different changelog entry (already fixed)

PR #1056, opened 2022-05-10
Add spec for refresh tokens

uhoreg:
spec PR for [MSC2918](https://github.com/matrix-org/matrix-spec-proposals/pull/2918)


turt2live:
(pull latest to fix CI - changelog complaint isn't with your changelog but rather a commit on `main`)

content/client-server-api/_index.md
```diff
@@ -336,42 +337,92 @@ to prevent the access token being leaked in access/HTTP logs. The query
...
+homeserver; clients indicate their support for refresh tokens by
+including a `refresh_token: true` property in the request body of the
+`/login` and `/register` endpoints. For example, homeservers may allow
+the use of non-expiring access tokens, or may expire access tokens
+anyways and rely on soft logout behaviour on clients that don't
+support refreshing.
 
 ### Soft logout
 
-When a request fails due to a 401 status code per above, the server can
```
richvdh, line 360: this section probably needs `{{% changed-in v="1.3" %}}` ?

content/client-server-api/_index.md
```diff
@@ -336,42 +337,92 @@ to prevent the access token being leaked in access/HTTP logs. The query
...
-`M_UNKNOWN_TOKEN` respectively.
+{{% changed-in v="1.3" %}} When credentials are required but missing or
+invalid, the HTTP call will return with a status of 401 and the error code,
+`M_MISSING_TOKEN` or `M_UNKNOWN_TOKEN` respectively.  Note that an error code
+of `M_UNKNOWN_TOKEN` could mean one of four things:
+
+1. the access token was never valid.
+2. the access token has been logged out.
+3. the access token has been [soft logged out](#soft-logout).
+4. the access token [needs to be refreshed](#refreshing-access-tokens).
```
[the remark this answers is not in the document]
uhoreg: Yes, I originally tried that but it added a line break.  But apparently using angle brackets instead of '%' fixes that.

data/api/client-server/refresh.yaml
```diff
@@ -0,0 +1,107 @@
...
+    post:
+      x-addedInMatrixVersion: "1.3"
+      summary: Refresh an access token
+      description: |-
+        Refresh an access token. Clients should use the returned access token
+        when making subsequent API calls, and store the returned refresh token
+        (if given) in order to refresh the new access token when necessary.
+
+        After an access token has been refreshed, a server can choose to
+        invalidate the old access token immediately, or can choose not to, for
```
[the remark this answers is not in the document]
uhoreg: I don't think that matters, since the client will just get an `UNKNOWN_TOKEN` and re-refresh the access token, which is fine.  However the server needs to keep the refresh token active until the client uses the new access token.  Which was in the MSC, but I forgot to include it.

content/client-server-api/_index.md
```diff
@@ -336,42 +337,92 @@ to prevent the access token being leaked in access/HTTP logs. The query
...
+must be discarded, and can be re-used if the client successfully
+re-authenticates.
 
-When `soft_logout` is true, the client can acquire a new access token by
-specifying the device ID it is already using to the login API. In most
-cases a `soft_logout: true` response indicates that the user's session
-has expired on the server-side and the user simply needs to provide
-their credentials again.
+The server indicates that the client is in a soft logout state by
+including a `soft_logout: true` parameter in an `M_UNKNOWN_TOKEN`
```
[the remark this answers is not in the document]
uhoreg: https://spec.matrix.org/unstable/client-server-api/#common-error-codes only defines the `soft_logout` parameter under `M_UNKNOWN_TOKEN`, so it should be OK.  And I kind of want us to prefer using the error codes in the spec rather than the HTTP status codes, since that's what we say clients should prefer.

content/client-server-api/_index.md
uhoreg: I've now added the `changed-in` below, in the paragraph that documents the changed behaviour.

data/api/client-server/refresh.yaml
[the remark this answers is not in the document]
uhoreg: ... which is now added

data/api/client-server/login.yaml
```diff
@@ -163,6 +171,23 @@ paths:
...
+                  obtain a new access token when it expires by calling the
+                  `/refresh` endpoint.
+                x-addedInMatrixVersion: "1.3"
+              expires_in_ms:
+                type: integer
+                description: |-
+                  The lifetime of the access token, in milliseconds. Once
+                  the access token has expired a new access token can be
+                  obtained by using the provided refresh token. If no
+                  refresh token is provided, the client will need re-log in
```
richvdh, line 187:
```suggestion
                  refresh token is provided, the client will need to re-log in
```

data/api/client-server/refresh.yaml
```diff
@@ -0,0 +1,110 @@
...
+schemes:
+  - https
+  - http
+basePath: /_matrix/client/v3
+consumes:
+  - application/json
+produces:
+  - application/json
+securityDefinitions:
+  $ref: definitions/security.yaml
```
richvdh, line 28: I believe this is not needed, given it doesn't take an access token.

data/api/client-server/registration.yaml
```diff
@@ -152,6 +158,27 @@ paths:
...
+
+                  Omitted if the `inhibit_login` option is false.
+                x-addedInMatrixVersion: "1.3"
+              expires_in_ms:
+                type: integer
+                description: |-
+                  The lifetime of the access token, in milliseconds. Once
+                  the access token has expired a new access token can be
+                  obtained by using the provided refresh token. If no
+                  refresh token is provided, the client will need re-log in
```
richvdh, line 176:
```suggestion
                  refresh token is provided, the client will need to re-log in
```

PR #1062, opened 2022-05-14
Spec `/relations` and aggregations

turt2live:
* [x] [MSC3676](https://github.com/matrix-org/matrix-spec-proposals/pull/3676) ([MD](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3676-transitioning-away-from-reply-fallbacks.md)) - Reply fallbacks relaxation
* [x] [MSC3666](https://github.com/matrix-org/matrix-spec-proposals/pull/3666) ([MD](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3666-bundled-aggregations-for-search.md)) - Bundling/aggregations for search
* [x] [MSC2674](https://github.com/matrix-org/matrix-spec-proposals/pull/2674) ([MD](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2674-event-relationships.md)) - Relationships
* [x] [MSC2675](https://github.com/matrix-org/matrix-spec-proposals/pull/2675) ([MD](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2675-aggregations-server.md)) - Server-side aggregation

Note: Originally this branch was intended to also carry [MSC3440](https://github.com/matrix-org/matrix-spec-proposals/pull/3440) ([MD](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3440-threading-via-relations.md); Threads), however it was determined that in order to include threads we'll need a bit more of the feature to land first. There shouldn't be any lingering references to threads (at least in a "threads are specced" capacity) - please point them out if they are still there.

**Reviewer**: The diff is a bit messy and the commit history isn't much better, sorry. To summarize what has happened: the rich replies stuff got broken out to a dedicated module (in preparation for Threads getting a dedicated module itself) and the relationships stuff got airlifted to a higher part of the spec. Then, edits to all the sections to make the text clear/match the MSCs.

Preview: https://pr1062--matrix-spec-previews.netlify.app


content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,47 @@ the topic to be removed from the room.
...
+
+`m.relates_to` is described as follows:
+
+{{% definition path="api/client-server/definitions/m.relates_to" %}}
+
+### Relationship types
+
+This specification describes the following relationship types:
+
+* [Rich replies](#rich-replies)
```
clokep, line 1879: It seems a little weird to include this here as rich replies use a different format:

```json
{
  "m.relates_to": {
    "m.in_reply_to": {
      "event_id": "$another_event"
    }
}
```

vs.

```json
{
  "m.relates_to": {
    "rel_type": "whatever",
    "event_id": "$another_event"
  }
}
```

turt2live: noted, however this is carryover from older versions of the spec. Will play with wording to try and clarify what's going on.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,198 @@ the topic to be removed from the room.
...
+Future versions of this specification are expected to require certain behaviours
+or aggregation of related events.
+{{% /boxes/note %}}
+
+#### Aggregations
+
+{{% added-in v="1.3" %}}
+
+Some relationships are "aggregated" by the server depending on their relationship
+type. This can allow a set of related events to be summarised as a subset of values.
```
clokep, line 1908: I think this is accurate, but hard to follow, maybe this is clearer:

```suggestion
Some events are "aggregated" by the server depending on their relationships.
This can allow a set of related events to be summarised.
```

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,198 @@ the topic to be removed from the room.
...
+time of handling, the client should continue to aggregate locally if it is aware of
+the relationship type's behaviour. For example, a client might increment a `count`
+on a related event's bundle if it saw a new event which referenced that event.
+
+{{% boxes/warning %}}
+The bundle provided by the server only includes events which were known at the time
+the event was *received*. This can mean that in a single `/sync` response an event
+will have a bundle and more events which qualify for that aggregation: in this case,
+the client *should* aggregate the events which are "after" the event in question on
+its own, as the server will not have considered them.
```
clokep, line 1993: I don't think this is correct -- the events available in that response would all be bundled into the aggregation, it is subsequent syncs which need to be considered.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,198 @@ the topic to be removed from the room.
...
+{{% /boxes/warning %}}
+
+{{% boxes/note %}}
+Events from [ignored users](#ignoring-users) do not appear in the bundle or aggregation.
+Clients will need to de-aggregate the events sent by ignored users to avoid them being
+considered in counts. Similarly, servers must ensure they do not consider events from
```
clokep, line 1999: I'm not sure what this is saying clients need to do -- the server should be handling the ignored users in bundles it provides?

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,198 @@ the topic to be removed from the room.
...
+ignored users when preparing a bundle for the client.
+{{% /boxes/note %}}
+
+When an event is redacted, the relations attached to it remain. However, when an event
+which uses a relation is redacted then the relation is broken. Thus, the server needs
+to de-aggregate or disassociate an event from its parent when it is redacted. Clients
+with local aggregation should do the same.
```
clokep, line 2006: What does "relations attached to it" mean? I think it means that if you're redacting the parent event it does not affect relations targeting that event.

As opposed to if you redact an event which has an `m.relates_to`.

(While working on the relations code it has become apparent it is very confusing to say "an event with a relation" or "related event" or similar things -- I think it would really help clarify things if the spec were to use defined terms of "parent" and "child" or "target" and "source" or some other terms that are they consistently used throughout.)

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,198 @@ the topic to be removed from the room.
...
+
+Additionally, if the server is missing portions of the room history then it may not be
+able to accurately aggregate the events.
+{{% /boxes/warning %}}
+
+#### API
+
+To retrieve relations for an event from the server, the client can call the following
+endpoint with relevant information. The endpoint does not aggregate the events and is
+instead paginated: clients can perform local aggregation if needed.
```
clokep, line 2029: Maybe "retrieve related events for an event"? (Is a "relation" the information in the `m.relates_to` field or the related event itself?)

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,165 @@
...
+                type: string
+                description: |-
+                  An opaque string representing a pagination token. The absence of this token
+                  means there are no prior results to fetch, i.e. this is the first batch.
+        404:
+          description: The supposed parent event was not found or you do not have permission to read this event.
+          examples:
+            application/json: {
+                "errcode": "M_NOT_FOUND",
+                "error": "Event not found."
```
clokep, line 160: Synapse seems to return "Unknown parent event.", but I think these are just examples anyway so doesn't really matter. 😄

turt2live: Error text is freeform :)

content/client-server-api/_index.md
clokep: Right, so this applies to as they're updating their bundle with new events or something? But I think they wouldn't even get events from ignored users, so not really sure when this would apply?

(Ignoring a user would annoyingly invalidate many bundles, probably.)

clokep: OK, that makes sense! :)

content/client-server-api/_index.md
turt2live: have changed the dynamic a bit to make it more clear, using the parent/child terminology to clear things up - thanks for the suggestion :)

content/client-server-api/_index.md
turt2live: I think this is clarified with parent/child terminology.

content/client-server-api/_index.md
turt2live: indeed - the MSC was a bit confusing on this. The section title mentioned client-specific behaviour, but once in the weeds it sounded like the server was the one limited by time of receipt.

From the MSC:

> Bundled aggregations on an event give a snapshot of what relations were known at the time the event was received.

Clarified in https://github.com/matrix-org/matrix-spec-proposals/pull/3826

turt2live:
@clokep github isn't letting me re-request review from you for some reason, but if you get a chance then a second look would be appreciated (non-blocking, though).

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+  - application/json
+securityDefinitions:
+  $ref: definitions/security.yaml
+paths:
+  "/rooms/{roomId}/relations/{eventId}/{relType}/{eventType}":
+    get:
+      summary: Get the cgukd events for a given parent event, with optional filter.
+      description: |-
+        Retrieve all of the chikd events for a given parent event, optionally filtering
+        down by relationship type and related event type.
```
clokep, line 35:
```suggestion
      summary: Get the child events for a given parent event, with optional filter.
      description: |-
        Retrieve all of the child events for a given parent event, optionally filtering
        down by relationship type and related event type.
```

clokep (approved):
Seems reasonable to me, minus the couple typos! 👍

data/api/client-server/relations.yaml
turt2live: I'm surprised the spellchecker isn't picking these up 😭

clokep: Maybe you got them so wrong it gave up? 😄

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+          type: integer
+          name: limit
+          description: |-
+            The maximum number of results to return in a single `chunk`. The server can
+            and should apply a maximum value to this parameter to avoid large responses.
+
+            Similarly, the server should apply a default value when not supplied.
+          required: false
+          x-example: 20
+      responses:
```
revidee, line 114: The [synapse implementation](https://github.com/matrix-org/synapse/blob/248046187940372c23466cb395b46ed97ebda1ed/synapse/handlers/relations.py#L152) does also include a field "original_event" containing the event without any aggregations. IMHO this should be included in the spec.

turt2live: Without an MSC, it can't be included here.

richvdh:
Is there a reason this needs to lump together the changes to rich replies and all the relations stuff? it makes review rather harder :/.

turt2live:
Yes: most of this was already written before threads got descoped, and is work we'll need to do for threads anyways. 

The commit history isn't great, but is somewhat reviewable.

data/api/client-server/relations.yaml
revidee: > About that field, [MSC2675](https://github.com/matrix-org/matrix-spec-proposals/pull/2675) defers to [MSC2676](https://github.com/matrix-org/matrix-spec-proposals/pull/2676) where it is not mentionned (anymore?).

In that case, either this or [MSC2675](https://github.com/matrix-org/matrix-spec-proposals/pull/2675) should be changed accordingly, either adding a note for the `original_event` and it's typing, or removing it entirely.

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+          x-example: "m.room.message"
+        - in: query
+          type: string
+          name: from
+          description: |-
+            The pagination token to start returning results from. If not supplied, results
+            start at the earliest topological event known to the server.
+
+            Can be a `next_batch` or `prev_batch` token from a previous call, or an equivalent
+            token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
```
richvdh, line 88: The [spec](https://spec.matrix.org/unstable/client-server-api/#get_matrixclientv3roomsroomidmessages) for `/messages` has this to say about its `from` param:

> This endpoint [sic] can also accept a value returned as a `start` token by a previous request to this endpoint, though servers are not required to support this. Clients should not rely on the behaviour.

... which seems to say that  a `start` from a previous `/messages` response is a thing that we're doing our best to deprecate, rather than allow to propagate?

In any case, I think it would be good to be explicit about what constitutes an "equivalent token".

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+        - in: query
+          type: string
+          name: to
+          description: |-
+            The pagination token to stop returning results at. If not supplied, results
+            continue up to `limit` or until there are no more events.
+
+            Like `from`, this can be a previous token from a prior call to this endpoint
+            or from `/messages` or `/sync` to limit to a section of timeline.
+          required: false
```
richvdh, line 102: the MSC doesn't even bother to define `to`. *throws hands up in despair*

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+              "prev_batch": "page1_token"
+            }
+          schema:
+            type: object
+            properties:
+              chunk:
+                title: "ChildEventsChunk"
+                type: array
+                description: |-
+                  The child events which point to the parent event, after applicable filters.
```
richvdh, line 144: what are applicable filters? There is no `filter` param?

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+              }],
+              "next_batch": "page2_token",
+              "prev_batch": "page1_token"
+            }
+          schema:
+            type: object
+            properties:
+              chunk:
+                title: "ChildEventsChunk"
+                type: array
```
richvdh, line 142: I believe this is mandatory

```suggestion
              chunk:
                title: "ChildEventsChunk"
                type: array
                required: true
```

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+                  An opaque string representing a pagination token. The absence of this token
+                  means there are no more results to fetch and the client should stop paginating.
+              prev_batch:
+                type: string
+                description: |-
+                  An opaque string representing a pagination token. The absence of this token
+                  means there are no prior results to fetch, i.e. this is the first batch.
+        404:
+          description: |-
+            The supposed parent event was not found or the user does not have permission to read
```
richvdh, line 160:
```suggestion
            The parent event was not found or the user does not have permission to read
```

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+---
+
+### Rich replies
+
+{{% changed-in v="1.3" %}}
+
+More common on [`m.room.message`](#mroommessage) events, rich replies are a
+special kind of [relationship](#forming-relationships-between-events) which
+effectively quotes the referenced event for the client to render/process how
+it wishes.
```
richvdh, line 12: clarity:

```suggestion
Rich replies are a
special kind of [relationship](#forming-relationships-between-events) which
effectively quotes the referenced event for the client to render/process how
it wishes. They are normally used with [`m.room.message`](#mroommessage) events.
```

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+
+Additionally, a rich reply can reference any other event type as of v1.3.
+Previously, a rich reply could only reference another `m.room.message` event.
+{{% /boxes/note %}}
+
+When possible, events SHOULD include a [fallback representation](#fallbacks-for-rich-replies)
+to allow clients which do not render rich replies to still see something which
+appears to be a quoted reply.
+
+Though rich replies are forming a relationship to another event, they do not
```
richvdh, line 28:
```suggestion
Though rich replies are form a relationship to another event, they do not
```

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+use `rel_type` to create this relationship. Instead, a subkey named `m.in_reply_to`
+is used to describe the reply's relationship, leaving the other keys of
```
richvdh, line 30:
```suggestion
is used to describe the reply's relationship, leaving the other properties of
```

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+`m.relates_to` to describe the primary relationship of the event. This means
+that if an event is simply in reply to another event, without further relationship,
+the `rel_type` and `event_id` keys of `m.relates_to` become *optional*.
```
richvdh, line 33:
```suggestion
the `rel_type` and `event_id` properties of `m.relates_to` become *optional*.
```

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+Clients that do support rich replies SHOULD provide the fallback format on
+replies, and MUST strip the fallback before rendering the reply. The
+specific fallback text is different for each `msgtype`, however the
+general format for the `body` is:
+
+```
+> <@alice:example.org> This is the original body
+
+This is where the reply goes
+```
```
richvdh, line 69: I don't think this is rendering how you hoped. Likewise in a number of places below.

richvdh (changes requested):
An impressive body of work! :)

Unfortunately, it's yielded an impressive number of comments from me, ranging from the quibbling to the show-stopping. Hopefully it is clear which is which.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
 
 {{% http-api spec="client-server" api="redaction" %}}
 
+### Forming relationships between events
+
+{{% changed-in v="1.3" %}}
```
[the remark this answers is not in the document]
turt2live: technically this is a change - it got moved.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+In some cases it is desirable to logically associate one event's contents with
+another event's contents. For example, when replying to a message, editing an
+event, or simply looking to add context for an event's purpose.
+
+Events are related to each other in a parent/child structure, where any event can
+become a parent by simply having a child event point at it. Parent events do not
+define their children, instead relying on the children to describe their parent.
```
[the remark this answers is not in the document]
turt2live: yea... I'd kinda like to test this terminology in production: if it becomes too confusing, it'd be a (series of) clarification PRs, which are easier to get through than a stack of MSCs 😇

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+{{% definition path="api/client-server/definitions/m.relates_to" %}}
+
+#### Relationship types
+
+{{% boxes/note %}}
+While this specification describes an `m.relates_to` object containing a `rel_type`, there
+is not currently any relationship type which uses this structure. Replies, described below,
+form their relationship outside of the `rel_type` in order to allow other, future, relationship
+types to make use of replies in addition to their normal behaviour.
```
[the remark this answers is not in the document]
turt2live: I feel like we need *something* to justify it, and it was cited out of band in various threads discussions for why we don't just make replies a `rel_type`. Open to other ideas though.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+consider child events from ignored users when preparing a bundle for the client.
+{{% /boxes/note %}}
+
+When a parent event is redacted, the child events which pointed to that parent remain, however
+when a child event is redacted then the relationship is broken. Therefore, the server needs
+to de-aggregate or disassociate the event once the relationship is lost. Clients with local
+aggregation or which handle redactions locally should do the same.
+
+It is suggested that clients perform local echo on aggregations. For instance, aggregating
+a new child event into a bundle optimistically until the server returns a failure or the client
```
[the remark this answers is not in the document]
turt2live: yea... if we can, I'd like to defer this until we have a real example in the spec.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+The actual aggregation format depends on the `rel_type`.
+
+{{% boxes/note %}}
+This specification does not currently describe any `rel_type` which require
+aggregation, however [namespaced](/appendices#identifier-grammar) relationship
+types might have aggregation behaviour.
+{{% /boxes/note %}}
+
+When aggregations are summarised on an event, it is known as a "bundled aggregation"
```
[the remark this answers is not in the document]
turt2live: 😬 I thought it was the same thing - will take a look and see what I can do about it.

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+        down by relationship type and related event type.
+
+        Note that while this endpoint indicates `relType` and `eventType` being
+        required, they can be individually omitted to retrieve all matching events.
+        When omitting the path parameters, a trailing slash must not be included
+        (otherwise the search will be for an empty string).
+
+        When combining the use of `relType` and `eventType`, the server will return
+        child events which match *both* conditions rather than *either*.
+      operationId: getChildEventsForParentEvent
```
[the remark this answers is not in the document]
turt2live: "but my image relates to the text I sent after!" - I'd rather maintain the same terminology used in the spec to the operation ID if we can.

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+          type: string
+          name: from
+          description: |-
+            The pagination token to start returning results from. If not supplied, results
+            start at the earliest topological event known to the server.
+
+            Can be a `next_batch` or `prev_batch` token from a previous call, or an equivalent
+            token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
+            or [`/sync`](/client-server-api/#get_matrixclientv3sync) to limit results to the
+            events returned by that section of timeline.
```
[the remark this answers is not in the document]
turt2live: quite honestly, while writing the whole thing on pagination I was trying to figure out how in the hell you're expected to use the API. I thought I fixed it before putting up for review, but evidentially not (sorry).

Another question is what use it is asking the server for events in the gap: surely just throw away your data and try again if you got confused? (and indeed: what does it mean to paginate off a `/sync` or `/messages` section? do we return pagination tokens for the paginated section, or can a client roll backwards over its filtered `/sync`?)

The best approach here is probably to:
1. Figure out wtf synapse is actually doing
2. Make the MSC and this PR reflect that
3. Open a bug saying it's bad (probably - if it's actually good then we can skip this step)

Figuring out what synapse is even doing might be a challenge though - will seek help if needed. 

(also, if what synapse is doing is just plain insane then let's figure out a plan to spec a sensible thing instead)

content/client-server-api/_index.md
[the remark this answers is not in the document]
richvdh: aggregation is the process of building up the summary of relating events. You can get hold of that aggregation in various ways; one of them is by having them *bundled* into said event.  That's AIUI, anyway.

data/api/client-server/relations.yaml
turt2live: for my own context when I revisit this: "filter" means the optional path params.

data/api/client-server/relations.yaml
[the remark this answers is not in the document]
richvdh: I don't disagree. It's just that this one is going to be slightly less amenable to the "we'll just clarify it later if we change our minds" approach, since `operationId`s are meant to be constant.

(also, the endpoint itself is called `/relations`, not `/children`).

On a range of -1 to +1, I'm about -0.3 on your name: I'd prefer to see a change but if you want to stick with what you have, that's fine.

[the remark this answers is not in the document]
richvdh: (and yes, my suggestion isn't great either)

content/client-server-api/modules/rich_replies.md
turt2live: 😭 😭 😭 😭 😭 

will fix, somehow.

data/api/client-server/relations.yaml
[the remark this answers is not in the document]
richvdh: synapse, for the record, just uses stream tokens here, and always paginates *forwards* from `from`, even if it was a `prev_batch` from a previous query. It's not *totally* insane, but it's not consistent with the pagination appendix, and it needs MSC3715 to let you go backwards.

(you can even take a `next_batch` from a `/relations` and feed it into a `/messages` or a `/sync` if you want. Is that a thing clients should be allowed to do? I doubt it, but who knows.)

data/api/client-server/relations.yaml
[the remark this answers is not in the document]
turt2live: your suggestion is a `-0.295` if I were to use your scale, so let's just use that.
```suggestion
      operationId: getRelatingEvents
```

turt2live:
thanks for the review, and apologies again for the dumpster fire that is a commit history on this. Hopefully will be improved for future MSC stacks.

I think your suggestions generally make sense - will take a closer look hopefully tomorrow and unpick pagination in the process.

content/client-server-api/_index.md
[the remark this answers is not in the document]
clokep: We do use `parent_id` in a few spots in the code, I don't think we use "child" anywhere though.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+State events do not currently receive the `m.relations` unsigned field. This is not
+necessarily a deliberate design decision, and MSCs which aim to fix this are welcome.
+{{% /boxes/warning %}}
+
+The endpoints where the server *should* include the `m.relations` unsigned field are:
+
+* [`GET /rooms/{roomId}/messages`](#get_matrixclientv3roomsroomidmessages)
+* [`GET /rooms/{roomId}/context/{eventId}`](#get_matrixclientv3roomsroomidcontexteventid)
+* [`GET /rooms/{roomId}/event/{eventId}`](#get_matrixclientv3roomsroomideventeventid)
+* [`GET /rooms/{roomId}/relations`](#get_matrixclientv3roomsroomidrelations)
```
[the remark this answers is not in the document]
clokep: This is used by the `m.thread` relations in MSC3440 (if you look up `/rooms/{roomId}/relations/m.thread`, you want to get back an array of events and they're relations, e.g. the reactions to each of those messages, for easy rendering).

content/client-server-api/_index.md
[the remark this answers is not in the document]
clokep: I don't think there's any other way o get hold of aggregations anymore -- the `/aggregations` endpoint was removed from MSC2675. That makes them essentially the same thing.

data/api/client-server/relations.yaml
turt2live: resolving thread: the field is destined to disappear instead. If an MSC wants to add it, a new one can be opened.

content/client-server-api/_index.md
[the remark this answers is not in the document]
richvdh: Ah, that's useful context, thanks @clokep.

Still, I find it easier to think about them as slightly separate concepts - not least because it's conceivable we might reintroduce another way of getting the aggregations in future.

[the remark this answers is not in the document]
clokep: > Still, I find it easier to think about them as slightly separate concepts - not least because it's conceivable we might reintroduce another way of getting the aggregations in future.

Yeah, that's fair. I assume your definitions are something like:

* Aggregation: a format for returning a summary or grouping of child events which share a parent event.
* Bundled aggregations: pre-calculated aggregations which are returned to clients as part of the `unsigned` field by the server.

content/client-server-api/_index.md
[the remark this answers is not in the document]
clokep: I also asked @turt2live to choose clear names to avoid some of the confusing conversations that @richvdh and I had while reviewing this code!

I don't know if parent/child is the proper thing, I think some code also uses target/source. I would stay away from "related" though since you can end up using it to describe both a parent-child relationship as well as a child-parent relations.

content/client-server-api/modules/rich_replies.md
richvdh: sorry.

```suggestion
Though rich replies form a relationship to another event, they do not
```

turt2live: I'm still reading it wrong, but maybe that's a me problem.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+Note how the `org.example.possible_annotations` bundle is an array compared to the
+`org.example.possible_thread` bundle where the server is summarising the state of
+the relationship in a single object. Both are valid bundles, and their exact types
+depend on the `rel_type`.
+
+{{% boxes/warning %}}
+State events do not currently receive the `m.relations` unsigned field. This is not
+necessarily a deliberate design decision, and MSCs which aim to fix this are welcome.
+{{% /boxes/warning %}}
```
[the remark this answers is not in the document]
turt2live: it was called out in one of the mscs 🤷

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,170 @@
...
+    get:
+      summary: Get the child events for a given parent event, with optional filter.
+      description: |-
+        Retrieve all of the child events for a given parent event, optionally filtering
+        down by relationship type and related event type.
+
+        Note that while this endpoint indicates `relType` and `eventType` being
+        required, they can be individually omitted to retrieve all matching events.
+        When omitting the path parameters, a trailing slash must not be included
+        (otherwise the search will be for an empty string).
```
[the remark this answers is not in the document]
turt2live: ... I wonder if there's a way to do this that doesn't involve duplication, like we did for `/state/{type}/{stateKey}`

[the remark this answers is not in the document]
turt2live: answer: no, we just have to copy/paste the thing 3 times. 

I'm very tempted to hack our validator to make this work without copy/paste, and call it an extension for parsers to handle.

content/client-server-api/_index.md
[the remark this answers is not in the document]
richvdh: there's a difference between mentioning in an MSC that a design isn't necessarily optimal, and adding text to the spec actively asking for proposals to change it. But :shrug:

content/client-server-api/_index.md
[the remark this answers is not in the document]
turt2live: Have rewritten the last bit as:

> Replies, described below,
> form their relationship outside of the `rel_type` as a legacy type of relationship. Future
> versions of the specification might change replies to better match the relationship structures.

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+{{% boxes/note %}}
+While this specification describes an `m.relates_to` object containing a `rel_type`, there
+is not currently any relationship type which uses this structure. Replies, described below,
+form their relationship outside of the `rel_type` in order to allow other, future, relationship
+types to make use of replies in addition to their normal behaviour.
+
+Custom `rel_type`s can, and should, still use the schema described above for relevant
+behaviour.
+{{% /boxes/note %}}
```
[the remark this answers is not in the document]
turt2live: moved up a few lines

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+
+This specification describes the following relationship types:
+
+* [Rich replies](#rich-replies) (**Note**: does not use `rel_type`).
+
+{{% boxes/note %}}
+This specification does not currently define any relation type which requires
+aggregation or has restrictions, however [namespaced](/appendices#identifier-grammar)
+relationship types might have these restrictions.
```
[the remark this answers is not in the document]
turt2live: have removed the box outright - is indeed duplicate

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+for events that are "out of range".
+
+Additionally, if the server is missing portions of the room history then it may not be
+able to accurately aggregate the events.
+{{% /boxes/warning %}}
+
+#### Relationships API
+
+To retrieve the child events for a parent from the server, the client can call the
+following endpoint with relevant information. This endpoint does not aggregate the child
```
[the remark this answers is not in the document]
turt2live: applied manually

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,221 @@ the topic to be removed from the room.
...
+events and is instead paginated: clients can perform local aggregation if needed. This
```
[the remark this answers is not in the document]
turt2live: Have rewritten as:

> To retrieve the child events for a parent from the server, the client can call the
> following endpoint. The events returned might require local aggregation by the client,
> however if those same events have bundles of their own then they should be present. A
> hypothetical example would be retrieving all responses to a thread, and those responses
> might have emoji reactions made against them: the client will need to construct the
> thread itself, but the events in the thread should have a reactions bundle on them.
> 
> The lack of aggregation behaviour allows clients to retrieve child events which don't
> require aggregation, but do make use of a `rel_type`.

data/api/client-server/relations.yaml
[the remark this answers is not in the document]
turt2live: copy/pasted

data/api/client-server/relations.yaml
turt2live: done so

data/api/client-server/relations.yaml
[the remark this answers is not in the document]
turt2live: I think I've done this, or at least something on the same page as that.

data/api/client-server/relations.yaml
turt2live: this appears to have been an MSC bug, ftr.

content/client-server-api/_index.md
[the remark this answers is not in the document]
turt2live: have clarified that with an example.

turt2live:
@richvdh this should be ready for re-review, with the commits since your last review being reviewable individually (largely).

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,219 @@ the topic to be removed from the room.
...
+{{% boxes/warning %}}
+Due to history visibility restrictions, child events might not be visible to the user
+if they are in a section of history the user cannot see. This can mean inaccurate aggregations
+for events that are "out of range".
+
+Additionally, if the server is missing portions of the room history then it may not be
+able to accurately aggregate the events.
+{{% /boxes/warning %}}
+
+#### Relationships API
```
richvdh, line 2036:
```suggestion
#### Relationships API

 {{% added-in v="1.3" %}}
```

... if the "Aggregations" subsection is getting an `added-in`, this needs one too.

content/client-server-api/_index.md
```diff
@@ -1927,17 +1919,17 @@ The actual aggregation format depends on the `rel_type`.
 
 {{% boxes/note %}}
 This specification does not currently describe any `rel_type` which require
```
richvdh, line 1921:
```suggestion
This specification does not currently describe any `rel_type`s which require
```

content/client-server-api/_index.md
```diff
@@ -1927,17 +1919,17 @@ The actual aggregation format depends on the `rel_type`.
...
-aggregation, however [namespaced](/appendices#identifier-grammar) relationship
-types might have aggregation behaviour.
+aggregation. This functionality forms a framework for future extensions.
 {{% /boxes/note %}}
 
-When aggregations are summarised on an event, it is known as a "bundled aggregation"
-or "bundle" for simplicity. The act of doing this is "bundling".
+Aggregations are sometimes automatically included by a server alongside the parent
+event. This This is known as a "bundled aggregation" or "bundle" for simplicity. The
```
richvdh, line 1926:
```suggestion
event. This is known as a "bundled aggregation" or "bundle" for simplicity. The
```

content/client-server-api/_index.md
```diff
@@ -1831,6 +1841,219 @@ the topic to be removed from the room.
...
+gives up on sending the event, at which point the event should be de-aggregated and an
+error or similar shown. The client should be cautious to not aggregate an event twice if
+it has already optimistically aggregated the event. Clients are encouraged to take this
+a step further to additionally track child events which target unsent/pending events,
+likely using the transaction ID as a temporary event ID until a proper event ID is known.
+
+{{% boxes/warning %}}
+Due to history visibility restrictions, child events might not be visible to the user
+if they are in a section of history the user cannot see. This can mean inaccurate aggregations
+for events that are "out of range".
```
richvdh, line 2030: I'm not quite sure what this is trying to say. The server might return an inaccurate aggregation? Or the client might miscalculate?

content/client-server-api/_index.md
```diff
@@ -2042,12 +2036,16 @@ able to accurately aggregate the events.
 #### Relationships API
 
 To retrieve the child events for a parent from the server, the client can call the
-following endpoint with relevant information. This endpoint does not aggregate the child
-events and is instead paginated: clients can perform local aggregation if needed. This
-allows clients to retrieve child events which do not require aggregation but still make
-use of `rel_type`.
+following endpoint. The events returned might require local aggregation by the client,
+however if those same events have bundles of their own then they should be present. A
```
richvdh, line 2040: >  The events returned might require local aggregation by the client,

isn't this true of every endpoint that returns events?

content/client-server-api/_index.md
```diff
@@ -2042,12 +2036,16 @@ able to accurately aggregate the events.
...
+hypothetical example would be retrieving all responses to a thread, and those responses
+might have emoji reactions made against them: the client will need to construct the
+thread itself, but the events in the thread should have a reactions bundle on them.
+
+The lack of aggregation behaviour allows clients to retrieve child events which don't
+require aggregation, but do make use of a `rel_type`.
```
richvdh, line 2046: isn't the definition of a "child event" (in this context) that it "makes use of a `rel_type`"  ?

data/api/client-server/relations.yaml
```diff
@@ -27,55 +27,303 @@ produces:
...
+      summary: Get the child events for a given parent event.
+      description: |-
+        Retrieve all of the child events for a given parent event.
+
+        Note that when paginating the `from` token must be *before* the `to` token, if
+        supplied. This means it's only possible to paginate "forwards" through events.
+        An example of an *invalid* call would be a `from` token from page 2 of the results,
+        and a `to` token from page 1: the server can't go backwards with this information.
+        The caller can use a `from` token from page 1 and a `to` token from page 2 to
+        paginate forwards over the same range, however.
```
richvdh, line 41: I don't think that's *invalid*, per se, it's just that you're asking for the empty set.

It's not the fact that `from` must be before `to` that determines the fact that you can only go forwards; it's that directionality isn't implied by the relative orders of `from` and `to`. (Currently we *always* go forwards. MSC3715 changes that.)

Likewise below, x2

data/api/client-server/relations.yaml
```diff
@@ -84,10 +332,9 @@ paths:
             The pagination token to start returning results from. If not supplied, results
             start at the earliest topological event known to the server.
 
-            Can be a `next_batch` or `prev_batch` token from a previous call, or an equivalent
-            token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
-            or [`/sync`](/client-server-api/#get_matrixclientv3sync) to limit results to the
-            events returned by that section of timeline.
+            Can be a `next_batch` or `prev_batch` token from a previous call, or a returned
+            `start`/`end` token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
```
richvdh, line 336:
```suggestion
            `start`/`end` token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages),
```

Likewise above x2

data/api/client-server/relations.yaml
```diff
@@ -84,10 +332,9 @@ paths:
             The pagination token to start returning results from. If not supplied, results
             start at the earliest topological event known to the server.
 
-            Can be a `next_batch` or `prev_batch` token from a previous call, or an equivalent
-            token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
-            or [`/sync`](/client-server-api/#get_matrixclientv3sync) to limit results to the
-            events returned by that section of timeline.
+            Can be a `next_batch` or `prev_batch` token from a previous call, or a returned
```
richvdh: I don't understand why it is useful to pass a `prev_batch` from a previous call here (at until MSC3716 lands).

Likewise above x2

[edit richvdh 2026-04-21: I meant MSC3715)

data/api/client-server/relations.yaml
```diff
@@ -84,10 +332,9 @@ paths:
...
+            `start`/`end` token from [`/messages`](/client-server-api/#get_matrixclientv3roomsroomidmessages)
+            or `next_batch`/`prev_batch` token from [`/sync`](/client-server-api/#get_matrixclientv3sync).
```
richvdh, line 337:
```suggestion
            or a `next_batch`/`prev_batch` token from [`/sync`](/client-server-api/#get_matrixclientv3sync).
```

Likewise above x2

data/api/client-server/relations.yaml
```diff
@@ -141,7 +389,8 @@ paths:
                 title: "ChildEventsChunk"
                 type: array
                 description: |-
-                  The child events which point to the parent event, after applicable filters.
+                  The child events of the requested event. If a `relType` or `eventType` was
+                  supplied on the URL, the events returned will match those details.
```
richvdh, line 393:
```suggestion
                  The child events of the requested event. The events returned will 
                  match the `relType` or `eventType` supplied in the URL.
```

Likewise above (with modifications) x2

data/api/client-server/relations.yaml
```diff
@@ -141,7 +389,8 @@ paths:
                 title: "ChildEventsChunk"
                 type: array
                 description: |-
```
richvdh, line 391: still missing its `required` flag

```suggestion
                title: "ChildEventsChunk"
                type: array
                required: true
                description: |-
```

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,421 @@
...
+              next_batch:
+                type: string
+                description: |-
+                  An opaque string representing a pagination token. The absence of this token
+                  means there are no more results to fetch and the client should stop paginating.
+              prev_batch:
+                type: string
+                description: |-
+                  An opaque string representing a pagination token. The absence of this token
+                  means there are no prior results to fetch, i.e. this is the first batch/page.
```
richvdh, line 406:
```suggestion
                  An opaque string representing a pagination token. The absence of this token
                  means this is the start of the result set, i.e. this is the first batch/page.
```

Likewise above x2

content/client-server-api/modules/rich_replies.md
```diff
@@ -0,0 +1,182 @@
...
+Some clients may not have support for rich replies and therefore need a
+fallback to use instead. Clients that do not support rich replies should
+render the event as if rich replies were not special.
+
+Clients that do support rich replies SHOULD provide the fallback format on
+replies, and MUST strip the fallback before rendering the reply. The
+specific fallback text is different for each `msgtype`, however the
+general format for the `body` is:
+
+```
```
richvdh, line 65: this is missing its `text` content-type-marker-thingio

richvdh (changes requested):
thanks for the updates on this!

content/client-server-api/_index.md
turt2live: I think this is trying to say "inaccurate bundle" instead, or some other variation of "you might be missing all related events - don't use polls or reactions for voting a prime minister in".

content/client-server-api/_index.md
turt2live: this is trying to clarify the confusion formed around why `/relations` would be listed as an endpoint which does aggregation, but isn't doing the aggregation on the top-level response (instead bundling for the events contained within). 

Very open to suggestions on this.

data/api/client-server/relations.yaml
turt2live: Sorry, I don't understand what needs to change here. If we swap "invalid" for "returns empty set", does that fix it?

data/api/client-server/relations.yaml
richvdh: fair, though the fact we explicitly permit it does rather imply that it might be a useful thing to do, which it's not.

data/api/client-server/relations.yaml
richvdh: That's a start!

We've also now discovered this endpoint goes backwards in time in its current, pre-MSC3715, implementation. Taking that into account:

```suggestion
        Note that when paginating the `from` token should be "after" the `to` token in
        terms of topological ordering, because it is only possible to paginate "backwards" 
        through events, starting at `from`.
        
        For example, passing a `from` token from page 2 of the results, and a `to` token
        from page 1, would return the empty set. The caller can use a `from` token from
        page 1 and a `to` token from page 2 to paginate over the same range, however.
```

I'm not sure the example helps, and I wonder if it would be clearer without.

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,421 @@
...
+            Note that in encrypted rooms this will typically always be `m.room.encrypted`
+            regardless of the event type contained within the encrypted payload.
+          required: true
+          x-example: "m.room.message"
+        - in: query
+          type: string
+          name: from
+          description: |-
+            The pagination token to start returning results from. If not supplied, results
+            start at the earliest topological event known to the server.
```
richvdh, line 333: s/earliest/most recent/.

Likewise above x2

data/api/client-server/relations.yaml
turt2live: that reads clear enough, I think. Certainly helps understand the problem at least :D

content/client-server-api/_index.md
```diff
@@ -1877,45 +1877,37 @@ is no plaintext copy). This is to ensure the client's behaviour matches the serv
...
-`m.relates_to` is described as follows:
-
-{{% definition path="api/client-server/definitions/m.relates_to" %}}
-
-#### Relationship types
+Relationships which don't match the schema, or which break the rules of a relationship,
+are simply ignored. Such an example might be the parent and child being in different
+rooms or the relationship missing properties required by the schema below. The events
+would appear independent of each other or optionally with an error message (if
+rendered/handled by the client exclusively).
```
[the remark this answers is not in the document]
turt2live: I think the exclusivity statement is from a previous draft that didn't get cleaned up properly (likely related to threads). Proposed wording makes a lot more sense - thanks 👍

data/api/client-server/relations.yaml
```diff
@@ -0,0 +1,421 @@
...
+            }
+          schema:
+            type: object
+            properties:
+              chunk:
+                title: "ChildEventsChunk"
+                type: array
+                description: |-
+                  The child events of the requested event. If a `relType` or `eventType` was
+                  supplied on the URL, the events returned will match those details.
```
richvdh, line 393: We need to define the order in which the results are returned. I believe it's most-recent first. (ie, reverse chronological order).

data/api/client-server/relations.yaml
turt2live: applied manually

data/api/client-server/relations.yaml
turt2live: removed the end batch tokens manually

data/api/client-server/relations.yaml
turt2live: applied manually

data/api/client-server/relations.yaml
turt2live: applied manually

content/client-server-api/_index.md
turt2live: it is, and I'm convinced there's no value coming from the latter two thirds of the paragraph as well (per other comments).

data/api/client-server/relations.yaml
richvdh: right, sorry.

richvdh (approved):
let's ship this at last

Issue #817, opened 2021-04-21
Exact Federation `Authorization` header format isn't clear 


ShadowJonathan:
#### Link to problem area

https://matrix.org/docs/spec/server_server/r0.1.4#request-authentication

#### Issue
Came up in `#conduit:fachschaften.org`; It's not clear that the parameters after `X-Matrix` can use quotes or not, [RFC2184](https://tools.ietf.org/html/rfc2184) was mentioned with the following quote:

> Note that quotes around parameter values are part of the value
> syntax; they are NOT part of the value itself.  Furthermore, it is
> explicitly permitted to have a mixture of quoted and unquoted
> continuation fields.

Maybe this clarification can be added as an informational footnote.

uhoreg:
The format for the Authorization header is actually in [RFC2735](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1), and *almost* matches what is done in Matrix.  The approach that I've taken in https://github.com/matrix-org/matrix-spec/pull/1067 is to say that we follow RFC2735, but if you want to be nice to servers, follow these more restricted rules.

Issue #511, opened 2019-07-29
IS spec doesn't mention that X-Matrix auth is different


turt2live:
It uses `destination_is` instead of `destination`

PR #1067, opened 2022-05-17
clarify federation Authorization header an add destination property

uhoreg:
spec PR for https://github.com/matrix-org/matrix-spec-proposals/pull/3383
fixes https://github.com/matrix-org/matrix-spec/issues/817

Preview: https://pr1067--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/shortcodes/added-in.html
+++ b/layouts/shortcodes/added-in.html
@@ -2,7 +2,7 @@
 {{ $this := .Params.this }}
 
 {{ if $this }}
-  <span>**[New in this version]**</span>
+  <span><strong>[New in this version]</strong></span>
 {{ else }}
-  <span>**[Added in `v{{ $ver }}`]**</span>
+  <span><strong>[Added in <code>v{{ $ver }}</code>]</strong></span>
 {{ end }} {{/* Do not leave an empty line at the end of this file otherwise the inline behaviour breaks. */}}
--- a/changelogs/server_server/newsfragments/1067.clarification
+++ b/changelogs/server_server/newsfragments/1067.clarification
@@ -1,0 +1,1 @@
+Clarify the format for the Authorization header.
--- a/changelogs/server_server/newsfragments/1067.feature
+++ b/changelogs/server_server/newsfragments/1067.feature
@@ -1,0 +1,1 @@
+Add a destination property to the Authorization header.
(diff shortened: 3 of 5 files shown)


turt2live:
@uhoreg is this ready for review?

uhoreg:
Should be ready now

layouts/shortcodes/added-in.html
```diff
@@ -2,7 +2,7 @@
 {{ $this := .Params.this }}
 
 {{ if $this }}
-  <span>**[New in this version]**</span>
```
uhoreg, line 5: These changes are needed, otherwise it displays the markdown in the federation docs, rather than the HTML

turt2live: hmm.. that's surprising. What version of hugo?

jcgruenhage:
https://github.com/matrix-org/matrix-spec/issues/511 which is sort-of related will be solved separately, because it's a bigger change I'll assume? Ofher than that, this looks hood to me. Thanks!

layouts/shortcodes/added-in.html
uhoreg: It happened with whatever version of hugo CI was using.  It worked fine in the CS bits.  I assume it's because it got substituted in, and then the file got included in the main file, and then reprocessed.  But it didn't work for the federation doc.

uhoreg:
> #511 which is sort-of related will be solved separately, because it's a bigger change I'll assume?

Oh, I didn't notice that one.  Yes, I think that would be a separate PR.  I think I'm already stretching it, making three changes in 1 PR.

layouts/shortcodes/added-in.html
turt2live: right, yea. `{{< shortcode >}}` is a literal while `{{% shortcode %}}` is interpreted - we need to use angle brackets to get the inline stuff to work, annoyingly.

turt2live: This is fun though (and not worth fixing in this PR, I think):

s2s spec:
image

c2s spec:
image

The code blocks are different. I think it's because of the table structure, but not really sure.

turt2live (approved):
otherwise lgtm - thanks!

PR #1082, opened 2022-05-24
Clarification on historical power level handling

neilalexander:
This PR replaces and therefore closes matrix-org/matrix-spec-proposals#3688 and includes the changes suggested from the review comments.

Preview: https://pr1082--matrix-spec-previews.netlify.app

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -68,17 +68,17 @@ Server names are resolved to an IP address and port to connect to, and
 have various conditions affecting which certificates and `Host` headers
 to send. The process overall is as follows:
 
-1.  If the hostname is an IP literal, then that IP address should be
+1. If the hostname is an IP literal, then that IP address should be
     used, together with the given port number, or 8448 if no port is
     given. The target server must present a valid certificate for the IP
     address. The `Host` header in the request should be set to the
     server name, including the port if the server name included one.
-2.  If the hostname is not an IP literal, and the server name includes
+2. If the hostname is not an IP literal, and the server name includes
     an explicit port, resolve the IP address using AAAA or A records.
     Requests are made to the resolved IP address and given port with a
     `Host` header of the original server name (with port). The target
     server must present a valid certificate for the hostname.
-3.  If the hostname is not an IP literal, a regular HTTPS request is
+3. If the hostname is not an IP literal, a regular HTTPS request is
     made to `https://<hostname>/.well-known/matrix/server`, expecting
     the schema defined later in this section. 30x redirects should be
     followed, however redirection loops should be avoided. Responses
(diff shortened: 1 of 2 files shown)


content/server-server-api.md
```diff
@@ -394,6 +394,24 @@ unspecified.
 For an `m.room.member` state event, the user given by the `state_key` of
 the event.
 
+**Historical String Power Levels** \
```
turt2live, line 397: shouldn't this be in the room version spec? (it'll also get adopted as part of v10 being specced if you'd prefer to make it someone else's problem to land, btw)

neilalexander: I mostly figured it made sense here because it affects all current room versions and nothing else documents that behaviour anywhere.

turt2live: ah, yea, if we didn't have a room version on the way then it'd land here for sure. v10 is currently in FCP and will land in Matrix 1.3 though (a few weeks away), so time delay seems to have bit us here :( 

if you're okay with it, I'm happy to take over the inclusion of this via v10 spec.

content/server-server-api.md
```diff
@@ -394,6 +394,24 @@ unspecified.
...
+In order to maintain backwards compatibility with early implementations,
+power levels can optionally be represented in string format instead of
+integer format. A homeserver must be prepared to deal with this by parsing
+the power level from a string. In these cases, the following formatting of the
+power level string is allowed:
+
+- a single Base10 integer, no float values or decimal points, optionally with leading zeroes;
```
jplatte, line 404: For maximum clarity, it should be noted here that this integer has to be within the usual `[MIN_SAFE_INTEGER, MAX_SAFE_INTEGER]` range (i.e. Synapse [still? / previously?] accepting smaller / larger ones is "just" a bug).

turt2live: @jplatte that was fixed in https://github.com/matrix-org/matrix-spec-proposals/pull/2540

jplatte: I'm talking specifically about stringified integers outside of that range, which were still accepted by Synapse in all room versions (i.e. including v6+) recently according to a comment on one of the enforce-integer-powerlevels related issues / PRs.

content/server-server-api.md
neilalexander: My feeling is that a "historical" section probably still makes sense here regardless of v10 because it's behaviour that a server implementor really needs to know exists from the past, although maybe we want to modify the text to clarify that the behaviour only applies to room versions 9 and below?

I'm more than happy for you to take over rolling something into v10 though either way.

content/server-server-api.md
turt2live: 😭 well that's frustrating

jplatte: I think it's fine actually, it wasn't found to be used / exploited in the wild and can really only happen if somebody intentionally creates such a PL event. It would actually be near impossible to support current Synapse's behavior in other HSes anyways, since Synapse converts to Python's `int`, which even allows values >64bit.

content/server-server-api.md
turt2live: The spec language is incredibly frustrating because we can't say "v9 or below", because they aren't linear. A note here is certainly worthwhile, though it might get moved into the room version spec and a note left here to say "note that power level parsing is dependent on the room version" or something.

turt2live: process-wise we increment by one, but the spec is much more strict on how it words things, for better or worse.

turt2live:
closing in favour of https://github.com/matrix-org/matrix-spec/pull/1099 which defines it as part of v10 (backporting to v1-9 and leaving a note in the s2s spec).

PR #1084, opened 2022-05-25
Clarify that /invite will respond with 200 if the user is already invited to the room

### LINKED ISSUE #1083: Clarify the response body/status of /invite when the user is already invited
**Link to problem area**: https://spec.matrix.org/v1.2/client-server-api/#post_matrixclientv3roomsroomidinvite

**Issue**

When a user is invited to the room (via `/invite`) while **already** being invited to the room, it's not clear from the spec on `/invite` what should happen. I couldn't see any text on https://spec.matrix.org/latest/client-server-api/#room-membership on what happens if a user transitions from the same state to the same state either.

Currently, Synapse will just return an empty response body as if you had invited a non-invited user. However, it would be worth clarifying that the endpoint is expected to do that to avoid relying on undefined behaviour.

Half-Shot:
Potentially fixes #1083 

I *believe* this might fall within the scope of minor text clarification rather than requiring a full MSC, as it's a behavior that has been in action for a long time (for Synapse at least, and possibly other implementations?). This doesn't seek to correct the response.

Preview: https://pr1084--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1084.clarification
+++ b/changelogs/client_server/newsfragments/1084.clarification
@@ -1,1 +1,1 @@
-Mention that the /rooms/$roomId/invite endpoint will return a 200 response if the user is already invited to the room.
+Mention that the `/rooms/{roomId}/invite` endpoint will return a 200 response if the user is already invited to the room.
--- a/changelogs/client_server/newsfragments/1084.clarification
+++ b/changelogs/client_server/newsfragments/1084.clarification
@@ -1,0 +1,1 @@
+Mention that the /rooms/$roomId/invite endpoint will return a 200 response if the user is already invited to the room.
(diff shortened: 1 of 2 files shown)


turt2live (changes requested):
This needs a changelog entry, and evidence that this is in fact the case please :)

turt2live:
@Half-Shot are you able to take a look at this?

Half-Shot:
Yep, let me sort that out.

changelogs/client_server/newsfragments/1084.clarification
```diff
@@ -0,0 +1 @@
+Mention that the /rooms/$roomId/invite endpoint will return a 200 response if the user is already invited to the room.
```
turt2live, line 1:
```suggestion
Mention that the `/rooms/{roomId}/invite` endpoint will return a 200 response if the user is already invited to the room.
```

turt2live (approved):
otherwise lgtm

PR #1093, opened 2022-05-27
Clarify that the resident server should sign the restricted join

### LINKED ISSUE #1092: Auth rule for restricted joins suggests an event should be signed by a user
**Link to problem area**: https://spec.matrix.org/v1.2/rooms/v9/#authorization-rules

**Issue**

> 4. If type is m.room.member:
> 
>     1. If no state_key key or membership key in content, reject.
>     2. If content has a join_authorised_via_users_server key:
>         1. If the event is not validly **signed by the user ID denoted by the key**, reject.
> 
> 

Emphasis mine. I think it should read "signed by _the homeserver of_ the user ID denoted by the key".

DMRobertson:
Fixes #1092.

Preview: https://pr1093--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1093.clarification
+++ b/changelogs/room_versions/newsfragments/1093.clarification
@@ -1,0 +1,1 @@
+Clarify that the resident server must sign restricted join event.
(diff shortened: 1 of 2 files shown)


turt2live (approved):
🙈

thanks!

PR #1097, opened 2022-05-30
fix: spaces hierarchy paramater types

TheOneWithTheBraid:
- changed `limit` parameter type to integer
- changed `query` parameter type to integer

A floating number does not make any sense here. Also, at least Synapse
does not allow floating point numbers in here.

Signed-off-by: TheOneWithTheBraid <the-one@with-the-braid.cf>

Preview: https://pr1097--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1097.clarification
+++ b/changelogs/client_server/newsfragments/1097.clarification
@@ -1,1 +1,1 @@
-Adjust the data types of the Spaces Hierarchy query parameters.
+Fix various typos throughout the specification.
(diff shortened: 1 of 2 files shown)


KitsuneRal (approved):

changelogs/client_server/newsfragments/1097.clarification
```diff
@@ -0,0 +1 @@
+Adjust the data types of the Spaces Hierarchy query parameters.
```
turt2live, line 1:
```suggestion
Fix various typos throughout the specification.
```

PR #1101, opened 2022-05-31
Deprecate the `sender_key` and `device_id` on Megolm events

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3700 ([Markdown](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3700-deprecate-sender-key.md))

The language around `m.room.encrypted` is a bit awkward because *technically* you can use the event to represent non-Megolm events, however that's considered an edge case at this time.

Preview: https://pr1101--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1531,6 +1531,15 @@ For example, Megolm sessions that were sent using the old session would
 have been lost. The client can attempt to retrieve the lost sessions
 through `m.room_key_request` messages.
 
+{{% boxes/note %}}
+Clients should send key requests for unknown sessions to all devices for
+the user which used the session rather than just the `device_id` or
+`sender_key` denoted on the event.
+
+This is due to a deprecation of the fields. See
+[`m.megolm.v1.aes-sha2`](#mmegolmv1aes-sha2) for more information.
+{{% /boxes/note %}}
+
 ##### `m.megolm.v1.aes-sha2`
 
 {{% changed-in v="1.3" %}}
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1624,10 +1624,10 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
-When a client is updating a session key in its store, the client MUST ensure:
+When a client is updating a Megolm session (room key) in its store, the client MUST ensure:
 
 * that the updated session data comes from a trusted source, where a trusted source
-  could be the signatures on the session or comes from the user's session key backup;
+  could be the signatures on the session or those from the user's session key backup;
   and
 * that the new session key has a lower message index than the existing session key.
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1533,6 +1533,8 @@ through `m.room_key_request` messages.
 
 ##### `m.megolm.v1.aes-sha2`
 
+{{% changed-in v="1.3" %}}
+
 The name `m.megolm.v1.aes-sha2` corresponds to version 1 of the Megolm
 ratchet, as defined by the [Megolm
 specification](http://matrix.org/docs/spec/megolm.html). This uses:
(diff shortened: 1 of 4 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1596,6 +1624,13 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
+When a client is updating a session key in its store, the client MUST ensure:
+
+* that the updated session data comes from a trusted source, where a trusted source
+  could be the signatures on the session or comes from the user's session key backup;
+  and
+* that the new session key has a lower message index than the existing session key.
```
dkasak, line 1632:
```suggestion
When a client is updating a Megolm session (room key) in its store, the client MUST ensure:

* that the updated session data comes from a trusted source, where a trusted source
  could be the signatures on the session or comes from the user's session key backup;
  and
* that the new session has a lower message index than the existing session.
```

In order to:

1. Be explicit about this being a Megolm session. The word "session" is overloaded.
2. Use "session" instead of "session key", since there's really no sane difference between these terms. We usually say "session" or "room key", and these are synonyms on the Matrix level.

This doesn't really fix the existing terminology mess and that should be cleaned up separately, but at least we won't introduce new ambiguities.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1580,10 +1582,36 @@ ratchet index that they have already decrypted. Care should be taken in
...
+Clients MUST NOT store or lookup sessions using the `sender_key` or `device_id`.
+
+In a future version of the specification the keys can be removed completely,
+including for sending new messages.
+{{% /boxes/note %}}
+
+{{% boxes/rationale %}}
+Removing the fields (eventually) improves privacy and security by masking the
+device which sent the encrypted message as well as reducing the client's
+dependence on untrusted data: a malicious server (or similar attacker) could
```
anoadragon453, line 1605:
```suggestion
dependence on untrusted data: a malicious homeserver (or similar attacker) could
```
nit

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1580,10 +1582,36 @@ ratchet index that they have already decrypted. Care should be taken in
...
+change these values, and other devices/users can simply lie about them too.
+
+We can remove the fields, particularly the `sender_key`, because the `session_id`
+is already globally unique, therefore making storage and lookup possible without
+the need for added context from the `sender_key` or `device_id`.
```
anoadragon453, line 1610:
```suggestion
We can remove the fields, particularly the `sender_key` as the `session_id`
is already globally unique. Storage and lookup is therefore possible without
the need for added context from the `sender_key` or `device_id`.
```

Sentence refactoring.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1596,6 +1624,13 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
+When a client is updating a session key in its store, the client MUST ensure:
+
+* that the updated session data comes from a trusted source, where a trusted source
+  could be the signatures on the session or comes from the user's session key backup;
```
anoadragon453, line 1630:
```suggestion
* that the updated session data comes from a trusted source, where a trusted source
  could be the signatures on the session or those from the user's session key backup;
```

anoadragon453 (commented):
Looks good overall! However I don't see a change here corresponding to this paragraph in MSC3700:

> When clients receive an encrypted event with an unknown session they will need to send a key request to all clients, rather than the device specified by `sender_key` and `device_id`. This is the current behaviour used by Element clients.

and am failing to find an existing paragraph in the spec that mentions this.

content/client-server-api/modules/end_to_end_encryption.md
turt2live: err, I think the proposed sentence is a bit inaccurate: the `sender_key` is being removed, not kept, so it's inaccurate to say it's "as" something. The `session_id` is something we already have.

content/client-server-api/modules/end_to_end_encryption.md
turt2live: applied manually

content/client-server-api/modules/end_to_end_encryption.md
turt2live: applied manually

content/client-server-api/modules/end_to_end_encryption.md
dkasak: I think it was perhaps meant as "as" in the sense of "because" or "since", but removing the comma made it ambiguous.

turt2live: if there's not anything unclear about the current text then I'd prefer to keep it, for "single voice" reasons.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1596,6 +1624,13 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
+When a client is updating a Megolm session (room key) in its store, the client MUST ensure:
+
+* that the updated session data comes from a trusted source, where a trusted source
+  could be the signatures on the session or those from the user's session key backup;
+  and
```
nico-famedly, line 1631: I think this is not quite right. The source should have an equal or higher trust, but the online key backup for example is not a trusted source, since anyone can upload sessions to it.

turt2live: I mean, this is straight from the MSC almost to the letter. It's possible it's considering it trusted because if the key works to decrypt the message then it's clearly valid enough, however that doesn't necessarily solve the concern.

Also, I don't think anyone except the user holding the keys to the backup can write to it? Smells like a security issue that shouldn't be discussed here if people can arbitrarily write to key backups.

dkasak: It's publicly known; see the intro to the [Symmetric Megolm Backup MSC](https://github.com/uhoreg/matrix-doc/blob/symmetric-backups/proposals/3270-symmetric-megolm-backup.md#msc3270-symmetric-megolm-backup).

I'm guessing Erik just wasn't aware of this when drafting the MSC and no one caught it. In the future we'll have symmetric backups which shouldn't have this problem.

turt2live: I suppose that means we drop the key backup from here? Not really sure what the conclusion is...

dkasak: Perhaps let's just say the following, without giving an example for now:

> that the updated session data comes from a trusted source

anoadragon453: > Perhaps let's just say the following, without giving an example for now:
>> that the updated session data comes from a trusted source

This sounds reasonable to me.

anoadragon453 (approved):

PR #1102, opened 2022-05-31
Unify & standardize the v1.3 changelog

turt2live:
* Reference MSCs where MSCs were not being referenced.
* Alter language to appear consistent and from a single voice.
* Bundle and group various changes together (will affect the final changelog - the rendered one still doesn't bundle appropriately).
* Move entries to the spec area they are intended to be in.

**This change does not have a changelog entry itself because it's not actually changing anything.**


anoadragon453 (approved):
All looks reasonable to me!

Issue #1048, opened 2022-05-05
Spec leaves wiggle room in the rules surrounding `m.room.create` events


DMRobertson:
**Link to problem area**: https://spec.matrix.org/v1.2/rooms/v9/#authorization-rules

**Issue**

The auth rules begin:
 
> 1. If type is m.room.create:
>     1. If it has any previous events, reject.
>     2. If the domain of the room_id does not match the domain of the sender, reject.
>     3. If content.room_version is present and is not a recognised version, reject.
>     4. If content has no creator field, reject.
>     5. Otherwise, allow.

I understand this to mean that `m.room.create` events:
- MUST NOT have any `prev_events`.
- (implicitly) MUST have a `sender`, a `room_id` and a `content` field, each with appropriate types.
  - this is vaguely guaranteed by the S-S API's [checks on PDUs](https://spec.matrix.org/v1.2/server-server-api/#checks-performed-on-receipt-of-a-pdu), which asserts that each incoming event "is a valid event". (See also #365.)
  - This presumably means that it compiles to the PDU format of the given room version, and additionally complies to the room-agnostic specification of [`m.room.create`'s content, defined in the C-S API spec](https://spec.matrix.org/v1.2/client-server-api/#mroomcreate).
- (implicitly): MUST have a `room_id` and `sender` which are matrix IDs.
  - Again, arguably covered by "the event is valid". 
- MUST have matching domains for the `room_id` and `sender`.
- MUST have a `content.room_version` field which represents a recognised room version.
- MUST have a `content.creator` field.

I can see the following holes and problems n this definition.

1. There is no requirement that `content.creator` is an MXID, nor that its domain matches that of the `room_id` or `sender`. Come to think of it, why is the `content.creator` required at all, when the event already contains a `sender`? Should servers consider the event invalid if the creator and sender disagree?
2. This doesn't mention `auth_events` at all. Should servers ignore any auth events on `m.room.create`, or reject such events? 
    - I don't think it makes sense to include any auth_events here at all. Maybe you could make an argument for room upgrades, but there's a "Previous Room" field for that in the event's content.
3. Suppose `@alice:server` creates a room `!room:server`. What is there to stop `@bob:server` from trying to create a new room with the same ID?
    - If so, how should homeservers processing these two events decide who wins? (`origin_server_ts`, tiebreaking lexicographically on the creator's MXID?)
[...]

PR #1103, opened 2022-05-31
Add `m.federate` into the auth rules

neilalexander:
This is apparently something that we've been doing since the beginning of time in both Synapse and Dendrite and somehow it never made it into the spec.

There's a discussion point here though, as Synapse and Dendrite are doing very slightly different things. Synapse compares the event sender domain to the domain part of the `room_id` field in the create event, whereas Dendrite compares to the domain part of the `sender` in the create event.

Happy to take advice on whether we think it's better to spec `sender` vs `sender` (on the basis that they should never have been different anyway) or whether we want to spec `sender` vs `room_id` (on the basis of being exactly what Synapse does).

### DIFF
--- a/content/rooms/fragments/v1-auth-rules.md
+++ b/content/rooms/fragments/v1-auth-rules.md
@@ -34,11 +34,14 @@ The rules are as follows:
         algorithm described in the server specification.
 3.  If event does not have a `m.room.create` in its `auth_events`,
     reject.
-4.  If type is `m.room.aliases`:
+4.  If the create event content has the field `m.federate` set to `false`
+    and the sender domain of the event does not match the sender domain of
+    the create event, reject.
+5.  If type is `m.room.aliases`:
     1.  If event has no `state_key`, reject.
     2.  If sender's domain doesn't matches `state_key`, reject.
     3.  Otherwise, allow.
-5.  If type is `m.room.member`:
+6.  If type is `m.room.member`:
     1.  If no `state_key` key or `membership` key in `content`, reject.
     2.  If `membership` is `join`:
         1.  If the only previous event is an `m.room.create` and the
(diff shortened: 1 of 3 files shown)


turt2live (approved):
generally lgtm.

I don't think we need to worry too much about `sender` vs `room_id` or vs `sender` of the create event because the auth rules also state:

> 1. If type is `m.room.create`:
>     1. [...]
>     2. If the domain of the `room_id` does not match the domain of the `sender`, reject.

effectively: if you compare to either, you're covered by another rule. For the spec though, let's do what Synapse does as it further implies that an unfederated room is "on" the given server, rather than being implied through other auth rules.

richvdh (changes requested):
lgtm but please add it to https://github.com/matrix-org/matrix-spec/blob/main/content/rooms/v10.md too (sorry!)

neilalexander:
@richvdh Done!

richvdh (approved):
thanks!

content/rooms/fragments/v1-auth-rules.md
```diff
@@ -34,11 +34,14 @@ The rules are as follows:
         algorithm described in the server specification.
 3.  If event does not have a `m.room.create` in its `auth_events`,
     reject.
-4.  If type is `m.room.aliases`:
+4.  If the create event content has the field `m.federate` set to `false`
+    and the sender domain of the event does not match the sender domain of
+    the create event, reject.
```
richvdh, line 39: belatedly, a thought occurs to me: this is unclear about which create event we are talking about here. Do we mean the create event in the auth events (as mentioned on the line above), or the create event in the state at that point in the DAG (which is what the rest of the auth rules use).

I mean, if the two are different, we have a problem too, but we should really be explicit here.

richvdh: see also https://github.com/matrix-org/matrix-spec/issues/1048

PR #1110, opened 2022-06-07
room_type is not a required parameter in practice

nico-famedly:
In practice servers seem to mirror what the room create event does and
leave out the room_type when unset.

I don't think the intention of the spec was to have the room type required
on normal rooms and the text seems to match that too.

Signed-off-by: Nicolas Werner <n.werner@famedly.com>

Preview: https://pr1110--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1110.clarification
+++ b/changelogs/client_server/newsfragments/1110.clarification
@@ -1,0 +1,1 @@
+Fix various typos throughout the specification.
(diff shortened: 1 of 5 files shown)


clokep:
This seems to match the language in MSC2946:

> `room_type`: the value of the `type` field from the room's `m.room.create` event, if any.

Not sure if the server-server API doc needs to be updated too or if it re-uses this same chunk?

nico-famedly:
Yeah, it also matches the spec text (which has that exact text). Didn't check the server-server API, let me do that.

nico-famedly:
Yes, the server-server APi has that field as required too. It also lists allowed_roomids and such as required, which contradicts the spec text as well, let me update that.

nico-famedly:
Done. room_type and allowed_roomids are optional according to the text a few lines above the field marking them as required.

turt2live (approved):
Thanks!

PR #1114, opened 2022-06-11
Fix typo in relations endpoint description

zecakeh:
English is not my native language but I'm pretty sure it's grammatically incorrect otherwise.

### DIFF
--- a/data/api/client-server/relations.yaml
+++ b/data/api/client-server/relations.yaml
@@ -286,7 +286,7 @@ paths:
       summary: Get the child events for a given parent event, with a given `relType` and `eventType`.
       description: |-
         Retrieve all of the child events for a given parent event which relate to the parent
-        using the given `relType` and have the given `eventType`.
+        using the given `relType` and having the given `eventType`.
 
         Note that when paginating the `from` token should be "after" the `to` token in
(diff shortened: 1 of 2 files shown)


data/api/client-server/relations.yaml
```diff
@@ -286,7 +286,7 @@ paths:
       summary: Get the child events for a given parent event, with a given `relType` and `eventType`.
       description: |-
         Retrieve all of the child events for a given parent event which relate to the parent
-        using the given `relType` and have the given `eventType`.
+        using the given `relType` and having the given `eventType`.
```
KitsuneRal, line 289: I don't think this is a correct change. The original meaning: "Retrieve (...) child events (...) which relate to the parent (...) and have the given `eventType`". Your text changes it to: "Retrieve (...) child events (...) which relate to the parent: 1) using the given `relType` and 2) having the given `eventType`". Having the given eventType does not pertain to the relation, it's a trait of a child event.

KitsuneRal (changes requested):
See below.

data/api/client-server/relations.yaml
zecakeh: Ah right, I didn't understand the sentence like that. I guess that no change is needed then and I should close this.

PR #1131, opened 2022-06-16
Quick hacky version dropdown

### LINKED ISSUE #951: Version selector on spec.matrix.org
**Link to problem area**:
https://spec.matrix.org

**Issue**
there are /unstable, /latest, and presumably more. It would be nice if there was a version switcher selector dropdown or similar somewhere easy to find on the page. All versions != latest should have a note in the header that points to latest == stable.

DMRobertson:
Very rough prototype. Corners cut everywhere. Bad javascript habits. I learned babby's first web development back when the `<font>` tag was cool.

Closes #951. (Sort of. We'd need to regenerate /v1.1 and /v1.2 to include the switcher.

- In conf.toml, define a list of historical version strings.
- Use these to generate a `select` element in the navbar.
  - Mark the current version of the spec as the selected option.
  - Include a `noscript` fallback without the selector.
- Add a piece of Javascript which does the following onDOMContentLoaded:
  - Look at the URL for a path segment of the form `/v1.1/` or `/latest/`.
  - If we can't find this, remove the dropdown and replace it with the fallback content.
  - Add an event listener so that selecting a different version changes the version-specific path segment in the URL. No attempt is made to ensure that the new URL exists.
    - The "historical" section is special cased. Selecting it navigates you to the current spec's changelog page which explains about the `r0.x.y` versions.

Somewhat cribbed from what `docs.python.org` does, but written from scratch. (And much less robust.) The machinery for that lives [here](https://github.com/python/docsbuild-scripts/blob/main/templates/switchers.js), but I haven't dug too hard into their setup.

Preview: https://pr1131--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1131.feature
+++ b/changelogs/room_versions/newsfragments/1131.feature
@@ -1,0 +1,1 @@
+Dummy changelog to make CI happy. Add a very hacky version switcher to navigate between different spec versions.
--- a/config.toml
+++ b/config.toml
@@ -52,6 +52,9 @@ current_version_url = "https://spec.matrix.org/latest"
 #major = "1"
 #minor = "2"
 #release_date = "February 02, 2022"
+# Released versions to show in the version dropdown. Th template will also include a "historical" entry,
+# and (if the status above is "stable") an "unstable" entry.
+history = [ "unstable", "v1.2", "v1.1", "historical" ]
 
 # User interface configuration
 [params.ui]
--- a/layouts/partials/hooks/body-end.html
+++ b/layouts/partials/hooks/body-end.html
@@ -15,3 +15,4 @@
 */}}
 
 <script defer language="javascript" type="text/javascript"  src="{{ "js/toc.js" | urlize | relURL }}"></script>
+<script defer language="javascript" type="text/javascript"  src="{{ "js/version_selector.js" | urlize | relURL }}"></script>
--- a/layouts/partials/navbar.html
+++ b/layouts/partials/navbar.html
@@ -8,9 +8,10 @@
 
 {{ $cover := .HasShortcode "blocks/cover" }}
 <nav class="js-navbar-scroll navbar navbar-expand navbar-light {{ if $cover}} td-navbar-cover {{ end }}flex-column flex-md-row td-navbar">
-        <a class="navbar-brand" href="{{ .Site.Home.RelPermalink }}">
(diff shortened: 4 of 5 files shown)


DMRobertson:
I emulated the `spec.matrix.org` deployment as follows:

```
git branch -D dmr/v1.2
git checkout -b dmr/v1.2 v1.2
git cherry-pick dmr/version-switcher
# resolve merge and complete the cherry-pick

git branch -D dmr/v1.1
git checkout -b dmr/v1.1 v1.1
git cherry-pick dmr/version-switcher
# resolve merge and complete the cherry-pick

git switch dmr/version-switcher
hugo --baseURL /unstable -d spec/unstable
git switch dmr/v1.2
hugo --baseURL /v1.2 -d spec/v1.2
hugo --baseURL /latest -d spec/latest
git switch dmr/v1.1
hugo --baseURL /v1.1 -d spec/v1.1

python -m http.server --directory spec
```

DMRobertson:
Will leave this parked for now, but can come back to it if there's interest.

DMRobertson:
> I'm interested. Can you give it a newsfragment so that the CI runs and we can see what it looks like?

Part of the trouble is that Netlify only builds one version of the spec, rather than all the versions (`/v1.1/, /v1.2/, /latest/, /unstable/`) hosted on spec.matrix.org

DMRobertson:
Here's a demo I recorded locally. Unfortunately OBS studio wasn't capturing the dropdown menu that pops up, but it's hopefully convincing enough.

https://user-images.githubusercontent.com/8614563/174054878-0acd7cab-be32-4bc2-86d4-353b0f639727.mp4

turt2live:
that video looks good enough imo, and the code is good enough too. What are the concerns keeping this in draft?

DMRobertson:
> What are the concerns keeping this in draft?

Nothing as such, other than it being a big hack. It could use some hands-on testing from others too.

I'm not sure how the spec gets deployed. This would need a new version of v1.1 and v1.2 (and now v1.3) to be built which includes the selector.

turt2live:
We can try to build v1.1, v1.2, and v1.3 again though if it ends up not working out then ohwell.jpg - we already don't have links for r0.5.0 (for example).

richvdh:
Correct me if I'm wrong, but does this mean that old spec versions will not include newer spec versions on the dropdown, unless we rebuild all previous spec versions on each release? That doesn't sound entirely tenable.

Can we populate the list of of versions from a separate `https://spec.matrix.org/versions.json` or something?

richvdh (changes requested):
ok, I don't think rebuilding all past versions of the spec on each release is something we particularly want to do.

Can you make the javascript populate the dropdown from a json file at `/versions.json` or something?

PR #1119, opened 2022-06-13
Note a more efficient computation of the auth difference.

DMRobertson:
Closes #1118; see that issue for rationale.

Preview: https://pr1119--matrix-spec-previews.netlify.app

PR #1132, opened 2022-06-16
Revert "Note a more efficient computation of the auth difference. (#1119)"

DMRobertson:
This reverts commit a707266e50a770a2bdb272600e09671fe6ca8e41.

The final sentence was incorrect. The rest of the changes were fine.

Given that 1.3 is due to be cut imminently, the most expedient thing to do is revert the whole thing.

Preview: https://pr1132--matrix-spec-previews.netlify.app


turt2live (approved):
landing without changelog given the no-op

Issue #1136, opened 2022-06-17
Auth rules are unclear about which checks apply to auth events and which to room state


richvdh:
Per https://spec.matrix.org/v1.3/server-server-api/#checks-performed-on-receipt-of-a-pdu, we check the auth rules three times for each event, based on different subsets of state:
 * based on the event’s auth events
 * based on the state before the event
 * based on the current state of the room

However, some of the auth rules *always* apply to the auth events, and some are completely independent of any room state input. In particular, per https://spec.matrix.org/v1.3/rooms/v10/#authorization-rules:
 * Rule 1 ("If type is `m.room.create`:") is completely independent of the room state.
 * Rule 2 ("Reject if event has `auth_events` that:") and rule 3 ("If event does not have a `m.room.create` in its `auth_events`, reject.") always look at the auth events, never the room state.
 * Rule 4 ("If the create event content has the field `m.federate` set to false...") is unclear, but since there should be only one create event in a room which is constant throughout the tests, we can consider it as always looking at the auth events.

The rest of the rules are intended to apply based on the room state based on the three different state sets mentioned above - but do not say as much.

This is somewhat confusing for the reader, but also makes a faithful implementation rather inefficient, because if we implement it literally we'll check the first four rules three times for exactly the same input.

I've recently (https://github.com/matrix-org/synapse/pull/13065) split the two sets of checks into two functions in Synapse, and rather wonder if we should do the same in the spec.

richvdh:
> Rule 4 ("If the create event content has the field m.federate set to false...") is unclear, but since there should be only one create event in a room which is constant throughout the tests, we can consider it as always looking at the auth events.

I've attempted to clarify this in #1137, ftr

PR #1137, opened 2022-06-17
events with rejected_auth_events must be rejected

richvdh:
This might be kinda obvious, but didn't seem to be spelt out anywhere.

Preview: https://pr1137--matrix-spec-previews.netlify.app

### DIFF
--- a/content/rooms/fragments/v1-auth-rules.md
+++ b/content/rooms/fragments/v1-auth-rules.md
@@ -37,14 +37,14 @@ The rules are as follows:
         performed on receipt of a
         PDU](server-server-api/#checks-performed-on-receipt-of-a-pdu), reject.
     4. If there is no `m.room.create` event among the entries, reject.
-    5. If the `m.room.create` event content has the field `m.federate` set to `false`
-       and the `sender` domain of the event does not match the `sender` domain of
-       the create event, reject.
-3.  If type is `m.room.aliases`:
+3. If the `content` of the `m.room.create` event in the room state has the
+   property `m.federate` set to `false`, and the `sender` domain of the event
+   does not match the `sender` domain of the create event, reject.
+4.  If type is `m.room.aliases`:
     1.  If event has no `state_key`, reject.
     2.  If sender's domain doesn't matches `state_key`, reject.
     3.  Otherwise, allow.
-4.  If type is `m.room.member`:
+5.  If type is `m.room.member`:
     1.  If no `state_key` key or `membership` key in `content`, reject.
     2.  If `membership` is `join`:
         1.  If the only previous event is an `m.room.create` and the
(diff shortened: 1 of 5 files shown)


content/rooms/fragments/v1-auth-rules.md
```diff
@@ -26,22 +26,25 @@ The rules are as follows:
...
-    the create event, reject.
-5.  If type is `m.room.aliases`:
+        algorithm described in the server specification, reject.
+    3.  If there are entries which were themselves rejected under the [checks
+        performed on receipt of a
+        PDU](server-server-api/#checks-performed-on-receipt-of-a-pdu), reject.
+    4. If there is no `m.room.create` event among the entries, reject.
+    5. If the `m.room.create` event content has the field `m.federate` set to `false`
+       and the `sender` domain of the event does not match the `sender` domain of
+       the create event, reject.
```
turt2live, line 42: is this really a check on the `auth_events` or is it a check on the event being inspected?

turt2live: multiplied by however many times it shows up in this PR

richvdh: I don't understand the question. It's a check on the `m.room.create` event which is referred to in the `auth_events` of the event being inspected.

turt2live: shouldn't the server be using the `m.room.create` event from the room rather than `auth_events` though? It feels more understandable to move this point back up a level rather than nest it unmodified into the `auth_events` checks.

richvdh: possibly, but that seems to be a change to its meaning. Currently we have:

3. If event does not have a `m.room.create` in its `auth_events`, reject.
4. If the create event...

It's not entirely clear which create event it is talking about in 4, but given the context from 3, surely it's the one from the `auth_events`.

This is all related to #1136.

richvdh: hrrrm, actually, looking at the source to Synapse, it does appear to apply the test to the create event in the room state, so I guess that's an argument that we should make that explicit.

richvdh: Ok, I've updated this to make it apply to the room state, rather than the auth events.

content/rooms/fragments/v1-auth-rules.md
```diff
@@ -26,22 +26,25 @@ The rules are as follows:
...
-4.  If the create event content has the field `m.federate` set to `false`
-    and the sender domain of the event does not match the sender domain of
-    the create event, reject.
-5.  If type is `m.room.aliases`:
+        algorithm described in the server specification, reject.
+    3.  If there are entries which were themselves rejected under the [checks
+        performed on receipt of a
+        PDU](server-server-api/#checks-performed-on-receipt-of-a-pdu), reject.
+    4. If there is no `m.room.create` event among the entries, reject.
+3. If the `content` of the `m.room.create` event in the room state has the
```
turt2live, line 40: we might want to clarify which room state (at or before the event in question?), though hopefully for the create event it doesn't matter.

turt2live: applies to other snippets too

turt2live (approved):
lgtm either way, but would feel better about having *which* state explicit.

content/rooms/fragments/v1-auth-rules.md
richvdh: well, it's the same state that all the other auth rules refer to. Per https://spec.matrix.org/v1.3/server-server-api/#checks-performed-on-receipt-of-a-pdu, we actually check them at three different states.

I agree all this could be clarified, but I don't think it's the job of this PR.

PR #1140, opened 2022-06-22
Update idp-brands.md

kloenk:
update apple link


turt2live (changes requested):
This link appears to be a 404?

Kloenk:
All the links have the `.` directly after the link. I did not touch that as the other links do the same. could that be the problem?

turt2live (approved):
turns out reviewing PRs on mobile causes you to miss important details like half the URL.

PR #1155, opened 2022-06-30
Describe return codes for account data endpoints + minor clarifications

### LINKED ISSUE #1153: Lack of comprehensive status codes/error conditions in account data endpoints
**Link to problem area**:

- https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3useruseridaccount_datatype
- https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3useruseridaccount_datatype
- https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv3useruseridroomsroomidaccount_datatype
- https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3useruseridroomsroomidaccount_datatype

**Issue**

- What happens if you try to GET or PUT account data for a different user? Synapse returns 403 with M_FORBIDDEN.
- What happens if you try to GET or PUT account data to a user_id that isn't a matrix id? 403 probably suffices, but 400 might also be appropriate.
- What happens if you try to GET a `type` of account data which isn't stored on the server? Synapse returns 404.
- Synapse also special-cases an attempt to PUT `m.fully_read`:,returning 405 (Method not allowed).
  - https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/rest/client/account_data.py#L109-L114
  - Ahh, this is mandated by https://spec.matrix.org/v1.3/client-server-api/#server-behaviour-10

Synapse implementation is https://github.com/matrix-org/synapse/blob/e24ff8ebe3d4119d377355402245947f7de61c00/synapse/rest/client/account_data.py

More generally, see #1145

DMRobertson:
See individual commits.

Closes #1153.

https://pr1155--matrix-spec-previews.netlify.app/client-server-api/#client-config

Preview: https://pr1155--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/account_data.md
+++ b/content/client-server-api/modules/account_data.md
@@ -7,9 +7,9 @@ type: module
 Clients can store custom config data for their account on their
 homeserver. This account data will be synced between different devices
 and can persist across installations on a particular device. Users may
-only view the account data for their own account
+only view the account data for their own account.
 
-The account\_data may be either global or scoped to a particular rooms.
+The account data may be either global or scoped to a particular room.
 
 #### Events
 
--- a/content/client-server-api/modules/account_data.md
+++ b/content/client-server-api/modules/account_data.md
@@ -10,7 +10,9 @@ and can persist across installations on a particular device. Users may
 only view the account data for their own account.
 
 The account data may be either global or scoped to a particular room.
-There is no inheritance mechanism here: a given `type` of data missing from a room's account data does not fall back to the global account data with the same `type`.
+There is no inheritance mechanism here: a given `type` of data missing
+from a room's account data does not fall back to the global account
+data with the same `type`.
 
 #### Events
 
--- a/content/client-server-api/modules/account_data.md
+++ b/content/client-server-api/modules/account_data.md
@@ -10,6 +10,7 @@ and can persist across installations on a particular device. Users may
 only view the account data for their own account.
 
 The account data may be either global or scoped to a particular room.
+There is no inheritance mechanism here: a given `type` of data missing from a room's account data does not fall back to the global account data with the same `type`.
 
 #### Events
 
--- a/data/api/client-server/account-data.yaml
+++ b/data/api/client-server/account-data.yaml
@@ -71,6 +71,9 @@ paths:
             application/json: {}
           schema:
             type: object
+        400:
+          description: |-
+            The request body is not a JSON object.
         403:
           description: |-
             The access token provided is not authorized to modify this user's account 
@@ -174,6 +177,9 @@ paths:
             application/json: {}
           schema:
             type: object
+        400:
+          description: |-
+            The request body is not a JSON object.
         403:
           description: |-
             The access token provided is not authorized to modify this user's account 
--- a/data/api/client-server/account-data.yaml
+++ b/data/api/client-server/account-data.yaml
@@ -214,7 +214,8 @@ paths:
             type: object
         400:
           description: |-
-            The request body is not a JSON object.
+            The request body is not a JSON object, or the given `roomID` is not a valid
+            room ID.
           examples:
             application/json: {
               "errcode": "M_BAD_JSON",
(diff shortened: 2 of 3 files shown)


content/client-server-api/modules/account_data.md
```diff
@@ -7,9 +7,10 @@ type: module
 Clients can store custom config data for their account on their
 homeserver. This account data will be synced between different devices
 and can persist across installations on a particular device. Users may
-only view the account data for their own account
+only view the account data for their own account.
 
-The account\_data may be either global or scoped to a particular rooms.
+The account data may be either global or scoped to a particular room.
+There is no inheritance mechanism here: a given `type` of data missing from a room's account data does not fall back to the global account data with the same `type`.
```
turt2live, line 13: Line wrapping

turt2live (dismissed):
generally this lgtm, though the errors need to reference the error schema and ideally have examples set for clarity/rendering in the spec.

DMRobertson:
> generally this lgtm, though the errors need to reference the error schema and ideally have examples set for clarity/rendering in the spec.

This should be done now. Error messages are cribbed from Synapse.

[f58ea57](https://github.com/matrix-org/matrix-spec/pull/1155/commits/f58ea57ca99943c7e54dce96433fc7bec83f48f4) scope-creeps this a little by suggesting that bad room ids should also be rejected with a 400 status code. (Synapse doesn't do this yet, but I'm about to PR a change making that so.) This caused me to notice #1157.

data/api/client-server/account-data.yaml
```diff
@@ -71,6 +71,38 @@ paths:
             application/json: {}
           schema:
             type: object
+        400:
+          description: |-
+            The request body is not a JSON object.
+          examples:
+            application/json: {
+              "errcode": "M_BAD_JSON",
```
richvdh, line 79: using the example to specify the errcode is far from ideal.

I'm not entirely sure how best to handle this, but one potential approach is to follow the example of [`POST /_matrix/client/v3/rooms/{roomId}/invite`](https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3roomsroomidinvite) and just list the potential errcodes in the description, which also sidesteps https://github.com/matrix-org/matrix-spec/issues/1157.

wdyt?

DMRobertson: sgtm, thanks

data/api/client-server/account-data.yaml
```diff
@@ -71,6 +71,38 @@ paths:
             application/json: {}
           schema:
             type: object
+        400:
+          description: |-
+            The request body is not a JSON object. Errcode: `M_BAD_JSON`.
```
richvdh, line 76: hum. https://spec.matrix.org/v1.3/client-server-api/#common-error-codes says we should use `M_NOT_JSON` in this case. If synapse is using `M_BAD_JSON`, I'd say that's a synapse bug. (And yes, those are terrible names for the error codes. Don't blame me.)

(similarly for `PUT /user/{userId}/rooms/{roomId}/account_data/{type}`)

richvdh (changes requested):

data/api/client-server/account-data.yaml
DMRobertson: Judging from https://github.com/matrix-org/synapse/blob/33abbc327813e65aaa91e10f98a31622c045004c/synapse/http/servlet.py#L671-L694 it looks like Synapse is being naughty.

DMRobertson: Actually, now I'm not sure if this change is correct. Synapse hits that code when it's parsed a JSON value that isn't an object (so list, bool, null, number or string). They're arguably all "valid JSON", just not the form of JSON that we were expecting?

richvdh: oh right, yes, I'd expect `M_BAD_JSON` for a list where we expect an object. Which means that this endpoint can return either `M_BAD_JSON` or `M_NOT_JSON`.

DMRobertson: See [b66b199](https://github.com/matrix-org/matrix-spec/pull/1155/commits/b66b1994a0641d7fe8b9cf5978b0a411f580e694)

PR #1158, opened 2022-07-01
Fix unintentional stateres change added in #1042

### LINKED ISSUE #1154: #1042 mistakenly changed the description of state res in v2--v10 rooms
**Link to problem area**:
- https://spec.matrix.org/v1.3/rooms/v9/#algorithm versus https://spec.matrix.org/v1.2/rooms/v9/#algorithm , steps 1
- #1042 

**Issue**

In v1.2 and prior:
> Take all power events and any events in their auth chains, recursively, **that appear in the full conflicted set** and order them by the reverse topological power ordering.

In v1.3:
> Select all power events that appear in the full conflicted set. **Compute the union of their auth chains, including the power events themselves.** Sort the union using the reverse topological power ordering.

(emphasis mine).

These are not equivalent. In v1.2 one only selects events in the full conflicted set. In the wording of v1.3, one can pull in auth events from outside the full conflicted set.

I still think the v1.2 text is confusing (the word `recursively`, in particular). I think we should either

- revert to v1.2 wording
- find a better way to express v1.2's wording. Perhaps:
  > Select the set $X$ of all power events in the full conflicted set. For each such power event $P$, enlarge $X$ by adding the events in the auth chain of $P$ which also belong to the full conflicted set. Order $X$ using the reverse topological power ordering.

---

I don't know how the spec release process would handle this (I'm pretty sure I read somewhere that spec releases are meant to be immutable?). Maybe a v1.3.1?

cc @erikjohnston.

DMRobertson:
Fixes #1154.

Preview: https://pr1158--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1158.clarification
+++ b/changelogs/room_versions/newsfragments/1158.clarification
@@ -1,0 +1,1 @@
+For room versions 2–10: correct a mistaken clarification to the state resolution algorithm.
--- a/content/rooms/fragments/v2-state-res.md
+++ b/content/rooms/fragments/v2-state-res.md
@@ -138,9 +138,11 @@ the auth event is not rejected.
 
 The *resolution* of a set of states is obtained as follows:
 
-1.  Select all *power events* that appear in the *full conflicted set*. Compute
-    the union of their auth chains, including the power events themselves.
-    Sort the union using the *reverse topological power ordering*.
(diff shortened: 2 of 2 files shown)


turt2live:
@erikjohnston @richvdh I think you'll be best positioned to review this.

erikjohnston (approved):

PR #1162, opened 2022-07-04
Present OpenAPI `additionalProperties` field

DMRobertson:
When looking at the "Authentication Data" type used in e.g. the account
deactivation endpoint

    https://spec.matrix.org/v1.3/client-server-api/#post_matrixclientv3accountdeactivate

it was not clear to me that clients and servers should expect additional
fields in this structure (the semantics of which are implied by the
`type` of the authentication, or that of the previously established
`session`.)

In the OpenAPI spec, We occasionally mark some ojects as allowing
arbitrary additional properties (`additionalProperties: true`, or
`additionalProperties: { description: "..."}`). In other places we are
more strict and provide a schema that additional properties must satisfy.
In this PR we aim to make the first kind of additional
properties (non-strict) more visible in the rendered spec.

See https://pr1162--matrix-spec-previews.netlify.app/client-server-api/#post_matrixclientv3accountdeactivate

image

Preview: https://pr1162--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -14,6 +14,7 @@
 {{ $caption := .caption }}
 {{ $properties := .properties}}
 {{ $required := .required}}
+{{ $additionalProperties := .additionalProperties }}
 
 {{ if $properties }}
 
@@ -93,6 +94,29 @@
 
     {{ end }}
 
+   {{/* NB: the OpenAPI spec says "Consistent with JSON Schema, additionalProperties
+         defaults to true." Our schemas tend to omit additionalProperties; I think we
+         effectively assume it defaults to false. Therefore: we only output regarding
+         additionalProperties if it is explicitly set (i.e. is non-nil). */}}
+   {{ if (ne $additionalProperties nil )}}
+  <tr class="row-additional-properties">
+   <td colspan="3">
+        {{ if (and (reflect.IsMap $additionalProperties)) -}}
+            {{ if (index $additionalProperties "description") -}}
+May have additional properties: {{ $additionalProperties.description }}
+            {{ end }}
+        {{/* TODO: should we handle the case where additional properties are permitted,
+                    but must follow an explicit schema? */}}
+        {{ else if $additionalProperties -}}
+May have additional properties.
+        {{ else if (not $additionalProperties) -}}
+        {{/* At present we don't have any schema with additionalProperties: false.
+        But if we ever do, let's support it. */}}
+May **not** have additional properties.
+        {{ end }}
+   </td>
+  </tr>
+    {{ end }}
 </table>
 
 {{ end }}
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -13,6 +13,15 @@
     In some cases (such as response body specifications) this isn't used, and
     instead properties have a `required` boolean attribute. We support this too.
 
+  * `additionalProperties`: one of
+
+    - `false` (no additional properties are permitted);
+    - `true` (additional properties pairs are permitted, with no constraints);
+    - an OpenAPI schema object [1] (additional properties are permitted, but 
+      with constraints on `property_data`); or
+    - `nil` (not specified in schema, use `true` as a default value).
+
+  [1]: https://swagger.io/specification/#schema-object
 */}}
 
 {{ $title := .title }}
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -103,7 +103,7 @@
    <td colspan="3">
         {{ if (and (reflect.IsMap $additionalProperties)) -}}
             {{ if (index $additionalProperties "description") -}}
-Notes on additional properties: {{ $additionalProperties.description }}
+<em>Notes on additional properties:</em> {{ $additionalProperties.description }}
             {{ end }}
             {{/* TODO: should we handle the case where additional properties are
                        permitted, but must follow an explicit schema? */}}
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -94,24 +94,21 @@
 
     {{ end }}
 
-   {{/* NB: the OpenAPI spec says "Consistent with JSON Schema, additionalProperties
-         defaults to true." Our schemas tend to omit additionalProperties; I think we
-         effectively assume it defaults to false. Therefore: we only output regarding
-         additionalProperties if it is explicitly set (i.e. is non-nil). */}}
+ {{/* Additional properties should either be an OpenAPI schema object, or a
+       boolean. If it's the former, look for a `description` and render it as a note.
+       If it's `true`, do nothing. If it's `false`, raise a build-time error:
+       "Everything is extensible in Matrix unless otherwise noted". */}}
    {{ if (ne $additionalProperties nil )}}
   <tr class="row-additional-properties">
    <td colspan="3">
         {{ if (and (reflect.IsMap $additionalProperties)) -}}
             {{ if (index $additionalProperties "description") -}}
-May have additional properties: {{ $additionalProperties.description }}
+Notes on additional properties: {{ $additionalProperties.description }}
             {{ end }}
-        {{/* TODO: should we handle the case where additional properties are permitted,
-                    but must follow an explicit schema? */}}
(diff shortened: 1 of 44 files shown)


layouts/partials/openapi/render-object-table.html
```diff
@@ -93,6 +94,29 @@
...
+May have additional properties: {{ $additionalProperties.description }}
+            {{ end }}
+        {{/* TODO: should we handle the case where additional properties are permitted,
+                    but must follow an explicit schema? */}}
+        {{ else if $additionalProperties -}}
+May have additional properties.
+        {{ else if (not $additionalProperties) -}}
+        {{/* At present we don't have any schema with additionalProperties: false.
+        But if we ever do, let's support it. */}}
+May **not** have additional properties.
```
turt2live, line 115: Everything is extensible in Matrix unless otherwise noted, so this is a bit misleading.

turt2live (commented):
generally this seems fine, though there seems to be a bit of confusion around what clients are able to send (which is anything it feels like - servers don't have to listen though)

layouts/partials/openapi/render-object-table.html
DMRobertson: I don't follow---what do you find misleading?

DMRobertson: Okay. Then perhaps this can be changed so that 

- we don't do anything if `additionalProperties: true`
- we raise a build-time error if `additionalProperties: false`
- Rather than "May have additional properties:" in the generated HTML, we write "Notes on additional properties:".

How does that sound?

DMRobertson: See https://swagger.io/specification/#schema-object

> The following properties are taken from the JSON Schema definition but their definitions were adjusted to the OpenAPI Specification.
> additionalProperties - Value can be boolean or object. Inline or referenced schema MUST be of a [Schema Object](https://swagger.io/specification/#schema-object) and not a standard JSON Schema. Consistent with JSON Schema, additionalProperties defaults to true.

DMRobertson: And for completeness: the original JSON schema additionalProperties are defined [here](https://json-schema.org/draft/2020-12/json-schema-core.html#additionalProperties).

DMRobertson: 3398cf0f is an attempt at this.

DMRobertson:
Oh. Looks like I have a merge to resolve.

layouts/partials/openapi/render-object-table.html
```diff
@@ -18,6 +18,7 @@
 {{ $title := .title }}
 {{ $properties := .properties}}
 {{ $required := .required}}
+{{ $additionalProperties := .additionalProperties }}
```
richvdh, line 21: please could you add something about this to the comment at the top of the file

layouts/partials/openapi/render-object-table.html
```diff
@@ -97,6 +98,26 @@
 
     {{ end }}
 
+ {{/* Additional properties should either be an OpenAPI schema object, or a
```
richvdh, line 101:
```suggestion
 {{/* additionalProperties should either be an OpenAPI schema object, or a
```

layouts/partials/openapi/render-object-table.html
```diff
@@ -97,6 +98,26 @@
...
+       boolean. If it's the former, look for a `description` and render it as a note.
+       If it's `true`, do nothing. If it's `false`, raise a build-time error:
+       "Everything is extensible in Matrix unless otherwise noted". */}}
+   {{ if (ne $additionalProperties nil )}}
+  <tr class="row-additional-properties">
+   <td colspan="3">
```
richvdh, line 107: it seems odd that we include the table row and empty cell if `additionalProperties` is true?

richvdh (changes requested):

layouts/partials/openapi/render-object-table.html
DMRobertson: Sure, see 6487e84c

layouts/partials/openapi/render-object-table.html
DMRobertson: Yes. A previous version of this PR used to have the cell say "May have additional properties" in that situation. Looks like I didn't clean up the HTML for when I removed that case.

layouts/partials/openapi/render-object-table.html
```diff
@@ -97,6 +107,27 @@
...
     {{ end }}
 
+ {{/* If additionalProperties is true, do nothing.
+      If additionalProperties is nil, treat is as the default (true) and do nothing.
+      If additionalProperties is false, raise a build-time error:
+      "Everything is extensible in Matrix unless otherwise noted".
+      Otherwise, it should be an OpenAPI schema object. Include the description
+      of that schema object, if it exists. */}}
+ {{ if (and (ne $additionalProperties nil) (ne $additionalProperties true)) }}
+  {{ if (and (reflect.IsMap $additionalProperties)) -}}
```
richvdh, line 117: redundant `and` here.

richvdh, line 117: can we restructure these `if` statements to reduce the nesting?

```
 {{ if (or (eq $additionalProperties nil) (eq $additionalProperties true)) }}
     {{/* do nothing */}}
  {{ else if (reflect.IsMap $additionalProperties) -}}
     ...
  {{ else if (not $additionalProperties) -}}
    ... 
  {{ end }}
```

or something.

layouts/partials/openapi/render-object-table.html
```diff
@@ -97,6 +107,27 @@
...
+  <tr class="row-additional-properties">
+   <td colspan="3">
+    <em>Notes on additional properties:</em> {{ $additionalProperties.description }}
+   </td>
+  </tr>
+   {{ end }}
+  {{/* TODO: should we handle the case where additional properties are
+             permitted, but must follow an explicit schema? */}}
+  {{ else if (not $additionalProperties) -}}
+   {{ errorf "Unexpected additionalProperties=false for %s" $title }}
```
richvdh, line 128: it's unclear to me that we should be raising an error in this case. Isn't it legitimate to specify that a given object may *not* have additional properties, even if that is currently unused?

richvdh (changes requested):

layouts/partials/openapi/render-object-table.html
DMRobertson: :shrug: There's some context here: https://github.com/matrix-org/matrix-spec/pull/1162#discussion_r914153682 ; I don't have any strong opinions here.

layouts/partials/openapi/render-object-table.html
richvdh: @turt2live, @DMRobertson: I'm not sure I'm following the discussion here. Can I just check that everyone is clear what we're discussing?

The question is around the correct behaviour if we encounter an object schema where `additionalProperties` is explictly set to `false` (as opposed to leaving it unset, which is equivalent to setting it to `true`).

[Earlier drafts](https://datatracker.ietf.org/doc/html/draft-wright-json-schema-validation-00#section-5.18) of JSON Schema give a clear definition of what that means:

> If "additionalProperties" is false, validation succeeds only if the instance is an object and all properties on the instance were covered by "properties" and/or "patternProperties".

In other words: this is explicitly about trying to define an object which **may not be extended**.

I don't entirely *object* to the decision to emit an error message, if we think it's somehow un-matrixy to try to define an object that isn't allowed any additional properties.  I'd just like to be sure that we made that decision with good information.

KitsuneRal: We actually have that case (exactly one, unless I missed some weirdly formatted), and it's in [`m.receipt` events](https://github.com/matrix-org/matrix-spec/blob/a45138c3b4a722ef4f4eb98788dc25c94a3dce61/data/event-schemas/schema/m.receipt.yaml#L58): . `additionalProperties: false` is used next to `patternProperties` to forbid anything that does not follow the pattern. Which is quite valid and smart, in my opinion. There's a big question however on what the homeserver is supposed to do with this boundary. I would treat it as a "MUST ignore" of sorts, whereas omitting `additionalProperties` entirely would correspond to "MAY ignore". Does it make sense?

richvdh:
@DMRobertson are you still interested in pursuing this? Anything we can do to unblock you?

DMRobertson:
It'd be nice to do this at some point, but I don't think I'm going to find the time. Let's close this off for now.

Issue #17, opened 2022-02-01
spec for policy rules does not match reality


richvdh:
**Link to problem area**:

According to https://spec.matrix.org/v1.1/client-server-api/#events-19, within policy rules:

> Glob characters * and ? can be used to match zero or more and one or more characters respectively.

**Issue**

According to discussion at https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$-g-gRQFjxbsHNe2OpGf5yGGHVNsJcVCga-sF4WwnWOA?via=matrix.org&via=libera.chat&via=envs.net, this does not match what happens in mjolnir.

Policy rules were introduced by MSC2313, which says: "Simple globs are supported  for defining entities (`*` and `?` as wildcards, **just like `m.room.server_acl`**)". Mjolnir *does* interpret globs in the same way as server_acl - ie, they match zero or more, and exactly one, character respectively.

This therefore seems to have been a bug introduced in the spec by matrix-org/matrix-spec-proposals#2434.

**Expected behaviour**

The use of `?` to match one or more characters would be ... unusual. Since it's never been interpreted that way, changing the spec seems more sensible.

PR #1165, opened 2022-07-10
Fix wording for globs in policy lists

deepbluev7:
The specification here does neither match how globs work in common
libraries nor do they match how they are used in practice currently or
how the MSC worded them (which said they should be like server ACLs). As
such this seems to be an issue introduced when writing the spec text.

Ref mjolnir:
- https://github.com/matrix-org/mjolnir/blob/b48904bc2b4fcd636176b12dbe173ad651930f23/src/models/ListRule.ts#L44

Ref matrix bot sdk (which implements the glob used above):
- https://github.com/turt2live/matrix-bot-sdk/blob/473e563236dd6edb25e7bd18d3517d1a61e037a1/src/helpers/MatrixGlob.ts#L26
- https://github.com/turt2live/matrix-bot-sdk/blob/f799b1fe1a72b9a4a6053c50cedfb43bee962558/test/helpers/MatrixGlobTest.ts#L44

Ref original MSC:
- https://github.com/matrix-org/matrix-spec-proposals/blob/c7b3d998537d21694a166b4a6a4cf0490ebc0cc2/proposals/2313-moderation-policy-rooms.md?plain=1#L36
- https://spec.matrix.org/v1.3/client-server-api/#server-access-control-lists-acls-for-rooms

fixes https://github.com/matrix-org/matrix-spec/issues/17

### DIFF
--- a/content/client-server-api/modules/moderation_policies.md
+++ b/content/client-server-api/modules/moderation_policies.md
@@ -91,7 +91,7 @@ included in the ban list.
 #### Events
 
 The `entity` described by the state events can contain `*` and `?` to
-match zero or more and one or more characters respectively. Note that
+match zero or more characters and exactly one character respectively. Note that
 rules against rooms can describe a room ID or room alias - the
 subscriber is responsible for resolving the alias to a room ID if
 desired.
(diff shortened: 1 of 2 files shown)


turt2live (approved):
thanks!

PR #1169, opened 2022-07-13
Clarify the power levels integer range

johannescpk:
To help with discussions which range of power levels are allowed. Without the clarification it's sometimes assumed that the max power level is 100.

As discussed on matrix: https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$GUmiy799UErZCQ-yi_WJS0kMNdafrjRlSYllW7fTKss?via=matrix.org&via=libera.chat&via=element.io

Signed-off-by: Johannes Becker <j.becker@famedly.com>

Preview: https://pr1169--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/schema/m.room.power_levels.yaml
+++ b/data/event-schemas/schema/m.room.power_levels.yaml
@@ -5,6 +5,8 @@ description: |-
   This event specifies the minimum level a user must have in order to perform a
   certain action. It also specifies the levels of each user in the room.
 
+  The allowed range of power levels is specified by the [Canonical JSON specification](https://spec.matrix.org/v1.3/appendices/#canonical-json).
+
   If a `user_id` is in the `users` list, then that `user_id` has the
   associated power level. Otherwise they have the default level
   `users_default`. If `users_default` is not supplied, it is assumed to be
--- a/data/event-schemas/schema/m.room.power_levels.yaml
+++ b/data/event-schemas/schema/m.room.power_levels.yaml
@@ -5,8 +5,6 @@ description: |-
   This event specifies the minimum level a user must have in order to perform a
   certain action. It also specifies the levels of each user in the room.
 
-  The allowed range of power levels is specified by the [Canonical JSON specification](/appendices/#canonical-json).
-
   If a `user_id` is in the `users` list, then that `user_id` has the
   associated power level. Otherwise they have the default level
   `users_default`. If `users_default` is not supplied, it is assumed to be
@@ -43,6 +41,9 @@ description: |-
   Server implementations should therefore ensure that each room has an
   `m.room.power_levels` event as soon as it is created. See also the
   documentation of the `/createRoom` API.
+
+  Note: the allowed range for power level values is `[-(2**53)+1, (2**53)-1]`,
+  as specified by the [Canonical JSON specification](/appendices/#canonical-json).
 
 properties:
   content:
--- a/data/event-schemas/schema/m.room.power_levels.yaml
+++ b/data/event-schemas/schema/m.room.power_levels.yaml
@@ -58,8 +58,8 @@ properties:
         type: object
       events_default:
         description: |-
-          The default level required to send message events. Can be
-          overridden by the `events` key.  Defaults to 0 if unspecified.
+            The default level required to send message events. Can be
+            overridden by the `events` key.  Defaults to 0 if unspecified.
         type: integer
       invite:
         description: The level required to invite a user. Defaults to 0 if unspecified.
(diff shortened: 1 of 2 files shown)


data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -5,6 +5,11 @@ description: |-
   This event specifies the minimum level a user must have in order to perform a
   certain action. It also specifies the levels of each user in the room.
 
+  The range of power levels is specified by the matrix-specific Canonical JSON
+  specification (https://spec.matrix.org/v1.3/appendices/#canonical-json), which
```
turt2live, line 9:
```suggestion
  specification (/appendices/#canonical-json), which
```

data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -5,6 +5,11 @@ description: |-
...
+  allows a range of `[-(2**53)+1, (2**53)-1]` for integers. Leading zeros are
+  not allowed.
```
turt2live, line 11: I don't see where leading zeros are banned?

data/event-schemas/schema/m.room.power_levels.yaml
turt2live: also we should just use a markdown link over "Canonical JSON", as we do elsewhere. It doesn't need the "matrix-specific" qualifier.

turt2live (changes requested):
thanks for taking a look at this! My review is mostly nitpicky based on the surrounding spec, sorry.

data/event-schemas/schema/m.room.power_levels.yaml
johannescpk: Good point, fixed.

data/event-schemas/schema/m.room.power_levels.yaml
johannescpk: That's my interpretation of

> int       = %x30 / ( %x31-39 *digit )

in the [grammar](https://spec.matrix.org/v1.3/appendices/#grammar). But maybe I'm mistaken?

johannescpk: As per https://datatracker.ietf.org/doc/html/rfc7159#section-6

> Leading zeros are not allowed.

As the purpose of this PR is to clarify the range, I think the fact about leading zeros fits there - as the grammar itself explains it via the the `int = ` notation.

data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -5,6 +5,8 @@ description: |-
   This event specifies the minimum level a user must have in order to perform a
   certain action. It also specifies the levels of each user in the room.
 
+  The allowed range of power levels is specified by the [Canonical JSON specification](/appendices/#canonical-json).
```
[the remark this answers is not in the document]
richvdh: Also, this is true for every numeric in the protocol, and the current phrasing makes it sound like it's specific to PLs.

Also also: making people go and grok the canonical-json spec rather than just telling them the answer seems... unhelpful?

I would phrase it as something like:

> Note: the allowed range for power level values is `[-(2**53)+1, (2**53)-1]`, as specified by the [Canonical JSON specification](/appendices/#canonical-json).

richvdh (commented):
Thanks!

Could you add a changelog entry? A few words in `1169.clarification` in https://github.com/matrix-org/matrix-spec/tree/fad9feb80c804bc0c3d6c1cba2ee36f9be79daa2/changelogs/client_server/newsfragments will be fine.

data/event-schemas/schema/m.room.power_levels.yaml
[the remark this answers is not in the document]
johannescpk: I've removed the actual values because of https://github.com/matrix-org/matrix-spec/pull/1169#discussion_r923618961. Personally I'd also prefer the values, but I can also see that potentially outdated values might be seen as an issue?

Other than that, I'll move it to the end and look into rephrasing so it doesn't sound specific for PLs.

data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -42,6 +42,9 @@ description: |-
   `m.room.power_levels` event as soon as it is created. See also the
   documentation of the `/createRoom` API.
 
+  Note: the allowed range for power level values is `[-(2**53)+1, (2**53)-1]`,
+  as specified by the [Canonical JSON specification](/appendices/#canonical-json).
```
richvdh, line 46: lots of "specification"

```suggestion
  as required by the [Canonical JSON specification](/appendices/#canonical-json).
```

richvdh (approved):
Thanks! Sorry this took so long.

PR #1170, opened 2022-07-13
Add missing "added in" spec version to some key backup endpoints

turt2live:
These endpoints were all added in v1.1, but we forgot to annotate a few of them.

Some were annotated in https://github.com/matrix-org/matrix-spec-proposals/pull/3425

Preview: https://pr1170--matrix-spec-previews.netlify.app


turt2live:
probably, yes, but this was a lot easier to notice from a web editor 😇

PR #1172, opened 2022-07-15
Disable blank issue reporting

turt2live:
Folks who want to report a blank issue can still get at the form if they try hard enough, but in general everything should have a label to make triage easier.


turt2live:
no changelog because meta

richvdh (approved):
I might find this annoying, but let's give it a go :)

turt2live:
workaround is to chop `/choose` of the URL, ftr.

PR #1166, opened 2022-07-12
Move various e2e defintions out to yaml files

richvdh:
We have code to generate tables, which we should use in the e2e section.

Preview: https://pr1166--matrix-spec-previews.netlify.app

PR #1178, opened 2022-07-21
Reinstate a smaller margin between struct tables

DMRobertson:
Without a margin, things look poor if the last row is blue: it then
merges into the caption at the top of the next table. 

I've also added a small border because I think it helps to break things
up a little.

Before: Before
After: after

Xref: https://github.com/matrix-org/matrix-spec/commit/ea42cd3c7b40caae0b6925e1620173dd004e04a7#r79080321 and https://github.com/matrix-org/matrix-spec/pull/1166#discussion_r926890649

Preview: https://pr1178--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -323,12 +323,15 @@ footer {
   table {
     table-layout: fixed;
     width: 100%;
+    margin: 1.7rem 0;
+    border-bottom: 2px solid $gray-200;
 
     caption {
       caption-side: top;
       color: $dark;
       font-size: 1rem;
       font-weight: $font-weight-bold;
+      border-top: 2px solid $gray-200;
     }
 
     th, td, caption {
--- a/changelogs/client_server/newsfragments/1178.clarification
+++ b/changelogs/client_server/newsfragments/1178.clarification
@@ -1,0 +1,1 @@
+Improve the styling of adjacent `<table>`s rendered from OpenAPI definitions.
--- a/changelogs/server_server/newsfragments/1178.clarification
+++ b/changelogs/server_server/newsfragments/1178.clarification
@@ -1,0 +1,1 @@
+Improve the styling of adjacent `<table>`s rendered from OpenAPI definitions.


richvdh (commented):
looks like a reasonable compromise, I guess.

It does mean that there is a bit of a gap before, for example, the `AuthData` table at https://pr1178--matrix-spec-previews.netlify.app/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2. 

That lead me into a rabbithole of wondering why other definition tables *don't* have this problem, and that in turn led me to this bit of CSS. which overrides the margin for a table when it follows a hard-coded list of other elements: https://github.com/matrix-org/matrix-spec/blob/79e3b7e629ccdd8bb99f1ee9821b47dc4a9b20c9/assets/scss/custom.scss#L310-L312

I wonder if it would be better to flip this around and instead specify a margin *only* when a table is followed by another table? wdyt?

richvdh:
> I wonder if it would be better to flip this around and instead specify a margin only when a table is followed by another table? wdyt?

I've done this over at https://github.com/matrix-org/matrix-spec/pull/1192, since I was poking in the area.

> I've also added a small border because I think it helps to break things up a little.

I don't disagree, but it also adds a second line after the 'Requires authentication' line, which looks a bit odd. I'm working on  some other changes which I hope will make this easier.

DMRobertson:
Is this now superseded/made defunct by #1192?

> It does mean that there is a bit of a gap before, for example, the AuthData table at [pr1178--matrix-spec-previews.netlify.app/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2](https://pr1178--matrix-spec-previews.netlify.app/client-server-api/#backup-algorithm-mmegolm_backupv1curve25519-aes-sha2).

Arg, yes. Some more classes on the tables would help here.

> That lead me into a rabbithole of wondering why other definition tables _don't_ have this problem, and that in turn led me to this bit of CSS. which overrides the margin for a table when it follows an `h2` or `h3`:
> 
> https://github.com/matrix-org/matrix-spec/blob/79e3b7e629ccdd8bb99f1ee9821b47dc4a9b20c9/assets/scss/custom.scss#L310-L312
>
> I wonder if it would be better to flip this around and instead specify a margin only when a table is followed by another table? wdyt?

I'm kind of surprised we need those kind of selectors, given that [margin](https://www.w3.org/TR/CSS22/box.html#collapsing-margins) [collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) is A Thing. But I'm naive and my CSS knowledge is largely stuck in the previous decade. No strong opinions though.

richvdh:
> Is this now superseded/made defunct by https://github.com/matrix-org/matrix-spec/pull/1192?

I think so? But up to you to figure out if there's anything here that you still want.

> I'm kind of surprised we need those kind of selectors, given that [margin](https://www.w3.org/TR/CSS22/box.html#collapsing-margins) [collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) is A Thing. But I'm naive and my CSS knowledge is largely stuck in the previous decade.

You and me both. There may well be better ways to do this.

DMRobertson:
> But up to you to figure out if there's anything here that you still want.

Looking at your changes as rendered on https://spec.matrix.org/unstable/client-server-api/ I'm happy---the original complaint is fixed. Thanks!

PR #1180, opened 2022-07-21
Remove a stray fullstop in "Other error codes"

DMRobertson:
This has also been irritating me for ages

### DIFF
--- a/changelogs/client_server/newsfragments/1180.clarification
+++ b/changelogs/client_server/newsfragments/1180.clarification
@@ -1,0 +1,1 @@
+Fix a punctuation typo in the list of "Other error codes".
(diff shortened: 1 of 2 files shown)


changelogs/client_server/newsfragments/1180.clarification
```diff
@@ -0,0 +1 @@
+Fix a punctuation typo in the list of "Other error codes".
```
uhoreg, line 1:
```suggestion
Fix various typos throughout the specification.
```

uhoreg (approved):
lgtm, aside from a change to the changelog entry.  Using that changelog entry means that it will get merged with other typo-fix (and similar) entries in the changelog, rather than having a whole bunch of individual changes in the list.

PR #1183, opened 2022-07-26
CSS hack to fix scroll anchoring

richvdh:
Disable explicit `height` setting for the top-level div, which breaks scroll anchoring

Preview: https://pr1183--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -29,6 +29,12 @@ Custom SCSS for the Matrix spec
  */
 @import "syntax.scss";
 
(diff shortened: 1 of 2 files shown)


richvdh:
fixes: https://github.com/matrix-org/matrix-spec/issues/894

richvdh:
workaround for https://github.com/google/docsy/issues/1116

turt2live (approved):
I don't think this needs a changelog ftr

richvdh:
> I don't think this needs a changelog ftr

No, but I wanted the preview to work :)

I'll get rid of the changelog

PR #1185, opened 2022-07-26
Clarify the use of JSON in requests and responses

### LINKED ISSUE #1182: Spec documentation does not say that request bodies are required
The OpenAPI definitions all have `required: true` for request bodies (meaning that clients must provide an empty JSON object even if they don't want to set any of the parameters), but this isn't reflected in the words of the spec.

richvdh:
Fixes #1182

Preview: https://pr1185--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -13,20 +13,32 @@ clients which maintain a full local persistent copy of server state.
 ## API Standards
 
 The mandatory baseline for client-server communication in Matrix is
-exchanging JSON objects over HTTP APIs. HTTPS is recommended for
-communication, although HTTP may be supported as a fallback to support
-basic HTTP clients. More efficient optional transports will in future be
-supported as optional extensions - e.g. a packed binary encoding over
-stream-cipher encrypted TCP socket for low-bandwidth/low-roundtrip
-mobile usage. For the default HTTP transport, all API calls use a
-Content-Type of `application/json`. In addition, all strings MUST be
-encoded as UTF-8.
+exchanging JSON objects over HTTP APIs. More efficient transports may be
+specified in future as optional extensions.
+
+HTTPS is recommended for communication. The use of plain HTTP is not
+recommended outside test environments.
 
 Clients are authenticated using opaque `access_token` strings (see [Client
 Authentication](#client-authentication) for details).
 
+All `POST` and `PUT` endpoints, with the exception of
+[`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload), require the
+client to supply a request body containing a (potentially empty) JSON object.
+Clients are *not* required to supply a `Content-Type` header.
+
+Similarly, all endpoints require the server to return a JSON object,
+with the exception of 200 responses to
+[`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](#get_matrixmediav3downloadservernamemediaid)
+and [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](#get_matrixmediav3thumbnailservernamemediaid).
+Servers should include a `Content-Type` header of `application/json` for all JSON responses.
+
+All JSON data, in requests or responses, must be encoded using UTF-8.
+
 See also [Conventions for Matrix APIs](/appendices#conventions-for-matrix-apis)
-in the Appendices for conventions which all Matrix APIs are expected to follow.
+in the Appendices for conventions which all Matrix APIs are expected to follow,
+and [Web Browser Clients](#web-browser-clients) below for additional
+requirements for server responses.
 
 ### Standard error response
 
@@ -196,6 +208,8 @@ only read state (e.g.: `/sync`, get account data, etc).
 The user is unable to reject an invite to join the server notices room.
 See the [Server Notices](#server-notices) module for more information.
 
+### Transaction identifiers
+
 The client-server API typically uses `HTTP PUT` to submit requests with
 a client-generated transaction identifier. This means that these
 requests are idempotent. The scope of a transaction identifier is a
@@ -207,8 +221,6 @@ integer is recommended).
 Some API endpoints may allow or require the use of `POST` requests
 without a transaction ID. Where this is optional, the use of a `PUT`
 request is strongly recommended.
-
-{{% http-api spec="client-server" api="versions" %}}
 
 ## Web Browser Clients
 
@@ -307,6 +319,8 @@ specify parameter values. The flow for this method is as follows:
         `base_url` value, then `FAIL_PROMPT`.
 
 {{% http-api spec="client-server" api="wellknown" %}}
+
+{{% http-api spec="client-server" api="versions" %}}
 
 ## Client Authentication
 
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -32,7 +32,7 @@ Similarly, all endpoints require the server to return a JSON object,
 with the exception of 200 responses to
 [`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](#get_matrixmediav3downloadservernamemediaid)
 and [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](#get_matrixmediav3thumbnailservernamemediaid).
-Servers should include a `Content-Type` header of `application/json` for all JSON responses.
+Servers msut include a `Content-Type` header of `application/json` for all JSON responses.
 
 All JSON data, in requests or responses, must be encoded using UTF-8.
 
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -25,7 +25,8 @@ Authentication](#client-authentication) for details).
 All `POST` and `PUT` endpoints, with the exception of
 [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload), require the
 client to supply a request body containing a (potentially empty) JSON object.
-Clients are *not* required to supply a `Content-Type` header.
+Clients should supply a `Content-Type` header of `application/json` for all requests with JSON bodies,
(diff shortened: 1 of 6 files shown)


content/client-server-api/_index.md
```diff
@@ -196,6 +208,8 @@ only read state (e.g.: `/sync`, get account data, etc).
 The user is unable to reject an invite to join the server notices room.
 See the [Server Notices](#server-notices) module for more information.
 
+### Transaction identifiers
```
richvdh, line 211: Gave this a new heading to stop it being stuck under "Other error codes"

content/client-server-api/_index.md
```diff
@@ -208,8 +222,6 @@ Some API endpoints may allow or require the use of `POST` requests
 without a transaction ID. Where this is optional, the use of a `PUT`
 request is strongly recommended.
 
-{{% http-api spec="client-server" api="versions" %}}
```
richvdh, line 211: and moved this into "discovery" (for want of a better place), for the same reason.

content/identity-service-api.md
```diff
@@ -38,8 +38,20 @@ Appendix.
 
 The mandatory baseline for identity server communication in Matrix is
 exchanging JSON objects over HTTP APIs. HTTPS is required for
-communication, and all API calls use a Content-Type of
-`application/json`. In addition, strings MUST be encoded as UTF-8.
+communication.
+
+All `POST` and `PUT` endpoints, with the exception of [`POST
+/_matrix/identity/v2/account/logout`](#post_matrixidentityv2accountlogout),
```
richvdh, line 44: there's always an exception.

uhoreg: Why is `/account/logout` an exception to this?

richvdh: I suspect it's because nobody thought about it before, but: it just is, according to both the openapi definition, and the `sydent` implementation.

richvdh: (and https://spec.matrix.org/v1.3/identity-service-api/#authentication, which says "no ... request body")

content/client-server-api/_index.md
```diff
@@ -13,20 +13,32 @@ clients which maintain a full local persistent copy of server state.
...
+HTTPS is recommended for communication. The use of plain HTTP is not
+recommended outside test environments.
 
 Clients are authenticated using opaque `access_token` strings (see [Client
 Authentication](#client-authentication) for details).
 
+All `POST` and `PUT` endpoints, with the exception of
+[`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload), require the
+client to supply a request body containing a (potentially empty) JSON object.
+Clients are *not* required to supply a `Content-Type` header.
```
uhoreg, line 28: :( Can we at least say that clients "should" supply a `Content-Type: application/json` header?

richvdh: I'm not really sure if that gains us much, unless we're planning to mandate it sometime soon? Servers still have to assume it's json even if there is no content-type header (or, for that matter, if the content-type header says it is something other than json, like text).

uhoreg: To me, "Client are *not* required to supply a `Content-Type` header.` sounds like an encouragement to not bother sending a `Content-Type`, which just ... feels wrong to me.  At least if we say that they "should" send a `Content-Type`, we're encouraging people to send proper HTTP.  Maybe "Clients should supply an appropriate `Content-Type` header, but servers should be able to handle requests that do not have one, or an incorrect one."?  Though that's quite verbose. :shrug: This isn't a blocking comment; I'm just reluctant to encourage people to be sloppy.

content/identity-service-api.md
uhoreg: It might be nice to put a "for historical reasons" somewhere in that sentence (though I'm not sure where it would fit best) to indicate why it's like that, and that it isn't because that endpoint is actually special (as opposed to the other exceptions (media upload/download) which have actual reasons for not being JSON).  But if there's no nice way to fit that in, it's not the end of the world.

richvdh:
```suggestion
All `POST` and `PUT` endpoints, with the exception (for historical reasons) of [`POST
/_matrix/identity/v2/account/logout`](#post_matrixidentityv2accountlogout),
```

content/client-server-api/_index.md
richvdh: I think "should" ~= "MUST", so just saying "should" implies that servers are allowed to enforce it.  So I could say this:

```suggestion
Clients should supply a `Content-Type` header of `application/json` for all requests with JSON bodies,
but this is not required.
```

but it feels a bit wordy. 

I don't really see what we gain, as guardians of the matrix spec, by discouraging people from this particular bit of sloppiness.

uhoreg: > I don't really see what we gain, as guardians of the matrix spec, by discouraging people from this particular bit of sloppiness.

It feels to me that, since we're building on HTTP, we should be encouraging good HTTP behaviour, but maybe I'm being overly picky.

Your suggested wording seems fine to me, but if you don't think it's worth changing, then that's OK.

uhoreg (approved):

content/identity-service-api.md
```diff
@@ -38,8 +38,20 @@ Appendix.
...
 The mandatory baseline for identity server communication in Matrix is
 exchanging JSON objects over HTTP APIs. HTTPS is required for
-communication, and all API calls use a Content-Type of
-`application/json`. In addition, strings MUST be encoded as UTF-8.
+communication.
+
+All `POST` and `PUT` endpoints, with the exception (for historical reasons) of [`POST
+/_matrix/identity/v2/account/logout`](#post_matrixidentityv2accountlogout),
+require the client to supply a request body containing a (potentially empty)
+JSON object. Clients are *not* required to supply a `Content-Type` header.
```
richvdh, line 46:
```suggestion
JSON object. Clients should supply a `Content-Type` header of `application/json`
for all requests with JSON bodies, but this is not required.
```

content/server-server-api.md
```diff
@@ -47,13 +47,41 @@ an HTTPS PUT request.
...
-low-bandwidth/low-roundtrip mobile usage. For the default HTTP
-transport, all API calls use a Content-Type of `application/json`. In
-addition, all strings MUST be encoded as UTF-8.
+The mandatory baseline for server-server communication in Matrix is
+exchanging JSON objects over HTTPS APIs. More efficient transports may be
+specified in future as optional extensions.
+
+All `POST` and `PUT` endpoints require the requesting server to supply a
+request body containing a (potentially empty) JSON object. Requesting servers
+are *not* required to supply a `Content-Type` header.
```
richvdh, line 56:
```suggestion
request body containing a (potentially empty) JSON object. Requesting servers
should supply a `Content-Type` header of `application/json` for all requests
with JSON bodies, but this is not required.
```

content/identity-service-api.md
```diff
@@ -38,8 +38,20 @@ Appendix.
...
+
+Similarly, all endpoints require the server to return a JSON object.  Servers
+should include a `Content-Type` header of `application/json` for all JSON
+responses.
```
richvdh, line 50:
```suggestion
Similarly, all endpoints require the server to return a JSON object.  Servers
must include a `Content-Type` header of `application/json` for all JSON
responses.
```

content/client-server-api/_index.md
```diff
@@ -13,20 +13,33 @@ clients which maintain a full local persistent copy of server state.
...
+[`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload), require the
+client to supply a request body containing a (potentially empty) JSON object.
+Clients should supply a `Content-Type` header of `application/json` for all requests with JSON bodies,
+but this is not required.
+
+Similarly, all endpoints require the server to return a JSON object,
+with the exception of 200 responses to
+[`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](#get_matrixmediav3downloadservernamemediaid)
+and [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](#get_matrixmediav3thumbnailservernamemediaid).
+Servers should include a `Content-Type` header of `application/json` for all JSON responses.
```
richvdh, line 35:
```suggestion
Servers msut include a `Content-Type` header of `application/json` for all JSON responses.
```

content/server-server-api.md
```diff
@@ -47,13 +47,41 @@ an HTTPS PUT request.
...
+
+Similarly, all endpoints in this specification require the destination server
+to return a JSON object.  Servers should include a `Content-Type` header of
```
richvdh, line 59:
```suggestion
to return a JSON object.  Servers must include a `Content-Type` header of
```

content/client-server-api/_index.md
```diff
@@ -13,20 +13,33 @@ clients which maintain a full local persistent copy of server state.
...
+[`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload), require the
+client to supply a request body containing a (potentially empty) JSON object.
+Clients should supply a `Content-Type` header of `application/json` for all requests with JSON bodies,
+but this is not required.
+
+Similarly, all endpoints require the server to return a JSON object,
+with the exception of 200 responses to
+[`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](#get_matrixmediav3downloadservernamemediaid)
+and [`GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`](#get_matrixmediav3thumbnailservernamemediaid).
+Servers msut include a `Content-Type` header of `application/json` for all JSON responses.
```
uhoreg, line 35: s/msut/must/

richvdh: fixed in https://github.com/matrix-org/matrix-spec/commit/3dc3150f6d8c1bf77bd41af2b7540c1632d18998

PR #1191, opened 2022-08-02
Render HTML anchors for definition blocks

richvdh:
It's handy to be able to link to these.


turt2live (approved):
Changelog for this would be a good fit for https://github.com/matrix-org/matrix-spec/pull/1194 😇

PR #1192, opened 2022-08-02
Reinstate margin between tables

richvdh:
This partially undoes a change made in https://github.com/matrix-org/matrix-spec/pull/1166, but only includes a margin where two tables are adjacent.

Includes changes from #1193 

Preview: https://pr1192--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -313,10 +313,6 @@ footer {
     margin: 1.5rem 0 .75rem 0;
   }
 
-  h2 + table, h3 + table, h3 + div.highlight {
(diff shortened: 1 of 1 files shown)


assets/scss/custom.scss
```diff
@@ -313,10 +313,6 @@ footer {
     margin: 1.5rem 0 .75rem 0;
   }
 
-  h2 + table, h3 + table, h3 + div.highlight {
-    margin-top: 0;
-  }
-
```
richvdh, line 319: this was made redundant by #1166; removing it now for tidyness

PR #1194, opened 2022-08-02
Add an "internal changes" changelog section

turt2live:
To cover the cases where we don't have any other reasonable place to put the changelog entry. This helps ensure we don't have to constantly bypass the CI just to get everything merged.

Expected case is pretty much everything will be a `clarification` here, except for the introduction of the changelog itself.

Preview: https://pr1194--matrix-spec-previews.netlify.app


uhoreg (approved):
Seems to work as advertised

PR #1195, opened 2022-08-02
Give rendered-data sections a background and some padding

richvdh:
Not sure if everyone will like this, but it gives definition sections for APIs, events, etc a background and a small indent, which I find very helpful to guide the eye.

Suggest reviewing commit-by-commit as I had to move some stuff out of the way first.

Preview: https://pr1195--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -277,8 +277,6 @@ footer {
   details {
     summary {
       padding: .5rem 0;
-      list-style-position: outside;
-
       p {
         max-width: 80%;
       }
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -359,7 +359,6 @@ footer {
     &.basic-info, &.basic-info th, &.basic-info td {
       table-layout: fixed;
       margin: 1rem 0 .5rem 0;
-      background-color: $white;
     }
 
     &.basic-info th {
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -342,16 +342,18 @@ footer {
       padding: 1rem;
     }
 
-    th {
-      background-color: $white;
-    }
+    &.object-table, &.response-table {
+        caption, tr {
+            background-color: $table-row-default;
+        }
 
-    caption, tr {
-      background-color: $table-row-default;
-    }
+        tr:nth-child(even) {
+            background-color: $table-row-alternate;
+        }
 
-    tr:nth-child(even) {
-      background-color: $table-row-alternate;
+        th {
+            background-color: $white;
+        }
     }
 
     &.basic-info, &.basic-info th, &.basic-info td {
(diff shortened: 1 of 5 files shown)


assets/scss/custom.scss
```diff
@@ -270,7 +270,9 @@ footer {
 
 /* Styles for sections that are rendered from data, such as HTTP APIs and event schemas */
 .rendered-data {
-  margin: 1rem 0 3rem 0;
+  background-color: $secondary-lightest-background;
```
turt2live, line 273: This seems to have a jarring reaction with the arrow on the details section:

image


(accented for visibility)

assets/scss/custom.scss
```diff
@@ -342,22 +344,19 @@ footer {
...
-    th {
-      background-color: $white;
-    }
-
-    caption, tr {
-      background-color: $table-row-default;
-    }
+    &.object-table, &.response-table {
+        caption, tbody tr {
+            background-color: $table-row-default;
```
turt2live, line 349: table rows now blend in with the background:

image

assets/scss/custom.scss
turt2live: This can be fixed with `list-style-position: inside;` on the `summary` (rule just below here)

assets/scss/custom.scss
turt2live: I'd suggest making `$table-row-alternate` be `white` instead, as it makes `#f4fafc` pop more. Alternatively, don't bother with striping and play around with table layout some more, though this is more involved and requires a bunch of awkward CSS to get right (we'd want to change various alignments and justifications, possibly copy changes as well).

turt2live (commented):
overall I think this is a good improvement for readability. The two interactions in the review are easily fixed too, I think.

assets/scss/custom.scss
richvdh: Yeah, I'd noticed that, but decided I would rather live with it than have the alternate table rows be *lighter* than the background, which looks really odd to me.

How about we give the table a light border? Done in b62be510 - I think it looks quite good, though ymmv

turt2live: image

it feels a bit like the table is now bleeding into the background instead of being the background, but also a border is a marked enough improvement for now. Typically we avoid table borders as it can feel overly restricting, though here it seems to fit well enough to break up the pages and pages of text.

Feels like a thing someone else can PR if they don't like it, now.

assets/scss/custom.scss
```diff
@@ -343,6 +343,17 @@ footer {
...
     &.object-table, &.response-table {
+        border: 1px $table-border-color solid;
+
+        caption {
+            // the caption is outside the table's border box,
+            // so we have to give it its own border.
+            border: 1px $table-border-color solid;
+
+            // ... but avoid double border between caption and table
+            border-bottom: 0px;
```
turt2live, line 354: web developer extreme nit: it's `0`, not `0px` (unitless)

turt2live (approved):
lgtm either way - thanks!

assets/scss/custom.scss
richvdh:
```suggestion
            border-bottom: 0;
```

PR #1197, opened 2022-08-02
Add CORP headers to media repo

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3828

Preview: https://pr1197--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -260,6 +260,29 @@ footer {
     border-left-width: 5px;
     background: $warning-background;
   }
+
+  // XXX: See the added-in-paragraph.html shortcode for more information on these styles.
+  &.added-in-paragraph {
+    // Remove the padding and margin to remove the box look
+    margin: 0 !important; // !important on both to override table-related rules
+    padding: 0 !important;
+
+    // Make pairs of "added-in" and content inline to each other. We do pairs so authors can
+    // describe two paragraphs with added-in prefixes within a single box, reducing DOM
+    // complexity. Each paragraph is expected to be prefixed with an added-in, however.
+    //
+    // XXX: We assume the added-in and text will be rendered as paragraph elements.
+    > p {
+      display: inline;
+    }
+    > p:nth-child(2n) { // "even" rule to target just the content paragraphs
+      // Force a paragraph break after the content (insert a couple <br /> tags)
+      &::after {
+        content: '\A\A';
+        white-space: pre;
+      }
+    }
+  }
 }
 
 /* Styles for sections that are rendered from data, such as HTTP APIs and event schemas */
--- a/content/client-server-api/modules/content_repo.md
+++ b/content/client-server-api/modules/content_repo.md
@@ -19,6 +19,12 @@ When serving content, the server SHOULD provide a
 `Content-Security-Policy` header. The recommended policy is
 `sandbox; default-src 'none'; script-src 'none'; plugin-types application/pdf; style-src 'unsafe-inline'; object-src 'self';`.
 
+{{% added-in v="1.4" %}}
+
+The server SHOULD additionally provide `Cross-Origin-Resource-Policy: cross-origin`
+when serving content to allow (web) clients access to APIs which interact
+with the media repository, such as `SharedArrayBuffer`.
+
 #### Matrix Content (MXC) URIs
 
 Content locations are represented as Matrix Content (MXC) URIs. They
--- a/content/client-server-api/modules/content_repo.md
+++ b/content/client-server-api/modules/content_repo.md
@@ -19,11 +19,13 @@ When serving content, the server SHOULD provide a
 `Content-Security-Policy` header. The recommended policy is
 `sandbox; default-src 'none'; script-src 'none'; plugin-types application/pdf; style-src 'unsafe-inline'; object-src 'self';`.
 
-{{% added-in v="1.4" %}}
-
-The server SHOULD additionally provide `Cross-Origin-Resource-Policy: cross-origin`
+{{% boxes/added-in-paragraph %}}
+{{< added-in v="1.4" >}} The server SHOULD additionally provide `Cross-Origin-Resource-Policy: cross-origin`
 when serving content to allow (web) clients access to APIs which interact
 with the media repository, such as `SharedArrayBuffer`.
+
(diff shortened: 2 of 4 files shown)


content/client-server-api/modules/content_repo.md
```diff
@@ -19,6 +19,12 @@ When serving content, the server SHOULD provide a
 `Content-Security-Policy` header. The recommended policy is
 `sandbox; default-src 'none'; script-src 'none'; plugin-types application/pdf; style-src 'unsafe-inline'; object-src 'self';`.
 
+{{% added-in v="1.4" %}}
+
+The server SHOULD additionally provide `Cross-Origin-Resource-Policy: cross-origin`
```
richvdh, line 24: I thought we weren't doing RFC2119 keywords?

content/client-server-api/modules/content_repo.md
```diff
@@ -19,6 +19,12 @@ When serving content, the server SHOULD provide a
 `Content-Security-Policy` header. The recommended policy is
 `sandbox; default-src 'none'; script-src 'none'; plugin-types application/pdf; style-src 'unsafe-inline'; object-src 'self';`.
 
+{{% added-in v="1.4" %}}
```
richvdh, line 22: I don't find it clear which part was added: was it the bit before (about C-S-P) or the bit after the '[Added in]' text?

content/client-server-api/modules/content_repo.md
```diff
@@ -19,6 +19,12 @@ When serving content, the server SHOULD provide a
...
+
+The server SHOULD additionally provide `Cross-Origin-Resource-Policy: cross-origin`
+when serving content to allow (web) clients access to APIs which interact
+with the media repository, such as `SharedArrayBuffer`.
```
richvdh, line 26: this doesn't sound right. `SharedArrayBuffer` doesn't interact with the media repo?

I think this should be something like '... allow clients which interact with the media repo to access restricted APIs such as `SharedArrayBuffer` by setting a `Cross-Origin-Embedder-Policy: require-corp` header', though I may be getting it slightly mixed up

richvdh (changes requested):

content/client-server-api/modules/content_repo.md
turt2live: Since when? We've always used SHOULD and MUST, just not MAY. It's in the API standards section...

content/client-server-api/modules/content_repo.md
turt2live: Precedent is always the text after. Suggestions welcome for how to fix it, though I will note that inline annotations are basically impossible.

richvdh: With an inline annotation, like https://github.com/matrix-org/matrix-spec/blame/main/content/client-server-api/_index.md#L363, or frankly, just:

``` 
**[Added in `v1.4`]**: foo bar
```

without any templating magic

content/client-server-api/modules/content_repo.md
richvdh: :shrug: fair enough. I thought at some point we'd decided not to use it as it's not familiar for most readers.

>  It's in the API standards section...

I don't see this.

content/client-server-api/modules/content_repo.md
turt2live: The inline annotation doesn't work in paragraphs for annoying and complicated reasons - it works on lists though.

I would prefer it if we always used the template for find-and-replace reasons, or if we decide to change the format down the line. I'm willing to codify this somewhere, if needed.

content/client-server-api/modules/content_repo.md
turt2live: I swear it was there. Somewhere we describe that we use MUST and SHOULD (which have the same meaning as must and should), but not MAY and such. 

If we don't have this written down somewhere, my whole life has been a lie 😭

content/client-server-api/modules/content_repo.md
turt2live: I think your wording is more correct though ultimately the same depending on perspective - will play with the wording a bit.

content/client-server-api/modules/content_repo.md
richvdh: well, I don't know. Can we stick it in a box of some kind? or use a bullet list, if that's enough to make the annotation work (one bullet for each of CSP and CORP).

I feel like we need to do *something* here: it's very unclear as it is.

content/client-server-api/modules/content_repo.md
richvdh: I'd search github for "MUST", but that isn't going to go well :-S

turt2live: even searching the repo for "MUST" (case sensitive) yields a ton of results, so I'm inclined to say that even though it's not written down, we follow the rule:

<details>
<summary>176 matches over 53 files (tall image)</summary>

image

</details>

content/client-server-api/modules/content_repo.md
turt2live: ftr, the inline issue is this: https://github.com/matrix-org/matrix-spec/issues/1204

I've fixed the inline-ness using a somewhat ugly hack, but it's probably fine enough for now.

assets/scss/custom.scss
```diff
@@ -260,6 +260,29 @@ footer {
...
+  &.added-in-paragraph {
+    // Remove the padding and margin to remove the box look
+    margin: 0 !important; // !important on both to override table-related rules
+    padding: 0 !important;
+
+    // Make pairs of "added-in" and content inline to each other. We do pairs so authors can
+    // describe two paragraphs with added-in prefixes within a single box, reducing DOM
+    // complexity. Each paragraph is expected to be prefixed with an added-in, however.
+    //
+    // XXX: We assume the added-in and text will be rendered as paragraph elements.
```
turt2live, line 274: as mentioned in the commit: we don't need the complexity of 2+ paragraphs yet, but it's good behaviour to have for when we need it in the future.

richvdh (approved):

richvdh:
can/should we simplify this now that #1204 is (hopefully) fixed?

turt2live:
https://github.com/matrix-org/matrix-spec/issues/1207

Issue #331, opened 2018-07-19
Upgrade OpenAPI


turt2live:
Currently we use 2.x - upgrading gets us a lot of benefits and appears relatively painless

richvdh:
I found https://blog.stoplight.io/difference-between-open-v2-v3-v31 a useful introduction to the changes here.

zecakeh:
Hi ! I'm interested in working on this, if no one else is currently doing it.

I already ran the [swagger2openapi](https://github.com/Mermade/oas-kit/tree/main/packages/swagger2openapi) script to see how much work is involved. It seems like it didn't break much on the OpenAPI side, but it did some linting at the same time so it's gonna take some time to check that nothing was removed.

Would it be interesting to switch directly to OpenAPI 3.1? Afaict, the tools used all support it (RapiDoc for the UI and swagger-parser for the CI), however the README mentions testing with swagger-ui, which doesn't have support yet.

richvdh:
I don't think we use swagger-ui any more (it's superseded by Rapidoc).

I'd be very excited to see work happening in this direction, however do have a look at https://github.com/matrix-org/matrix-spec-proposals/pull/1446#issuecomment-409393255 which contains some notes on a previous attempt at this work.

zecakeh:
Thanks I already took a look at this thread. I wrote a simple conversion script that at least keeps the licensing comments at the top.

I see that the `check-swagger-sources.py` script doesn't seem to be used or mentioned anywhere, but looks like a predecessor of `validator.js`? So I'm wondering if I should also update it for eventual changes.

richvdh:
I don't think `check-swagger-sources.py` is a predecessor to `validator.js` - they have both coexisted for a long time. Still, I'm not entirely sure what the difference is, and it's true that `check-swagger-sources.py` doesn't seem to be used anywhere (and it doesn't currently pass).

So, I wouldn't block any upgrade job on maintaining `check-swagger-sources.py` - we can do that separately.

afranke:
See also #8

zecakeh:
I opened #1310 (still a draft). Feedback is very welcome.

PR #1200, opened 2022-08-03
Use auth header instead of query param for hs->as comms

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/2832

Preview: https://pr1200--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/application_service/newsfragments/1200.breaking
+++ b/changelogs/application_service/newsfragments/1200.breaking
@@ -1,0 +1,1 @@
+Replace homeserver authorization approach with an `Authorization` header instead of `access_token` when talking to the application service, as per [MSC2832](https://github.com/matrix-org/matrix-spec-proposals/pull/2832).
--- a/data/api/application-service/definitions/security.yaml
+++ b/data/api/application-service/definitions/security.yaml
@@ -1,4 +1,5 @@
 # Copyright 2018 New Vector Ltd
+# Copyright 2022 The Matrix.org Foundation C.I.C.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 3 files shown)


changelogs/application_service/newsfragments/1200.breaking
```diff
@@ -0,0 +1 @@
+Replace homeserver authorization approach with an `Authorization` header instead of `access_token` when talking to the application service, as per [MSC2832](https://github.com/matrix-org/matrix-spec-proposals/pull/2832).
```
turt2live, line 1: "breaking change" to *really* highlight that this is different now.

data/api/application-service/definitions/security.yaml
```diff
@@ -13,6 +14,6 @@
 # limitations under the License.
 homeserverAccessToken:
   type: apiKey
-  description: The `hs_token` provided by the application service's registration.
-  name: access_token
-  in: query
+  name: Authorization
+  in: header
+  description: The `Bearer` `hs_token` provided by the application service's registration.
```
richvdh, line 19: sigh. openapi 3.0 has better support for bearer tokens (https://swagger.io/docs/specification/authentication/bearer-authentication/). It really feels like we need to sort out https://github.com/matrix-org/matrix-spec/issues/331.

turt2live: indeed :( 

https://github.com/matrix-org/matrix-spec/pull/1200/commits/c510b1254022646803fb29fde96474d873cf8e89 was me thinking I could use v3 but reverting to v2

richvdh (approved):

Issue #1204, opened 2022-08-09
Client-server spec gets processed for Markdown twice


turt2live:
The other specs do not appear to be affected by this for some reason. I haven't looked into why or what might be causing this.

The effect of this bug is that shortcodes which would normally be inline end up as blocks. For example, stick this somewhere in the prose:

```
{{< added-in v="1.4" >}} This is meant to have an "Added In" before it, inline. Not as a block element above this text.
```

The angle brackets mean to handle it as a literal (`%` means process it), but regardless it ends up spitting out HTML into the interstitial document, which then gets a second markdown process over it and converts the HTML to a block element. Ideally we stick the HTML from the shortcodes into a *rendered* markdown document so the formatting is preserved.

richvdh:
I think #1205 fixes this?

PR #1205, opened 2022-08-09
include CS-api modules with `{{<`

richvdh:
... because that fixes everything. Or at least, #1204.

Preview: https://pr1205--matrix-spec-previews.netlify.app


turt2live (approved):
well that's less sinister than I was expecting

thanks!

PR #1206, opened 2022-08-09
Spec reference relationships

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3267

Preview: https://pr1206--matrix-spec-previews.netlify.app


content/client-server-api/_index.md
```diff
@@ -2643,3 +2644,4 @@ systems.
 {{% cs-module name="server_notices" %}}
 {{% cs-module name="moderation_policies" %}}
 {{% cs-module name="spaces" %}}
+{{< cs-module name="reference_relations" >}}
```
turt2live, line 2647: per https://github.com/matrix-org/matrix-spec/pull/1205

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+
+Generically referencing another event can be done with a `rel_type` of `m.reference`
+as a form of [relationship](#forming-relationships-between-events). There is no
+implied meaning behind the reference, and is usually context-dependent. Such an
+example is the [key verification framework](#key-verification-framework) which uses
+reference relations to associate distinct events with a specific verification attempt.
+
+When bundled, a reference relations appear as an object with a `chunk` field to
+indicate all the events which `m.reference` the event (the parent). Currently,
+only a single `event_id` field is present on the events in the `chunk`.
```
clokep, line 17: There's also potentially a `next_batch` token to pass back into `/relations` to get more?

This seems to be at this discussion which was never resolved in the MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3267#discussion_r883541312

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+
+{{% added-in v="1.4" %}}
+
+Generically referencing another event can be done with a `rel_type` of `m.reference`
+as a form of [relationship](#forming-relationships-between-events). There is no
+implied meaning behind the reference, and is usually context-dependent. Such an
+example is the [key verification framework](#key-verification-framework) which uses
+reference relations to associate distinct events with a specific verification attempt.
+
+When bundled, a reference relations appear as an object with a `chunk` field to
```
richvdh, line 15: s/bundled/aggregated/ ? with a link to https://spec.matrix.org/v1.3/client-server-api/#aggregations ?

richvdh: Also s/field/property/ per https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#general.

```suggestion
The aggregation format of `m.reference` relations consists of a single `chunk` property to
```

turt2live: > s/bundled/aggregated/ ? with a link to https://spec.matrix.org/v1.3/client-server-api/#aggregations ?

The MSC says bundled and I was hoping that would magically fit in our spec language 😭

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+    }
+    // other content fields as required
+  }
+  // other fields as required by events
+}
+```
+
+The [bundle](#aggregations) under `m.relations` would appear similar to the following:
+
+```json5
```
turt2live, line 42:
```suggestion
```json
```

content/client-server-api/modules/reference_relations.md
turt2live: I'd be inclined to intentionally punt that to a new MSC rather than backport it into MSC3267, I think. There's some questions about how it would work that start to feel like MSC-needing territory, like what if there's no more at the time the bundle is made or how long the server needs to keep the token valid for.

turt2live: Also, in practice clients *should* receive the follow-up events in most cases because they'll get the bundle from `/sync` and then be actively syncing again. There's a chance that they'll go offline again and create a gap, but backfill should repopulate the client's stores for local aggregation, or the event will be old enough where new references aren't necessarily important (for the current usecases in the spec, at least).

clokep: That all sounds reasonable...but if we're not going to spec it, we should stop returning it from synapse (but I have no idea if clients are depending on it).

turt2live: I'd be comfortable removing it from Synapse tbh - it doesn't seem all that important to clients, but equally I am unsure if there's anyone out there using it (if they are, they can write an MSC imo)

clokep: OK, we should file a Synapse issue at the very least then!

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+      "rel_type": "m.reference",
+      "event_id": "$another_event"
+    }
+    // other content fields as required
+  }
+  // other fields as required by events
+}
+```
+
+The [bundle](#aggregations) under `m.relations` would appear similar to the following:
```
richvdh, line 40: s/bundle/aggregation/ ?

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+
+Generically referencing another event can be done with a `rel_type` of `m.reference`
+as a form of [relationship](#forming-relationships-between-events). There is no
+implied meaning behind the reference, and is usually context-dependent. Such an
+example is the [key verification framework](#key-verification-framework) which uses
+reference relations to associate distinct events with a specific verification attempt.
+
+The aggregation format of `m.reference` relations consists of a single `chunk` property to
+indicate all the events which `m.reference` the event (the parent). Currently,
+only a single `event_id` field is present on the events in the `chunk`.
```
richvdh, line 17: I think it would be nice if this and the example aggregation should be grouped together under a "Server-side aggregation of `m.reference` relationships" heading, for consistency with edits.

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+{{% added-in v="1.4" %}}
+
+Generically referencing another event can be done with a `rel_type` of `m.reference`
+as a form of [relationship](#forming-relationships-between-events). There is no
+implied meaning behind the reference, and is usually context-dependent. Such an
+example is the [key verification framework](#key-verification-framework) which uses
+reference relations to associate distinct events with a specific verification attempt.
+
+The aggregation format of `m.reference` relations consists of a single `chunk` property to
+indicate all the events which `m.reference` the event (the parent). Currently,
```
richvdh, line 16:
```suggestion
The aggregation format of `m.reference` relations consists of a single `chunk` property, which
lists all the events which `m.reference` the event (the parent). Currently,
```

content/client-server-api/modules/reference_relations.md
```diff
@@ -0,0 +1,51 @@
...
+---
+
+### Reference relations
+
+{{% added-in v="1.4" %}}
+
+Generically referencing another event can be done with a `rel_type` of `m.reference`
+as a form of [relationship](#forming-relationships-between-events). There is no
+implied meaning behind the reference, and is usually context-dependent. Such an
+example is the [key verification framework](#key-verification-framework) which uses
```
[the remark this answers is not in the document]
turt2live: applied manually

content/client-server-api/modules/reference_relations.md
turt2live: I think this is a bundle, in reference to the same language `m.replace` uses.

content/client-server-api/modules/reference_relations.md
turt2live: applied manually

content/client-server-api/_index.md
```diff
@@ -2639,4 +2640,5 @@ systems.
 {{< cs-module name="moderation_policies" >}}
 {{< cs-module name="spaces" >}}
 {{< cs-module name="event_replacements" >}}
+{{< cs-module name="threading" >}}
```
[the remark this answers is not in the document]
turt2live: yea, we should indeed. Will take a look in a different PR.

Issue #676, opened 2020-08-07
Login/register endpoints say the device_id and stuff aren't required in the response


turt2live:
This is very much wrong

anoadragon453:
> Same for register endpoint.

For `/register`, the spec currently says that `device_id` and `access_token` are only required if `inhibit_login` in the request is false. Synapse also [follows this advice](https://github.com/matrix-org/synapse/blob/d6e94ad9d9b22efe0f5eb64d946c10a542068f7c/synapse/rest/client/register.py#L803-L820).

PR #1210, opened 2022-08-12
Require `access_token`, `device_id` and `user_id` fields in `/login` response

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/676

Preview: https://pr1210--matrix-spec-previews.netlify.app


richvdh (approved):
lgtm!

PR #1211, opened 2022-08-16
Spec event edits

richvdh:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/2676

Preview: https://pr1211--matrix-spec-previews.netlify.app


content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+an event's content property, such as `formatted_body` (see [`m.room.message`
+`msgtypes`](#mroommessage-msgtypes)).
+
+#### Validity of message edit events
+
+There are a number of requirements on message edit events, which must be satisfied for the replacement to be considered valid:
+
+ * As with all event relationships, the original event and replacement event
+   must have the same `room_id` (i.e. it is not possible to send an event in
+   one room and then an edited version in a different room).
```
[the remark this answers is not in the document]
turt2live: phrasing: half of the `i.e.` bits are saying "You..." but the others say "It...". It's easier language-wise to stick to "You..." than it is to try and adopt "It..." across the points.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+ * The original event and replacement event must have the same `sender`
+   (i.e. you cannot edit someone else's messages).
+
+ * The replacement and original events must have the same `type` (i.e. editing
+   an event cannot change its type).
+
+ * Neither the replacement nor original events may have a `state_key` property
+   (i.e. it is not possible to edit a state event).
```
turt2live, line 59:
```suggestion
   (i.e. you cannot edit state events at all).
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+   must have the same `room_id` (i.e. it is not possible to send an event in
+   one room and then an edited version in a different room).
+
+ * The original event and replacement event must have the same `sender`
+   (i.e. you cannot edit someone else's messages).
+
+ * The replacement and original events must have the same `type` (i.e. editing
+   an event cannot change its type).
+
+ * Neither the replacement nor original events may have a `state_key` property
```
turt2live, line 58:
```suggestion
 * Neither the replacement nor original events have a `state_key` property
```

turt2live: "may" is highly ambiguous when reading as a non-native speaker, so we avoid it. I think the sentence works just fine without the word rather than trying to find a replacement.

If we do want a word in here though, "must not have a state_key" would be the phrasing to adopt.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+ * The replacement event (once decrypted, if appropriate) must have an
+   `m.new_content` property.
+
+If any of these criteria are not satisfied, implementations should ignore the
+replacement event (the content of the original should not be replaced, and the
+edit should not be included in the server-side aggregation).
+
+Note that the [`msgtype`](#mroommessage-msgtypes) property of replacement
+`m.room.message` events need *not* be the same as in the original event. For
```
turt2live, line 73:
```suggestion
`m.room.message` events does *not* need to be the same as in the original event. For
```

turt2live: non-native speaker clarity

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+#### Editing encrypted events
+
+If the original event was [encrypted](#end-to-end-encryption), the replacement
+should be too. In that case, `m.new_content` is placed in the content of the
+encrypted payload. As with all event relationships, the `m.relates_to` property
+must be sent in the unencrypted (cleartext) part of the event.
+
+For example, a replacement for an encrypted event might look like this:
+
+For example, an encrypted replacement event might look like this:
```
turt2live, line 85: I prefer the second, but either are fine to me.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+```json
```
turt2live, line 87:
```suggestion
```json5
```

content/client-server-api/_index.md
```diff
@@ -1938,14 +1938,14 @@ parent event, for example.
 {{% /boxes/note %}}
 
 To allow the server to aggregate and find child events for a parent, the `m.relates_to`
-key of an event MUST be included in the plaintext copy of the event. It cannot be
+key of an event MUST be included in the cleartext (unencrypted) part of the event. It cannot be
```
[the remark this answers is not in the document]
turt2live: How it's presented in the edits module is clearer: the nearby examples help clarify what is being talked about, and the paragraphs take the time to be clear about what goes where. Unlike here, where the word "cleartext" (and "plaintext") is largely on its own, unsupported by other bits of text/examples.

I don't think we need to change the terminology in the edits module, but would suggest splitting this conversion out to its own PR so we can think about the surrounding text a bit more.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+            "body": "Hello! My name is bar",
+            "msgtype": "m.text"
+        }
+    }
+}
+```
+
+Note that:
+
+ * There is no `m.relates_to` property in the encrypted payload. (Any such property would be ignored.)
```
turt2live, line 124:
```suggestion
 * There is no `m.relates_to` property in the encrypted payload. If there was, it would be ignored.
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+ * There is no `m.new_content` property in the cleartext content of the `m.room.encrypted` event. (Again, any such property would be ignored.)
```
turt2live, line 125:
```suggestion
 * There is no `m.new_content` property in the cleartext content of the `m.room.encrypted` event. As above, if there was then it would be ignored.
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+The payload of an encrypted replacement event must be encrypted as normal, including ratcheting any [Megolm](#mmegolmv1aes-sha2) session as normal. The original Megolm ratchet entry should **not** be re-used.
```
turt2live, line 127: This is quite confusing in this context, I think because it kinda comes out of nowhere and uses terms not currently present in the spec. It sounds like we're trying to warn the reader to not re-use sessions? (if so, we should say that in a warning or info box rather than as a floating paragraph)

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+
+#### Applying `m.new_content`
+
+When applying a replacement, the `content` of the original event is replaced
+entirely by the `m.new_content` from the replacement event, with the exception
+of `m.relates_to`, which is left *unchanged*. (Any `m.relates_to` property
+within `m.new_content` should be ignored.)
```
turt2live, line 135: This says "replace" a lot:

```suggestion
When applying a replacement, the `content` of the original event is treated as
overwritten entirely by `m.new_content`, with the exception of `m.relates_to`,
which is left *unchanged* because any `m.relates_to` property within `m.new_content`
is ignored. Note that the original event's `content` is not actually overwritten
or replaced by the new content: this is to provide history of the changes.
```

also added a bit to clarify that the original event doesn't change because that'd cause problems for auth and such: we're just dealing with it at the client level like that, because we can.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+        },
+        "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$original_event_id"
+        }
+    }
+}
+```
+
+... then the end result is an event as shown below:
```
turt2live, line 171: would be ideal to clarify this is affecting *representation* of the event and not that the server go out and do an `UPDATE` against its database.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+The aggregation format of `m.replace` relationships gives the `event_id`,
+`origin_server_ts`, and `sender` of the **most recent** replacement event. The
+most recent event is determined by comparing `origin_server_ts`; if two or more
+replacement events have identical `origin_server_ts`, the event with the
+lexicographically largest `event_id` is treated as more recent.
+
+This aggregation is bundled into the `unsigned/m.relations` property of any
+event that is the target of an `m.replace` relationship. For example:
+
+```json
```
turt2live, line 205:
```suggestion
```json5
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+{
+  "event_id": "$original_event_id",
+  // ...
```
turt2live, line 208:
```suggestion
  // irrelevant fields not shown
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+            "rel_type": "m.replace",
+            "event_id": "$some_event_id"
+        },
+        "algorithm": "m.megolm.v1.aes-sha2",
+        "sender_key": "<sender_curve25519_key>",
+        "device_id": "<sender_device_id>",
+        "session_id": "<outbound_group_session_id>",
+        "ciphertext": "<encrypted_payload_base_64>"
+    }
+    // ...
```
turt2live, line 101:
```suggestion
    // irrelevant fields not shown
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+given event (for example, if an event is edited multiple times). These should
+be [aggregated](#aggregations) by the homeserver.
+
+The aggregation format of `m.replace` relationships gives the `event_id`,
+`origin_server_ts`, and `sender` of the **most recent** replacement event. The
+most recent event is determined by comparing `origin_server_ts`; if two or more
+replacement events have identical `origin_server_ts`, the event with the
+lexicographically largest `event_id` is treated as more recent.
+
+This aggregation is bundled into the `unsigned/m.relations` property of any
```
turt2live, line 202:
```suggestion
This aggregation is bundled under the `unsigned` property as `m.relations` for any
```

turt2live: the `/` notation is relatively foreign and unclear within the spec

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+        "origin_server_ts": 1649772304313,
+        "sender": "@editing_user:localhost"
+      }
+    }
+  }
+}
+```
+
+If the original event is
+[redacted](http://localhost:1313/client-server-api/#redactions), any
```
turt2live, line 222:
```suggestion
[redacted](#redactions), any
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+An exception applies to [`GET /_matrix/client/v3/rooms/{roomId}/event/{eventId}`](#get_matrixclientv3roomsroomideventeventid),
+which should return the unmodified event (though the relationship should still
+be bundled, as described above).
+
+#### Client behaviour
+
+Clients can often ignore `m.replace` events, since any events the server
+returns via the C-S API will be updated by the server to account for subsequent
+edits.
```
turt2live, line 242:
```suggestion
Clients can often ignore `m.replace` events, because any events returned
by the server to the client will be updated by the server to account for
subsequent edits.
```

"since" has too many meanings, like "may". "C-S API" is not a word we use.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+unable to do so. This happens in the following situations:
+
+ * The client has already received and stored the original event before the
+   message edit event arrives.
+
+ * The original event (and hence its replacement) are encrypted.
+
+Client authors are reminded to take note of the requirements for [Validity of
+message edit events](#validity-of-message-edit-events), and to ignore any
+invalid edit events that may be received.
```
turt2live, line 254:
```suggestion
invalid edit events that are received.
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+removes that edit revision. This has little effect if there were subsequent
+edits. However, if it was the most recent edit, the event is in effect
+reverted to its content before the redacted edit.
+
+Redacting the *original* message in effect removes the message, including all
+subsequent edits, from the visible timeline. In this situation, homeservers
+will return an empty `content` for the original event as with any other
+redacted event, and as
+[above](#server-side-aggregation-of-mreplace-relationships) the replacement
+events will not be bundled with the original event.
```
turt2live, line 277:
```suggestion
events will not be bundled with the original event. Note that the subsequent edits are
not actually redacted themselves: they simply serve no purpose within the visible timeline.
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+ * In contrast to the original reply, there should be no `m.in_reply_to`
+   property in the the `m.relates_to` object, since it would be redundant (see
+   [Applying `m.new_content`](/#applying-mnew_content) above, which notes that
+   the original event's `m.relates_to` is preserved), as well as being contrary
+   to the spirit of the event relationships mechanism which expects only one
+   "parent" per event.
+
+ * `m.new_content` should **not** contain any [reply
+   fallback](https://spec.matrix.org/v1.3/client-server-api/#fallbacks-for-rich-replies),
```
turt2live, line 292:
```suggestion
   fallback](#fallbacks-for-rich-replies),
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+events will not be bundled with the original event.
+
+#### Edits of replies
+
+Some particular constraints apply to events which replace a
+[reply](#rich-replies). In particular:
+
+ * In contrast to the original reply, there should be no `m.in_reply_to`
+   property in the the `m.relates_to` object, since it would be redundant (see
+   [Applying `m.new_content`](/#applying-mnew_content) above, which notes that
```
turt2live, line 286:
```suggestion
   [Applying `m.new_content`](#applying-mnew_content) above, which notes that
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+
+ * `m.new_content` should **not** contain any [reply
+   fallback](https://spec.matrix.org/v1.3/client-server-api/#fallbacks-for-rich-replies),
+   since it is assumed that any client which can handle edits can also display
+   replies natively. However, the `content` of the replacement event should provide
+   fallback content for clients which support neither rich replies nor edits.
+
+An example of an edit to a reply is as follows:
+
+```json
```
turt2live, line 299:
```suggestion
```json5
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+{
+  "type": "m.room.message",
```
turt2live, line 301:
```suggestion
  "type": "m.room.message",
  // irrelevant fields not shown
```

turt2live (commented):
This is a first pass review of the PR, not taking into consideration what the MSC actually says (just relying on what I remember of the MSC). 

Most of the comments here are about voice (effectively) as well as general clarity: where possible, I've suggested alternative wording alongside rationale for the change. 

I believe this PR is factually accurate enough to ask that this batch of comments be dealt with before receiving a second pass of review, though if you'd prefer to have the second pass first then re-request review and I'll take a look.

turt2live:
also, deserving of its own comment: thank you for writing this up 😄 it's clear quite a lot of work went into this, and is well written :)

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+{{% added-in v="1.4" %}}
+
+Event replacements, or "message edit events", are events that use an [event
+relationship](http://localhost:1313/client-server-api/#forming-relationships-between-events)
+with a `rel_type` of `m.replace`, which indicates that the original event is
+intended to be replaced.
+
+An example of a message edit event might look like this:
+
+```json
```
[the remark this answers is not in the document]
richvdh: json5 doesn't get synatax-highlighted by hugo. OTOH, its json syntax highlighter is happy to accept comments.

content/client-server-api/_index.md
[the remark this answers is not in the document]
richvdh: "plaintext" is simply wrong here (plaintext is stuff that was/will be encrypted; cleartext is stuff that is never encrypted; cf https://www.hypr.com/cleartext/), so I was just trying to fix an error while I was in the area. Happy to punt it if you'd prefer though.

content/client-server-api/modules/event_replacements.md
richvdh:
```suggestion
 * The replacement and original events must not have a `state_key` property
```

content/client-server-api/modules/event_replacements.md
richvdh: but both is twice as good, right? :facepalm:

richvdh, line 85:
```suggestion
For example, a replacement for an encrypted event might look like this:
```

richvdh:
```suggestion
For example, a replacement for an encrypted event might look like this:
```

content/client-server-api/modules/event_replacements.md
richvdh: yeah, exactly. I'll stick it in an info box.

richvdh: > It sounds like we're trying to warn the reader to not re-use sessions?

To be more precise: we're trying to warn the reader not to re-use the same ratchet entry in a given session. A new ratchet entry in the existing session is fine.

content/client-server-api/modules/event_replacements.md
richvdh: One of the ideas here is that client impls *should* be allowed to overwrite the original event's `content`. You're right that servers need to preserve the original though.

```suggestion
When applying a replacement, the `content` of the original event is treated as
being overwritten entirely by `m.new_content`, with the exception of `m.relates_to`,
which is left *unchanged*. Any `m.relates_to` property within `m.new_content`
is ignored. (Note that server implementations must not *actually* overwrite
the original event's `content`: instead the server presents it as being overwritten
when it is served over the client-server API. See [Server-side replacement of content](#server-side-replacement-of-content)
below.
```

content/client-server-api/modules/event_replacements.md
richvdh: Do you think the above is now enough for this?

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,319 @@
...
+```
+
+If the original event is
+[redacted](http://localhost:1313/client-server-api/#redactions), any
+`m.replace` relationship should **not** be bundled with it (whether or not any
+subsequent replacements are themselves redacted). Note that this behaviour is
+specific to the `m.replace` relationship. See also [redactions of edited
+events](#redactions-of-edited-events) below.
+
+#### Server-side replacement of content
```
richvdh, line 228:
```suggestion
##### Server-side replacement of content
```

content/client-server-api/modules/event_replacements.md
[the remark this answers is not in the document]
turt2live: that's annoying, but not for reasons that matter.

content/client-server-api/_index.md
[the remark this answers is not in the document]
turt2live: Punting it and calling it a release blocker clarification would be the ideal workflow here, I think.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,326 @@
...
+Note that:
+
+ * There is no `m.relates_to` property in the encrypted payload. If there was, it would be ignored.
+ * There is no `m.new_content` property in the cleartext content of the `m.room.encrypted` event. As above, if there was then it would be ignored.
+
+{{% boxes/note %}}
+The payload of an encrypted replacement event must be encrypted as normal, including
+ratcheting any [Megolm](#mmegolmv1aes-sha2) session as normal. The original Megolm
+ratchet entry should **not** be re-used.
+{{% /boxes/note %}}
```
turt2live, line 129: I'm afraid this still doesn't make any sense to me as a reader - we have zero mention of a "ratchet entry" in the spec, and it's not a term I've heard come up in conversation before. I realize this is coming from the MSC, but it also doesn't make sense there having re-read it.

Is this the "ratchet index" (a defined term in the spec) or can we just say it's "encrypted like any other event" and avoid the problem entirely?

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,326 @@
...
+
+#### Applying `m.new_content`
+
+When applying a replacement, the `content` of the original event is treated as
+being overwritten entirely by `m.new_content`, with the exception of `m.relates_to`,
+which is left *unchanged*. Any `m.relates_to` property within `m.new_content`
+is ignored. (Note that server implementations must not *actually* overwrite
+the original event's `content`: instead the server presents it as being overwritten
+when it is served over the client-server API. See [Server-side replacement of content](#server-side-replacement-of-content)
+below.)
```
turt2live, line 140:
```suggestion
is ignored.

{{% boxes/note %}}
Note that server implementations must not *actually* overwrite
the original event's `content`: instead the server presents it as being overwritten
when it is served over the client-server API. See [Server-side replacement of content](#server-side-replacement-of-content)
below.
{{% /boxes/note %}}
```

this seems important enough to be coloured blue, and is a bit long for brackets in retrospect.

turt2live (commented):
overall this looks good to me. For blockers: 
* Splitting out the cleartext description in the relationships section (the wording can stay unchanged in the edits module)
* Consistency on the feature name within the spec
* Stylistic things mentioned as suggestions (could be argued otherwise, though)

content/client-server-api/_index.md
richvdh, line 1941:
```suggestion
key of an event MUST be included in the plaintext copy of the event. It cannot be
```

content/client-server-api/_index.md
```diff
@@ -1938,14 +1938,14 @@ parent event, for example.
...
 exclusively recorded in the encrypted payload as the server cannot decrypt the event
 for processing.
 
 {{% boxes/warning %}}
 If an encrypted event contains an `m.relates_to` in its payload, it should be
-ignored and instead favour the plaintext `m.relates_to` copy (including when there
-is no plaintext copy). This is to ensure the client's behaviour matches the server's
+ignored and instead favour the cleartext `m.relates_to` copy (including when there
+is no cleartext copy). This is to ensure the client's behaviour matches the server's
```
richvdh, line 1948:
```suggestion
ignored and instead favour the plaintext `m.relates_to` copy (including when there
is no plaintext copy). This is to ensure the client's behaviour matches the server's
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,326 @@
...
+`msgtype` etc.) provide a fallback for clients which do not understand
+replacement events.
+
+`m.new_content` can include any properties that would normally be found in
+an event's content property, such as `formatted_body` (see [`m.room.message`
+`msgtypes`](#mroommessage-msgtypes)).
+
+#### Validity of message edit events
+
+There are a number of requirements on message edit events, which must be satisfied for the replacement to be considered valid:
```
richvdh, line 46:
```suggestion
#### Validity of replacement events

There are a number of requirements on replacement events, which must be satisfied for the replacement to be considered valid:
```

content/client-server-api/_index.md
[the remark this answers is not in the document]
richvdh: filed as https://github.com/matrix-org/matrix-spec/issues/1223. Not sure if it needs to actually be a blocker.

content/client-server-api/modules/event_replacements.md
```diff
@@ -0,0 +1,326 @@
...
+The `content` of the replacement must contain a `m.new_content` property which
+defines the replacement `content`. The normal `content` properties (`body`,
+`msgtype` etc.) provide a fallback for clients which do not understand
+replacement events.
+
+`m.new_content` can include any properties that would normally be found in
+an event's content property, such as `formatted_body` (see [`m.room.message`
+`msgtypes`](#mroommessage-msgtypes)).
+
+#### Validity of message edit events
```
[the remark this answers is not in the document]
richvdh: done, I hope

content/client-server-api/modules/event_replacements.md
richvdh: Right, it should be "ratchet value", from https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md#the-megolm-ratchet-algorithm.

My general feeling is: if you're implementing edits of encrypted events, you'll know what this means, and if you're not, you don't need to worry about it.

> can we just say it's "encrypted like any other event"

Well, at that point, it doesn't seem to give any information at all, so we might as well just omit it.

TBH it seems kindof obvious to me that we *wouldn't* reuse the old key, so I'd be happy to omit this block, but this text was added in response to @erikjohnston's question at https://github.com/matrix-org/matrix-spec-proposals/pull/2676#discussion_r913844941. Erik: do you have any thoughts on whether this text is necessary?

turt2live: due to lack of comment I'm inclined to believe it's just a safety net thing. It doesn't feel particularly safe to assume that the person implementing edits of encrypted events is also aware of the intricacies of encryption, however I'm happy to go with whatever at this point.

turt2live (approved):
let's just send it and make clarifications after the fact if needed.

PR #1237, opened 2022-09-15
Simplify password requirements

NicolaiSoeborg:
A strong password is not necessary complex (https://xkcd.com/936/). Having arbitrary password requirements is a bit of a pain - we should rather push for 2FA / passwordless auth. 

I don't think this is a breaking change as it only change a "SHOULD" sentence. Any mismatch between clients / servers should still be 'spec complient' within the old text.

ECP = Enforce Complex Password
Client ECP - Server ECP: ✅
Client ECP - Server not-ECP: ✅
Client not-ECP - Server not-ECP: ✅
Client not-ECP - Server ECP: Potentially `M_WEAK_PASSWORD` which the client should already be able to handle.

This change will probably make it easier to implement [MSC2000](https://github.com/matrix-org/matrix-spec-proposals/pull/2000/)

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -1203,9 +1203,8 @@ forwarded to the login endpoint during the login process. For example:
 #### Notes on password management
 
 {{% boxes/warning %}}
-Clients SHOULD enforce that the password provided is suitably complex.
-The password SHOULD include a lower-case letter, an upper-case letter, a
-number and a symbol and be at a minimum 8 characters in length. Servers
+Clients SHOULD enforce that the password provided is suitably strong.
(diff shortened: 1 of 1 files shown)


KitsuneRal:
1. Relaxing the requirements (even SHOULDs) should still go through an MSC process, in my opinion.
2. I don't see how this makes it easier to implement MSC2000. That MSC sets forth its own language to define password complexity. If that language is implemented it has to be done to the extent of the MSC, not to the extent the current spec describes; and if it's not, the MSC is irrelevant to whatever is currently said in the spec.

uhoreg:
Yes, I agree that this should go through the MSC process before being merged.

PR #1239, opened 2022-09-16
Clarify that /context always returns 'event' even if limit is zero

sumnerevans:
Signed-off-by: Sumner Evans <me@sumnerevans.com>

---
name: Clarify that /context always returns 'event' even if limit is zero
about: The spec is not clear that the limit does not apply to the event itself.
title: ''
labels: ''
assignees: ''

---

### Pull Request Checklist

Preview: https://pr1239--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1239.clarification
+++ b/changelogs/client_server/newsfragments/1239.clarification
@@ -1,0 +1,1 @@
+Clarified that /context always returns 'event' even if limit is zero. Contributed by @sumnerevans at @beeper.
--- a/changelogs/client_server/newsfragments/1239.clarification
+++ b/changelogs/client_server/newsfragments/1239.clarification
@@ -1,1 +1,1 @@
-Clarified that /context always returns 'event' even if limit is zero. Contributed by @sumnerevans at @beeper.
+Clarify that `/context` always returns `event` even if `limit` is zero. Contributed by @sumnerevans at @beeper.
--- a/data/api/client-server/event_context.yaml
+++ b/data/api/client-server/event_context.yaml
@@ -60,7 +60,7 @@ paths:
             The maximum number of context events to return. The limit applies
             to the sum of the `events_before` and `events_after` arrays. The
             requested event ID is always returned in `event` even if `limit` is
-            0. Default: 10.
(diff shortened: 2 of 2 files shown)


changelogs/client_server/newsfragments/1239.clarification
```diff
@@ -0,0 +1 @@
+Clarified that /context always returns 'event' even if limit is zero. Contributed by @sumnerevans at @beeper.
```
turt2live, line 1:
```suggestion
Clarify that `/context` always returns `event` even if `limit` is zero. Contributed by @sumnerevans at @beeper.
```

data/api/client-server/event_context.yaml
```diff
@@ -57,7 +57,10 @@ paths:
           type: integer
           name: limit
           description: |-
-            The maximum number of events to return. Default: 10.
+            The maximum number of context events to return. The limit applies
+            to the sum of the `events_before` and `events_after` arrays. The
+            requested event ID is always returned in `event` even if `limit` is
+            0. Default: 10.
```
turt2live, line 63:
```suggestion
            0. Defaults to 10.
```

data/api/client-server/event_context.yaml
```diff
@@ -57,7 +57,10 @@ paths:
           type: integer
           name: limit
           description: |-
-            The maximum number of events to return. Default: 10.
+            The maximum number of context events to return. The limit applies
+            to the sum of the `events_before` and `events_after` arrays. The
```
turt2live, line 61: is there something to prove that this is true? It's generally sensible, though we do need evidence that this PR is a bug fix rather than a change.

turt2live (commented):
generally seems sane to me - thanks!

data/api/client-server/event_context.yaml
richvdh: It's what synapse does: https://github.com/matrix-org/synapse/blob/v1.72.0/synapse/handlers/room.py#L1365-L1366

Let's just get this shipped.

richvdh (approved):

PR #1243, opened 2022-09-21
Fix typos in the spec related to account data

anoadragon453:
This PR:

* Only use an underscore between 'account' and 'data' when referring to a field.
* Correct some grammar, and one instance of "set" that should have been "get".
* Add markdown links to /sync.

Preview: https://pr1243--matrix-spec-previews.netlify.app


uhoreg (approved):
Looks reasonable

PR #1260, opened 2022-10-01
Fix definition of `data-mx-color` values

progval:
Preview: https://pr1260--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1260.clarification
+++ b/changelogs/client_server/newsfragments/1260.clarification
@@ -1,0 +1,1 @@
+Fix definition of `data-mx-color` and `data-mx-bg-color` values
--- a/changelogs/client_server/newsfragments/1260.clarification
+++ b/changelogs/client_server/newsfragments/1260.clarification
@@ -1,1 +1,1 @@
-Fix definition of `data-mx-color` and `data-mx-bg-color` values
+Fix various typos throughout the specification.
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -47,8 +47,9 @@ Not all attributes on those tags should be permitted as they may be
 avenues for other disruption attempts, such as adding `onclick` handlers
 or excessively large text. Clients should only permit the attributes
 listed for the tags below. Where `data-mx-bg-color` and `data-mx-color`
-are listed, clients should translate the value (a 6-character hex color
-code) to the appropriate CSS/attributes for the tag.
+are listed, clients should translate the value (a `#` character followed
+by a 6-character hex color code) to the appropriate CSS/attributes for
+the tag.
 
 `font`
 `data-mx-bg-color`, `data-mx-color`, `color`


progval:
Signed-off-by: Valentin Lorentz <progval+git @ progval.net>

changelogs/client_server/newsfragments/1260.clarification
```diff
@@ -0,0 +1 @@
+Fix definition of `data-mx-color` and `data-mx-bg-color` values
```
turt2live, line 1:
```suggestion
Fix various typos throughout the specification.
```

turt2live (commented):
otherwise lgtm

changelogs/client_server/newsfragments/1260.clarification
progval: It's not just a typo though; this PR changes the meaning of the spec

turt2live: It doesn't: A colour code is somewhat universally understood to be a `#` with 3-6 hex characters following, however the Matrix spec wanted to clarify that it wanted the 6 character variant specifically, forgetting to reference the `#` for added clarity. This makes it a typo change rather than a spec change (otherwise it'd have required an MSC or larger discussion).

Or in other words, the intent of the spec was to imply that `#` is required even if not specifically mentioned.

progval: That's why I put it in the "Clarification" section instead of "Breaking changes".

I wrote my own client by following the spec, so it implemented this without the `#`.

If someone else wrote this PR with "Fix various typos throughout the specification." in the changelog; then I wouldn't notice this entry when reading the changelog, even though it would be useful to me.

https://spec.matrix.org/v1.4/changelog/#client-server-api has other examples of changes that would be breaking if non-negligeable amounts of implementations followed the spec to the letter in the "Spec Clarifications" section that are not referred to as typoes, such as `Clarify that policy rule globs work like ACL globs`

turt2live: There's a chasm of difference between the ACL behaviour and colour codes :(

Colour codes have their own grammar outside of Matrix and the spec is/was leaning on that (implied) grammar: a quick search on the internet for "html 6 character colour code" reveals a fairly universal language of `#RRGGBB` and still calling it a 6 character (or symbol) code, despite obviously being 7 in length.

A non-typo changelog would be more appropriate if we also defined which characters were red, green, and blue (and what "hex" meant, potentially), however doing so would be well out of scope of Matrix and thus rejected as a change. As it stands, there's already an argument that this change to include the `#` isn't *strictly* required, though I do agree it adds enough value to be merged, just not with its own dedicated changelog entry :)

progval: > a quick search on the internet for "html 6 character colour code" reveals a fairly universal language of #RRGGBB and still calling it a 6 character (or symbol) code, despite obviously being 7 in length.

That's for HTML, and most websites do that because they are aimed to web developers, but `data-mx-color` and `data-mx-bg-color` are not part of HTML.

In other contexts, such as IM, the `#` is usually absent: [IRC](https://modern.ircdocs.horse/formatting.html#hex-color), [XMPP](https://xmpp.org/extensions/xep-0331.html)

> A non-typo changelog would be more appropriate if we also defined which characters were red, green, and blue (and what "hex" meant, potentially)

I don't get it. Isn't the purpose of a changelog to tell developers implementing Matrix what changes they should pay attention to?

My change replaces a description which is either incorrect or ambiguous, because at least one person (me) misunderstood it. Shouldn't the changelog tell other developers who potentially misunderstood it as well they should check this change out?

turt2live: It ultimately maps to an HTML/CSS value which would require a `#`. Honestly, this is the sort of change which ends up on or near the fence of what is a typo and what is not - here I think it still ends up being a typo considering the scope of the change.

turt2live (approved):
thanks!

Issue #578, opened 2020-01-13
Typing CS documentation contains mistake about the room_id


bmarty:
In https://matrix.org/docs/spec/client_server/latest#id49, the part:

> but still have a room_id key

is not correct.

Also in the sample of https://matrix.org/docs/spec/client_server/latest#m-typing:

```json
{
    "content": {
        "user_ids": [
            "@alice:matrix.org",
            "@bob:example.com"
        ]
    },
    "room_id": "!jEsUZKDJdhlrceRyVU:example.org",
    "type": "m.typing"
}
```

contains `room_id` field, which is actually not the case.

The typing event is already in the rooms.join.{room_id}.ephemeral.events list of the sync response, so `room_id` is not necessary in the event content. And it's actually not there. The issue is just in the documentation.

richvdh:
in https://spec.matrix.org/v1.3/client-server-api/#typing-notifications, looks like the example has been updated but not the text.

PR #1265, opened 2022-10-04
Remove reference to a `room_id` key for typing events

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/578

Typing events do not contain a `room_id` key. They are instead scoped to a room, for example under a specific room's ephemeral events listing in a [`/sync`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixclientv3sync) response.

Removing the reference to `room_id` helps reduce confusion.

Preview: https://pr1265--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/typing_notifications.md
+++ b/content/client-server-api/modules/typing_notifications.md
@@ -7,7 +7,7 @@ type: module
 Users may wish to be informed when another user is typing in a room.
 This can be achieved using typing notifications. These are ephemeral
 events, so they do not form part of the
-[Event Graph](index.html#event-graphs), even though they are scoped
+[Event Graph](index.html#event-graphs). Typing notifications are scoped
 to a room.
 
 #### Events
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/typing_notifications.md
```diff
@@ -6,8 +6,8 @@ type: module
 
 Users may wish to be informed when another user is typing in a room.
 This can be achieved using typing notifications. These are ephemeral
-events scoped to a `room_id`. This means they do not form part of the
-[Event Graph](index.html#event-graphs) but still have a `room_id` key.
+events scoped to a room, but they do not form part of the
```
richvdh, line 9: ephemeral events inherently don't form part of the event graph, so "but" is misleading here. 

```suggestion
events scoped to a room, so they do not form part of the
```

richvdh (approved):
lgtm otherwise

Issue #1112, opened 2022-06-07
Auth rules in all room versions talk about "keys" but the documentation style wants them to be called "properties"


turt2live:
https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#stylistic-notes
https://spec.matrix.org/v1.2/rooms/v9/#authorization-rules (and all other room versions)

turt2live:
while we're here, fix the reference to "dictionaries" and instead call them "objects".

PR #1269, opened 2022-10-04
Clarify auth rules for `m.room.power_levels` events

### LINKED ISSUE #699: Auth rules wrongly claim that users can demote others of the same power level
The auth rules (v1 as a base) only consider a case where the user's power level is *higher*, but *higher* is somewhat ambiguous. It should read 'equal to or greater than'.

For reference, here's the current auth rules:
image

When deleting an entry (not changing it), 10d triggers but not 10e.

richvdh:
General clarifications, but in particular: fix some misleading text which implied you could demote other users with the same PL.

Fixes #699.

Preview: https://pr1269--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1269.clarification
+++ b/changelogs/room_versions/newsfragments/1269.clarification
@@ -1,0 +1,1 @@
+Reword the event auth rules to clarify that users cannot demote other users with the same power level.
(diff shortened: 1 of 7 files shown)


turt2live:
Looks like this also fixes https://github.com/matrix-org/matrix-spec/issues/1112

richvdh:
> Looks like this also fixes #1112

it's a start. Other bits of the auth rules still mention "keys".

erikjohnston (approved):

Issue #18, opened 2022-01-31
In-room m.key.verification.request message event doesn't seem to be documented


jplatte:
**Link to problem area**: None

**Issue**

I can't find any part of the v1.1 spec that mentions that `m.key.verification.request` is a possible `msgtype` for `m.room.message` and that the `m.key.verification.request` reference in `relates_to` is actually the event ID of an `m.room.message` event (not a `m.key.verification.request` event).

PR #1271, opened 2022-10-04
add in-room m.key.verification.request msgtype

uhoreg:
fixes https://github.com/matrix-org/matrix-spec/issues/18

ref: https://github.com/matrix-org/matrix-spec-proposals/pull/2241

Apparently, when in-room verification was added, the `m.key.verification.request` msgtype was forgotten.

Preview: https://pr1271--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -565,7 +565,7 @@ in this specification must be under the `m.key.verification` namespace
 and any other event types must be namespaced according to the Java
 package naming convention.
 
-{{% event event="m.room.message$m.key.verification.request" %}}
+{{% event event="m.room.message$m.key.verification.request" title="`m.room.message` with `msgtype: m.key.verification.request`" %}}
 
 {{% event event="m.key.verification.request" %}}
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -434,11 +434,14 @@ shared by the other messages of that session.
 
 In general, verification operates as follows:
 
-- Alice requests a key verification with Bob by sending an
-  `m.key.verification.request` event. This event indicates the verification
-  methods that Alice's client supports. (Note that "Alice" and "Bob" may in
-  fact be the same user, in the case where a user is verifying their own
-  devices.)
+- Alice requests a key verification with Bob by sending an key verification
+  request event.  If the verification is being requested in a room, this will
+  be an event with type `m.room.message` and `msgtype`
+  `m.key.verification.request`; if the verification is being requested using
+  to-device messaging, this will be an event with type
+  `m.key.verification.request`.  This event indicates the verification methods
+  that Alice's client supports. (Note that "Alice" and "Bob" may in fact be the
+  same user, in the case where a user is verifying their own devices.)
 - Bob's client prompts Bob to accept the key verification. When Bob accepts
   the verification, Bob's client sends an `m.key.verification.ready` event.
   This event indicates the verification methods, corresponding to the
@@ -561,6 +564,8 @@ properties to the verification messages. Event types for methods defined
 in this specification must be under the `m.key.verification` namespace
 and any other event types must be namespaced according to the Java
 package naming convention.
+
+{{% event event="m.room.message$m.key.verification.request" %}}
 
 {{% event event="m.key.verification.request" %}}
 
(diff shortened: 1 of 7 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -434,11 +434,14 @@ shared by the other messages of that session.
 
 In general, verification operates as follows:
 
-- Alice requests a key verification with Bob by sending an
-  `m.key.verification.request` event. This event indicates the verification
-  methods that Alice's client supports. (Note that "Alice" and "Bob" may in
-  fact be the same user, in the case where a user is verifying their own
-  devices.)
+- Alice requests a key verification with Bob by sending an key verification
```
richvdh, line 437:
```suggestion
- Alice requests a key verification with Bob by sending a key verification
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -434,11 +434,14 @@ shared by the other messages of that session.
...
+  request event.  If the verification is being requested in a room, this will
+  be an event with type `m.room.message` and `msgtype`
+  `m.key.verification.request`; if the verification is being requested using
+  to-device messaging, this will be an event with type
+  `m.key.verification.request`.  This event indicates the verification methods
+  that Alice's client supports. (Note that "Alice" and "Bob" may in fact be the
+  same user, in the case where a user is verifying their own devices.)
```
richvdh, line 444: It would be nice to link to the definitions of these things, but maybe if we were doing that, we'd want to do it for the whole section, and it becomes a major task.

data/event-schemas/schema/m.room.message$m.key.verification.request.yaml
```diff
@@ -0,0 +1,67 @@
...
+  Requests a key verification in a room.  When requesting a key verification
+  using to-device messaging, an event with type [`m.key.verification.request`](#mkeyverificationrequest)
+  should be used.
+properties:
+  content:
+    properties:
+      body:
+        type: string
+        description: |-
+          a fallback message to alert users that their client does not support
```
richvdh, line 14:
```suggestion
          A fallback message to alert users that their client does not support
```

data/event-schemas/examples/m.room.message$m.key.verification.request.yaml
```diff
@@ -0,0 +1,19 @@
+{
```
richvdh, line 1: this makes an example appear under the general definition of `m.room.message` (https://pr1271--matrix-spec-previews.netlify.app/client-server-api/#mroommessage), which feels confusing to me.

Worse, where there *ought* to be an example, is just an empty "Examples" heading:

image

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -562,6 +565,8 @@ in this specification must be under the `m.key.verification` namespace
 and any other event types must be namespaced according to the Java
 package naming convention.
 
+{{% event event="m.room.message$m.key.verification.request" %}}
```
richvdh, line 568: this presents itself as literally `m.room.message$m.key.verification.request`, which is obscure to say the least:

image

It's also in conflict to the way that `m.text` and co are represented, though I'm not sure that format would be terribly useful either:

image

Maybe the `render-event.html` template thing needs to be updated to honour the `title` of the event definition, then we can set that to something like "`m.room.message` with `msgtype: m.key.verification.request`" ?

data/event-schemas/schema/m.room.message$m.key.verification.request.yaml
```diff
@@ -0,0 +1,67 @@
...
+          this event.
+        type: string
+      msgtype:
+        enum:
+          - m.key.verification.request
+        type: string
+    required:
+      - from_device
+      - methods
+      - msgtype
```
richvdh, line 60: should this not include `to` ?

uhoreg:
```suggestion
    required:
      - from_device
      - methods
      - msgtype
      - to
```

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: This and https://github.com/matrix-org/matrix-spec/pull/1271#discussion_r992649121 will be fixed by https://github.com/matrix-org/matrix-spec/pull/1283

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: links added

richvdh (approved):
lgtm

Issue #1212, opened 2022-08-17
knock_room_state is used in an example, but never explained properly


jplatte:
**Link to problem area**: https://spec.matrix.org/v1.3/client-server-api/#mroommember

**Issue**

One of the examples for `m.room.member` contains a `knock_room_state` field in `unsigned`, but this is not mentioned anywhere else. It seems similar to `invite_room_state` which is also only explained very briefly (not listed in one of the tables, just mentioned in the text), if it actually exists rather than just being an artifact that was never meant to enter the spec.

anoadragon453:
I believe that it's an error that `invite_room_state` should be known to the client, see https://github.com/matrix-org/matrix-spec/issues/848.

The examples for `invite_room_state` and `knock_room_state` should be removed from the CS API spec. The linked issue covers `invite_room_state`, whereas this issue covers `knock_room_state`.

Note: `invite_room_state` is only included in the `unsigned` field of a `m.room.member` event when using the v1 Federation invite endpoint: https://spec.matrix.org/v1.4/server-server-api/#put_matrixfederationv1inviteroomideventid

[In v2](https://spec.matrix.org/v1.4/server-server-api/#put_matrixfederationv2inviteroomideventid), stripped invite state has moved to a field in the request body.

Still, it doesn't belong in the CS API.

Issue #848, opened 2021-06-18
Clarification needed in invite stripped state


t3chguy:
Two areas of the spec doc talk about invite stripped state:

https://matrix.org/docs/spec/client_server/r0.6.1#m-room-member
`invite_room_state` in the Unsigned of your own `m.room.member` event

and

https://matrix.org/docs/spec/client_server/r0.6.1#get-matrix-client-r0-sync
`rooms.invite.<room_id>.invite_state`

The stored event has the `invite_room_state` field but Synapse strips it when including the membership event in `invite_state`. https://github.com/matrix-org/synapse/blob/develop/synapse/rest/client/v2_alpha/sync.py#L326-L341
The `/state` API doesn't work in Synapse pre-join so it seems like a client has no way to ever see a `invite_room_state` in the real world.

![image](https://user-images.githubusercontent.com/2403652/122575187-62dff800-d048-11eb-9ee9-d2c9cb0f44de.png)

js-sdk respects `invite_state` but also expects a `invite_room_state` at the top level (sibling of unsigned) of an event for whatever reason which is obviously wrong.

anoadragon453:
Indeed, Synapse does not make this field accessible to clients. Clients should instead make use of `invite_state` in a `/sync` response.

`invite_room_state` is included in the `unsigned` field of a `m.room.member` event when using the v1 Federation invite endpoint: https://spec.matrix.org/v1.4/server-server-api/#put_matrixfederationv1inviteroomideventid

[In v2](https://spec.matrix.org/v1.4/server-server-api/#put_matrixfederationv2inviteroomideventid), stripped invite state has moved to a field in the request body.

I think this all happened because Synapse does a bit of a hack to store stripped state that it receives from an invite over federation, storing it in the `unsigned` field of an event and then popping that field whenever it needs to return that event to a client or server. The only area it doesn't pop that field is when sending events to application services.

Regardless, it should be removed from the CS spec.

zecakeh:
For some follow up here, it seems that the field is actually present for appservices, and it was decided in https://github.com/matrix-org/matrix-spec/pull/1273#issuecomment-1285556106 that an MSC should be created to move the fields from the CS API to the AS API.

PR #1273, opened 2022-10-05
Remove `invite_room_state`, `knock_room_state` from the CS API. Fix examples in SS API

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/1212
Fixes https://github.com/matrix-org/matrix-spec/issues/848

More context in the linked issues. This PR makes a couple changes:

* It removes any reference to `invite_room_state` or `knock_room_state` on `m.room.member` in the CS API. Clients should instead be receiving this information down [/sync](http://localhost:1313/client-server-api/#get_matrixclientv3sync), under `rooms.invite.!room.invite_state`.
* It removes the `unsigned.invite_room_state` from the [`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`](http://localhost:1313/server-server-api/#put_matrixfederationv2inviteroomideventid) response example.
    * This wasn't specced, but Synapse is returning it, which may be why it is here. I believe that's a bug though: https://github.com/matrix-org/synapse/pull/14064


data/event-schemas/schema/m.room.member.yaml
```diff
@@ -136,28 +132,5 @@ properties:
...
-          invite_room_state:
-            description: |-
-              A subset of the state of the room at the time of the invite, if `membership` is `invite`.
-              Note that this state is informational, and SHOULD NOT be trusted; once the client has
-              joined the room, it SHOULD fetch the live state from the server and discard the
-              invite_room_state. Also, clients must not rely on any particular state being present here;
-              they SHOULD behave properly (with possibly a degraded but not a broken experience) in
-              the absence of any particular events here. If they are set on the room, at least the
-              state for `m.room.avatar`, `m.room.canonical_alias`, `m.room.join_rules`, and `m.room.name`
-              SHOULD be included.
```
anoadragon453, line 149: It's a shame to lose this explanation. I suppose it could be moved elsewhere, perhaps https://spec.matrix.org/v1.4/client-server-api/#_matrixclientv3sync_invitestate?

turt2live:
What if the client isn't syncing - how is it supposed to get this information? I'm not so sure it's a spec bug that these fields exist, at least.

turt2live:
Appservices who don't sync :)

anoadragon453:
Ahh. OK, yes, appservices do receive this information from Synapse via `unsigned.invite_room_state` today.

It still feels odd to include this field in the CS API spec when clients have no way to ever see it. Perhaps it should just be an extension in the AS API spec? Are application services the only consumers of this field?

turt2live:
It might be the mandela effect, but I swear clients had/have access to this via `unsigned` too. It sort of feels like a thing which might need an MSC? (at least to identify how clients get access to this if they aren't `/sync`ing, such as when previewing a room they were linked to or whatever)

data/event-schemas/schema/m.room.member.yaml
anoadragon453: Actually, this advice is already included under https://spec.matrix.org/v1.4/client-server-api/#stripped-state.

anoadragon453:
> It might be the mandela effect, but I swear clients had/have access to this via `unsigned` too.

It *shouldn't* have been the case in Synapse, at least. We try to strip the field before sending it to clients in all cases: https://github.com/matrix-org/synapse/blob/06df5d4250f54d5a95b0c90bfc9352ec6f02c520/synapse/events/utils.py#L371-L377

The only case where this conditional doesn't `pop` the entries is for sending to application services, and down `/sync` (but later on in the /sync code [we `pop` the field anyways](https://github.com/matrix-org/synapse/blob/b4ec4f5e71a87d5bdc840a4220dfd9a34c54c847/synapse/rest/client/sync.py#L368)).

> It sort of feels like a thing which might need an MSC? (at least to identify how clients get access to this if they aren't `/sync`ing, such as when previewing a room they were linked to or whatever)

It could perhaps all just be replaced by [`GET /_matrix/client/v1/rooms/{roomId}/hierarchy`](https://spec.matrix.org/v1.4/client-server-api/#get_matrixclientv1roomsroomidhierarchy) or [MSC3266](https://github.com/matrix-org/matrix-spec-proposals/pull/3266) (room summary API), which both return stripped state events?

richvdh:
> > It might be the mandela effect, but I swear clients had/have access to this via `unsigned` too.
>
> It _shouldn't_ have been the case in Synapse, at least. We try to strip the field before sending it to clients in all cases: [matrix-org/synapse@`06df5d4`/synapse/events/utils.py#L371-L377](https://github.com/matrix-org/synapse/blob/06df5d4250f54d5a95b0c90bfc9352ec6f02c520/synapse/events/utils.py#L371-L377)

The obvious question here is: how would clients get hold of the invite event anyway? It doesn't really matter if they would see the room state via `unsigned.invite_room_state` if they can't see the event itself, other than via `/sync` (which, as noted, exposes the info in a better way.

One thought: is the whole thing a holdover from old-style `/events`?

> The only case where this conditional doesn't `pop` the entries is for sending to application services, and down `/sync` (but later on in the /sync code [we `pop` the field anyways](https://github.com/matrix-org/synapse/blob/b4ec4f5e71a87d5bdc840a4220dfd9a34c54c847/synapse/rest/client/sync.py#L368)).

I think it's worth noting that, up until https://github.com/matrix-org/synapse/pull/6739/files#diff-cd497404c31b1cbaa0bcc442a4c10f1151d151c01e1208e40aede55635e9ab8eL325, there was a comment:

>  If this is an invite **for somebody else**, then we don't care about the invite_room_state

(emphasis mine), which adds some background on the intention here.

Still, I'm not sure we can just rip it out - particularly for appservices - without an MSC.

richvdh (commented):
^

richvdh:
yes, an MSC sounds good...

richvdh:
@anoadragon453 is this still on your radar?

anoadragon453:
@richvdh Not really I'm afraid, it's definitely fallen off.

richvdh:
Closing for now then, I guess.

PR #1276, opened 2022-10-06
Fix up description of `knock_room_state` field, which implied the required field was optional

anoadragon453:
[MSC2403](https://github.com/matrix-org/matrix-spec-proposals/pull/2403) states that [this field is required](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2403-knock.md#put-_matrixfederationv1send_knockroomideventid). I'm not sure I actually agree with that, but it's what the MSC says.

Before this PR, the description of this field looked like:

image

This PR drops the optional part:

image

Preview: https://pr1276--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1276.clarification
+++ b/changelogs/client_server/newsfragments/1276.clarification
@@ -1,0 +1,1 @@
+Fix various typos throughout the specification.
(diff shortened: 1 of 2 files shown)


turt2live (approved):
This would have been a copy/paste fail - good spot!

PR #1278, opened 2022-10-11
Remove `origin` field from transactions

### LINKED ISSUE #374: Make transaction.origin and pdu.origin optional/go away

jplatte:
Part of #374. According to latest comments there, the inclusion of `origin` can likely be considered a spec bug, which is why this isn't an MSC with a migration plan or anything like that.

Preview: https://pr1278--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/1278.clarification
+++ b/changelogs/server_server/newsfragments/1278.clarification
@@ -1,0 +1,1 @@
+Remove `origin` field from transactions which was redundant and not enforced to be present by Synapse.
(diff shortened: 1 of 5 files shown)


anoadragon453 (commented):
Indeed this field is redundant when we already have an `origin` field in the Authorization header that's sent along with a transaction.

The changeset looks good. @richvdh I'm not entirely sure from conversation in https://github.com/matrix-org/matrix-spec/issues/374, but do you consider this fine to remove without an MSC?

richvdh:
Er no, I think this needs an MSC. There may exist server implementations which rely on `transaction.origin`.

> According to latest comments there, the inclusion of origin can likely be considered a spec bug, which is why this isn't an MSC with a migration plan or anything like that.

I think this is referring to https://github.com/matrix-org/matrix-spec/issues/374#issuecomment-1065343037. I'm having a hard time deciphering that myself, but have attempted to clarify on #374.

jplatte:
Okay, in that case I'll close this (for now).

Issue #897, opened 2021-09-20
Spec refers to EDUs and account data as "events"


richvdh:

jplatte:
Re. EDUs, what events count as that actually? Some or all of these should be:

* Presence, metadata: `sender`
* Ephemeral room events (`m.receipt`, `m.typing`)
* To-device events, metadata: `sender`

Also, room account data events and ephemeral room events have the `room_id` as metadata, conceptually? It's just that they're only ever received through `/sync` where they are grouped by room ID so it can be omitted.

richvdh:
EDUs are exactly those you list: presence, receipts, typing, to-device. They are listed in https://spec.matrix.org/v1.4/server-server-api, though not in a way that makes them very easy to find.

(Generally I am not a fan of the whole EDU concept: it gives the false impression that there is something in common between those four mechanisms other than the fact they all appear in the `edus` property in a federation `/send` request. In reality, each has its own special rules for how servers should handle them. So why don't they just have their own properties in `/send`?)

richvdh:
related: https://github.com/matrix-org/matrix-spec/issues/1241

turt2live:
> So why don't they just have their own properties in `/send`?

fwiw, crypto appears to do this in places like `/sync`, and it feels messier than it actually is. Developer perception for them being all the same schema I think would help, but only if they are actually the same sort of thing (beyond an arbitrary heading).

turt2live:
even more confusing is the [Client-Server API refers](https://spec.matrix.org/v1.13/client-server-api/#send-to-device-messaging) to to-device messages as both "messages" and "events", going as far as including an event type in the URL. When this gets translated to federation, the [API changes](https://spec.matrix.org/v1.13/server-server-api/#send-to-device-messaging) to be an `m.direct_to_device` EDU containing sub-events, breaking the single datum principle a bit.

PR #1283, opened 2022-10-12
improve display of event subtypes

uhoreg:
improvements to `m.<event type>$<subtype>` events

- examples are now shown under the subtype. In addition, `m.room.message` only shows one example, rather than a whole bunch
- allow setting the title to something that looks more sensible

fixes https://github.com/matrix-org/matrix-spec/issues/815

Preview: https://pr1283--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -736,7 +736,7 @@ following error codes are used in addition to those already specified:
 -   `m.mismatched_commitment`: The hash commitment did not match.
 -   `m.mismatched_sas`: The SAS did not match.
 
-{{% event event="m.key.verification.start$m.sas.v1" %}}
+{{% event event="m.key.verification.start$m.sas.v1" title="`m.key.verification.start` with `method: m.sas.v1`" %}}
 
 {{% event event="m.key.verification.accept" %}}
 
@@ -1145,7 +1145,7 @@ base64).
 
 ###### Verification messages specific to QR codes
 
-{{% event event="m.key.verification.start$m.reciprocate.v1" %}}
+{{% event event="m.key.verification.start$m.reciprocate.v1" title="`m.key.verification.start` with `method: m.reciprocate.v1`" %}}
 
 #### Sharing keys between devices
 
--- a/layouts/partials/events/example.html
+++ b/layouts/partials/events/example.html
@@ -1,4 +1,13 @@
-{{ $example_content := partial "json-schema/resolve-refs" (dict "schema" .example "path" "event-schemas/examples") }}
+{{/*
+
+  Renders an event example.  Resolves `$ref`s, serializes as JSON, and ensures
+  that it can be included in HTML.
+
+  This partial is called with the example data as its context.
+
+*/}}
+
+{{ $example_content := partial "json-schema/resolve-refs" (dict "schema" . "path" "event-schemas/examples") }}
 {{ $example_json := jsonify (dict "indent" "  ") $example_content }}
 {{ $example_json = replace $example_json "\\u003c" "<" }}
 {{ $example_json = replace $example_json "\\u003e" ">" | safeHTML }}
--- a/layouts/partials/events/example.html
+++ b/layouts/partials/events/example.html
@@ -1,0 +1,8 @@
+{{ $example_content := partial "json-schema/resolve-refs" (dict "schema" .example "path" "event-schemas/examples") }}
+{{ $example_json := jsonify (dict "indent" "  ") $example_content }}
+{{ $example_json = replace $example_json "\\u003c" "<" }}
+{{ $example_json = replace $example_json "\\u003e" ">" | safeHTML }}
+
+```json
+{{ $example_json }}
+```
--- a/layouts/partials/events/example.html
+++ b/layouts/partials/events/example.html
@@ -3,7 +3,7 @@
   Renders an event example.  Resolves `$ref`s, serializes as JSON, and ensures
   that it can be included in HTML.
 
-  This partial is called with the example data as its context.
+  This partial is called with the example event object as its context.
 
 */}}
 
--- a/layouts/partials/events/render-event.html
+++ b/layouts/partials/events/render-event.html
@@ -8,7 +8,7 @@
     If `desired_example_name` is omitted we render all examples
     whose names start with the `event_name`.
     * `title` (optional): the title to display.  May contain markdown.  Defaults to
-    `event_name` wrapped in a <code> element
+    `event_name` wrapped in a <code> element.
 
 */}}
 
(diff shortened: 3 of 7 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -736,7 +736,7 @@ following error codes are used in addition to those already specified:
 -   `m.mismatched_commitment`: The hash commitment did not match.
 -   `m.mismatched_sas`: The SAS did not match.
 
-{{% event event="m.key.verification.start$m.sas.v1" %}}
+{{% event event="m.key.verification.start$m.sas.v1" title="`m.key.verification.start` with `method: m.sas.v1`" %}}
```
richvdh, line 739: Any reason not to put this title in the definition itself? jsonschema defines a `title` attribute which seems like it would be ideal (https://json-schema.org/draft/2020-12/json-schema-validation.html#name-title-and-description)

layouts/partials/events/render-event.html
```diff
@@ -20,7 +22,7 @@
 <summary>
 
   <h1 id="{{ anchorize $event_name }}">
-   <code>{{ $event_name }}</code>
+   {{ with .title }}{{ $.title | markdownify }}{{ else }}<code>{{ $event_name }}</code>{{ end }}
```
richvdh, line 25: I think you can

```suggestion
   {{ with .title }}{{ . | markdownify }}{{ else }}<code>{{ $event_name }}</code>{{ end }}
```

?

layouts/partials/events/example.html
```diff
@@ -0,0 +1,8 @@
+{{ $example_content := partial "json-schema/resolve-refs" (dict "schema" .example "path" "event-schemas/examples") }}
```
richvdh, line 1: please can you give this a comment? what does it do, what is the input?

layouts/partials/events/render-event.html
```diff
@@ -72,11 +74,10 @@ <h2>Examples</h2>
     */}}
     {{ if $desired_example_name }}
         {{ if eq $example_name $desired_example_name }}
-            {{ $example_content := partial "json-schema/resolve-refs" (dict "schema" $example "path" "event-schemas/examples") }}
-```json
-{{ jsonify (dict "indent" "  ") $example_content }}
-```
+            {{ partial "events/example" (dict "example" $example)}}
```
richvdh, line 77: given this only takes a single parameter, it might be easier to pass it directly, rather than wrap it in a dict?

```suggestion
            {{ partial "events/example" $example }}
```

layouts/partials/events/render-event.html
```diff
@@ -72,11 +74,10 @@ <h2>Examples</h2>
...
         {{ end }}
+    {{ else if eq $event_name $example_name }}
+        {{ partial "events/example" (dict "example" $example)}}
```
richvdh, line 80: Looks like this is to cover the case that `event_name` has a `$`, so will never be matched by the case below? Could you add a comment to say so?

layouts/partials/events/render-event.html
```diff
@@ -72,11 +74,10 @@ <h2>Examples</h2>
...
     {{/*
       If `$desired_example_name` is not given, we will include any
       examples whose first part (before "$") matches the event name
```
richvdh, line 83: (From the description of the PR)

> (examples are still shown under the main type, but are now additionally shown under the subtype)

I'm far from convinced this is useful. Having 15 different examples under https://pr1283--matrix-spec-previews.netlify.app/client-server-api/#mroommessage doesn't seem to add much, particularly when the examples are all duplicated later. What if we just used the `m.text` example (via `desired_example_name`, presumably)?

layouts/partials/events/render-event.html
richvdh: (assuming you keep the case below)

richvdh (changes requested):

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: mainly because some event definitions already use the `title` attribute (e.g. https://github.com/matrix-org/matrix-spec/blob/main/data/event-schemas/schema/m.room.power_levels.yaml) and set it to something other than what the current title is in the spec.

richvdh: The example you give looks like it should be the `description` rather than the `title`.

Given we know that none of the titles are rendered, maybe we just rip out the existing ones with no loss of value?

Sorry to make the job (even) bigger, but I feel like this is going to be a maintenance headache we should just solve rather than hack around for now

uhoreg: I tried doing this, and ran into a snag.  All the events inherit (using `allOf`, usually indirectly) from `core-event-schema/event.yaml`, so I need to remove all the `title` properties from everything in the chain.  However, removing the title property from `event.yaml` turns things like

image

becomes

image

:frowning_face: 

The only way I can think of to actually get this to work the way we want it to is to modify all the event yaml files to have a `title` property set to the event type (e.g. `title: m.room.message` in `m.room.message.yaml`), which seem ... suboptimal.

richvdh: ugh what a pain in the butt.

It feels like really the problem here is that things like this shouldn't be referring to `event.yaml` anyway, and should use a more specific type (with an appropriate title).

image

(see also my complaints at https://github.com/matrix-org/matrix-spec/issues/897)

... however, maybe fixing that right now is a bit out of scope and we should just stick with your original suggestion.

layouts/partials/events/render-event.html
uhoreg: I tried to do this with https://github.com/matrix-org/matrix-spec/pull/1283/commits/d6345f0df2fec12c08aaadac6fe3c6423a4bf87b#diff-4bdc5bb1f13036061393a5799f5ec426a67bf2850c8083d21e0b649877ff6bc8L13 but `m.room.message` still has all its examples, so I seem to be missing something. :-/

uhoreg: Figured it out.  This is one done now.

layouts/shortcodes/event.html
```diff
@@ -17,4 +17,4 @@
 {{ $event_data = partial "json-schema/resolve-refs" (dict "schema" $event_data "path" $path) }}
 {{ $event_data := partial "json-schema/resolve-allof" $event_data }}
 
-{{ partial "events/render-event" (dict "event_name" .Params.event "event_data" $event_data)}}
+{{ partial "events/render-event" (dict "event_name" .Params.event "event_data" $event_data "desired_example_name" .Params.desired_example_name "title" .Params.title)}}
```
richvdh, line 20: we should document the new params here as well.

richvdh (changes requested):

richvdh (approved):
thanks for your perseverence :heart:

PR #1284, opened 2022-10-12
improve documentation of ephemeral events

uhoreg:
I'm not entirely sure if what I've done here is exactly right, but I think that we need to define ephemeral events in one spot, and then just use the "ephemeral" terminology instead of saying "events that aren't persisted e.g. typing".  Opinions on whether this PR looks good, or whether there's a better place to define it, are welcome.

Addresses part of https://github.com/matrix-org/matrix-spec/issues/136

Preview: https://pr1284--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1284.clarification
+++ b/changelogs/client_server/newsfragments/1284.clarification
@@ -1,0 +1,1 @@
+Improve documentation about ephemeral events.
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -1521,7 +1521,7 @@ any given point in time:
 
 ### Types of room events
 
-Room events are split into three categories:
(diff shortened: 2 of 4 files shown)


richvdh:
I am unenthusiastic, for the reasons set out in https://github.com/matrix-org/matrix-spec/issues/897 (though as @jplatte notes, `m.receipt` and `m.typing` are less egregious than some).

I'd really prefer to see `m.receipt` and `m.typing` called out as particular cases in the `/sync` documentation, rather than mixing them in with the generic "events" documentation.

uhoreg:
Huh.  Should have pressed the "Ready for review" button earlier. :-/

richvdh (approved):
lgtm

PR #1261, opened 2022-10-01
Receipts: Add `thread_id` to the `/receipt` endpoint

zecakeh:
It seems to have been omitted in #1255.

Per [MSC3771](https://github.com/matrix-org/matrix-spec-proposals/pull/3771).

Preview: https://pr1261--matrix-spec-previews.netlify.app

PR #1286, opened 2022-10-13
Define a 400 response from {PUT,GET,DELETE} `/directory/rooms/{roomAlias}`

DMRobertson:
Noticed when comparing Synapse against the spec in https://github.com/matrix-org/synapse/issues/14176.

Not important, maybe a nice-to-have?

Preview: https://pr1286--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1286.clarification
+++ b/changelogs/client_server/newsfragments/1286.clarification
@@ -1,1 +1,1 @@
-Define a 400 response from {PUT,GET,DELETE} `/directory/rooms/{roomAlias}`.
+Define a 400 response from `/_matrix/client/v3/directory/rooms/{roomAlias}`.
--- a/changelogs/client_server/newsfragments/1286.clarification
+++ b/changelogs/client_server/newsfragments/1286.clarification
@@ -1,0 +1,1 @@
+Define a 400 response from {PUT,GET,DELETE} `/directory/rooms/{roomAlias}`.
--- a/data/api/client-server/directory.yaml
+++ b/data/api/client-server/directory.yaml
@@ -37,7 +37,9 @@ paths:
         - in: path
           type: string
           name: roomAlias
-          description: The room alias to set.
+          description: |
+            The room alias to set. Its format is defined
+            [in the appendices](/appendices/#room-aliases).
           required: true
           x-example: "#monkeys:matrix.org"
         - in: body
@@ -95,7 +97,9 @@ paths:
         - in: path
           type: string
           name: roomAlias
-          description: The room alias.
+          description: |
+            The room alias. Its format is defined
+            [in the appendices](/appendices/#room-aliases).
           required: true
           x-example: "#monkeys:matrix.org"
       responses:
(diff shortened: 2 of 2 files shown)


data/api/client-server/directory.yaml
```diff
@@ -61,6 +61,15 @@ paths:
             application/json: {}
           schema:
             type: object
+        400:
+          description: The given `roomAlias` is not a valid room alias.
```
turt2live, line 65: "Define valid". We can probably copy a lot of the text from https://github.com/matrix-org/matrix-spec/pull/1261 and use it here.

Generally though, I'm more in favour of fixing the spec to say that 400 `M_INVALID_PARAM` is something which can appear on every endpoint rather than each endpoint individually. This is likely a larger effort than this PR should take on, however.

changelogs/client_server/newsfragments/1286.clarification
```diff
@@ -0,0 +1 @@
+Define a 400 response from {PUT,GET,DELETE} `/directory/rooms/{roomAlias}`.
```
turt2live, line 1:
```suggestion
Define a 400 response from `/_matrix/client/v3/directory/rooms/{roomAlias}`.
```

(we don't typically define the http methods when it's all of them)

turt2live (commented):
otherwise lgtm

data/api/client-server/directory.yaml
DMRobertson: I would ideally link to https://spec.matrix.org/latest/appendices/#room-aliases here, but I couldn't find an example of such a hyperlink from the OpenAPI spec.

DMRobertson: There is https://github.com/matrix-org/matrix-spec/blob/dcc1f4adee2be41069045ddbf3ef913e9c19d116/data/api/client-server/definitions/one_time_keys.yaml#L21 but this a) is hardcoded to a spec version and b) doesn't appear in the spec (#1287)

DMRobertson: Oh, directly above this:

https://github.com/matrix-org/matrix-spec/blob/dcc1f4adee2be41069045ddbf3ef913e9c19d116/data/api/client-server/definitions/one_time_keys.yaml#L7

That'll do.

DMRobertson:
I have snuck-in a recommendation that clients URL-encode the hash character: [`36aeb1c` (#1286)](https://github.com/matrix-org/matrix-spec/pull/1286/commits/36aeb1c689bca2579bf15eff7fe5b6b7cee8971b)

turt2live:
> I have snuck-in a recommendation that clients URL-encode the hash character: [`36aeb1c` (#1286)](https://github.com/matrix-org/matrix-spec/pull/1286/commits/36aeb1c689bca2579bf15eff7fe5b6b7cee8971b)

I don't think that's a thing that should be snuck in here - it sounds like "do http correctly" which is out of scope for our spec.

DMRobertson:
Fair enough; that's reverted.

turt2live:
(please don't force push after review is given - it makes things harder on everyone)

data/api/client-server/directory.yaml
turt2live: > There is
> 
> https://github.com/matrix-org/matrix-spec/blob/dcc1f4adee2be41069045ddbf3ef913e9c19d116/data/api/client-server/definitions/one_time_keys.yaml#L21
> 
> but this a) is hardcoded to a spec version and b) doesn't appear in the spec (#1287)

This looks like a bug, ftr. Will take a look.

turt2live (approved):

PR #1289, opened 2022-10-14
Fix /refresh endpoint to be V1 instead of V3

TobiasFella:
The implementations in synapse and matrix-js-sdk use this as /v1/refresh, which, according to @KitsuneRal, is correct

Signed-off-by: Tobias Fella <fella@posteo.de>

Preview: https://pr1289--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -403,7 +403,7 @@ Access tokens can expire after a certain amount of time. Any HTTP calls that
 use an expired access token will return with an error code `M_UNKNOWN_TOKEN`,
 preferably with `soft_logout: true`. When a client receives this error and it
 has a refresh token, it should attempt to refresh the access token by calling
-[`/refresh`](#post_matrixclientv3refresh). Clients can also refresh their
+[`/refresh`](#post_matrixclientv1refresh). Clients can also refresh their
 access token at any time, even if it has not yet expired. If the token refresh
 succeeds, the client should use the new token for future requests, and can
 re-try previously-failed requests with the new token. When an access token is
--- a/data/api/client-server/refresh.yaml
+++ b/data/api/client-server/refresh.yaml
@@ -19,7 +19,7 @@ host: localhost:8008
 schemes:
   - https
   - http
-basePath: /_matrix/client/v3
+basePath: /_matrix/client/v1
 consumes:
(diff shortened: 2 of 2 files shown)


turt2live (changes requested):
As an r0 endpoint, it got elevated to v3 in MSC2844

The implementations would be incorrect.

KitsuneRal:
Does that effectively mean that refresh tokens do not exist in the wild despite being specced for half a year?

turt2live:
Seemingly so, but that wasn't really the requirement for implementation on that MSC.

clokep:
> As an r0 endpoint, it got elevated to v3 in [MSC2844](https://github.com/matrix-org/matrix-spec-proposals/issues/2844)

I don't think this was an `r0` endpoint. It looks like it landed at the same time as MSC2844 in Matrix 1.3.

(Note the MSC does use `r0` since it was written before MSC2844, MSC2918 merged after it though FWIW.)

KitsuneRal:
Wait, I actually don't see it being an r0 endpoint in scope for MSC2844.

turt2live:
All endpoints in the spec at the time of MSC2844 landing were in scope - that includes MSC2918 which was specified as an r0 endpoint and landed in the same spec version as MSC2844, ergo became `/v3` automatically. MSCs which landed *after* the v1.1 spec release (the first release with MSC2844 endpoints) but specified `/r0` *also* got elevated to `v3`, though I don't have specific examples on hand.

turt2live:
err, sorry: point of clarification on the refresh tokens bits: since it got merged as `r0`, it triggered the clauses of MSC2844 and got bumped to `/v3`.

turt2live:
for even more clarity on the timeline:

1. MSC2844 was accepted into the spec January 2021
2. MSC2918 was accepted into the spec October 2021
3. MSC2844 was formally specified a few days later in October 2021
4. Matrix 1.1 was released November 2021, codifying MSC2844
5. MSC2918 was formally specified in June 2022 (8 months later than MSC2844)
6. Matrix 1.3 was released June 2022, codifying MSC2918

With this timeline in mind, MSC2918 was in fact r0-stable at the time of Matrix 1.1 and thus subject to MSC2844's changes.

Issue #1275, opened 2022-10-06
E2EE clarifications in the aftermath of the 2022-09-28 disclosure


dkasak:
In the aftermath of https://matrix.org/blog/2022/09/28/upgrade-now-to-address-encryption-vulns-in-matrix-sdks-and-clients, we should make some spec clarifications so that it's harder for a new implementer to repeat the same mistakes. Listing some random ones I've thought of, feel free to edit more in.

- [ ] Specify explicitly that all encrypted to-device messages must be encrypted with Olm, not Megolm.

  - [ ] Specify mitigations against SAS attack. Namely that
    - An implementation should fix the list of keys it's attempting to verify at the beginning of the operation.
    - The implementation should from that point on refer to those keys solely by citing the public keys themselves, not by their "key IDs".
    - The implementation should first check whether a given key ID is a MSK (referring to a cross-signing user identity), and only then check whether it matches a device.

  - [ ] Specify that secrets should only be accepted from your own, trusted devices.
    - Note explicitly that when receiving a secret in an (Olm-encrypted) to-device message, the receiver must check that the device key of the other side of the Olm session ("sender key") matches an own, trusted device.

      It's *not* enough to check the device ID.

  - [ ] Specify what it means for a device to be trusted (i.e. it's when it's verified using cross-signing. Is it anything else?)

PR #1294, opened 2022-10-18
add some e2ee clarifications

uhoreg:
fixes https://github.com/matrix-org/matrix-spec/issues/1275

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -987,6 +987,17 @@ If a user's client sees that any other user has changed their master
 key, that client must notify the user about the change before allowing
 communication between the users to continue.
 
+Since device IDs and cross-signing keys occupy the same namespace, clients must
+ensure that they use the correct keys when verifying. While servers must not
+allow devices to have the same IDs as cross-signing keys, a malicious server
+could construct such a situation, so clients must not rely on the server being
+well-behaved and should take precautions against this. For example, clients
+should refer to keys using the public keys rather than only by the device
+ID. Clients should also fix the keys that are being verified, and ensure that
+they do not change in the course of verification. Clients may also display a
+warning and refuse to verify a user when it detects that the user has a device
+with the same ID as a cross-signing key.
+
 A user's user-signing and self-signing keys are intended to be easily
 replaceable if they are compromised by re-issuing a new key signed by
 the user's master key and possibly by re-verifying devices or users.
@@ -1587,7 +1598,11 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
-When a client is updating a Megolm session (room key) in its store, the client MUST ensure:
+When a client receives a Megolm session, the client MUST ensure that the
+session was received via a channel that ensures authenticity of the messages.
+Practically speaking, this means that Megolm sessions must be received via Olm.
+
+When a client is updating a Megolm session in its store, the client MUST ensure:
 
 * that the updated session data comes from a trusted source.
 * that the new session key has a lower message index than the existing session key.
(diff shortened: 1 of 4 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1587,7 +1598,11 @@ that they can decrypt future messages encrypted using this session. An
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
-When a client is updating a Megolm session (room key) in its store, the client MUST ensure:
+When a client receives a Megolm session, the client MUST ensure that the
+session was received via a channel that ensures authenticity of the messages.
```
dkasak, line 1611: Do we need to complicate here, rather than just demanding it's via Olm (for now)? My concern is that we don't explain what kind of authenticity properties we are demanding of the channel here.

data/event-schemas/schema/m.secret.request.yaml
```diff
@@ -0,0 +1,40 @@
...
+      requesting_device_id:
+        type: string
+        description: |-
+          The ID of the device requesting the secret.
+      request_id:
+        type: string
+        description: |-
+          A random string uniquely identifying (with respect to the requester
+          and the target) the target for a secret. If the secret is requested
+          from multiple devices at the same time, the same ID may be used for
```
dkasak, line 29:
```suggestion
          from multiple devices at the same time, the same ID MAY be used for
```

data/event-schemas/schema/m.secret.send.yaml
```diff
@@ -0,0 +1,32 @@
...
+description: |-
+  Sent by a client to share a secret with another device, in response to an
+  `m.secret.request` event. It must be encrypted as an `m.room.encrypted` event
+  using [Olm](#molmv1curve25519-aes-sha2), then sent as a to-device event.
+
+  The `request_id` must match the ID previously given in an `m.secret.request`
+  event. The recipient must ensure that this event comes from a device that the
+  `m.secret.request` event was originally sent to, and that the device is
+  trusted. This should be done by checking the sender key of the Olm session that
+  the event was sent over.
```
dkasak, line 13:
```suggestion
  The `request_id` must match the ID previously given in an `m.secret.request`
  event. The recipient must ensure that this event comes from a device that the
  `m.secret.request` event was originally sent to, and that the device is
  a trusted device owned by the recipient. This should be done by checking the
  sender key of the Olm session that the event was sent over.
```

We should only accept secrets from our own devices.

dkasak (changes requested):
What's there looks mostly good, but I have a few suggestions.

Also, in relation to https://github.com/matrix-org/matrix-spec/issues/1275, I don't think this PR addresses these points:

- The implementation should first check whether a given key ID is a MSK (referring to a cross-signing user identity), and only then check whether it matches a device.
- Specify what it means for a device to be trusted

Did I just miss it?

dbkr (commented):
(Not approving yet since Denis has outstanding comments)

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: In theory, a client could use some other type of channel that's custom to the client.  Then again, this PR already says that the `m.room_key` event "is encrypted as an `m.room.encrypted` event using Olm", so I guess I'll just change it to say that it must be sent over Olm.

uhoreg:
> * Specify what it means for a device to be trusted

Since we mean "verified" when we say "trusted", maybe we should just say "verified" instead?

dkasak (changes requested):
I left a final comment that needs addressing before merging, but otherwise this looks good to me.

The final comment concerns the goal of ending up not being vulnerable to key confusion bugs, which means we either have to flat out refuse to proceed in case of key ID collisions or we have to guarantee a consistent key checking order. I think I prefer flat out refusing, because it's simpler, unless you see a problem with it.

Once either suggestion is merged, I'm considering this good for merging.

dkasak (approved):
Yay, looks good!

dbkr (approved):

PR #1295, opened 2022-10-19
Update docsy (hugo theme) git submodule

anoadragon453:
This PR updates our git submodule checkout of [docsy](https://www.docsy.dev/) to https://github.com/matrix-org/docsy/commit/a0032f8db919a6c67ba6cdef2c455f105b6272a2, which is [the latest current commit](https://github.com/google/docsy/commit/90843e360b921ea8417995e0261a355f2f3aed7a) of google/docsy + [some changes to locally-load fonts/css/js](https://github.com/google/docsy/compare/main...matrix-org:docsy:master).

The previous version of docsy was >1 year old.

---

A few visual changes that I've noticed. These look fine and don't require fixing IMO:

**(Table cell vertical text alignment) Before**

image

**After**

image

---

**(Breadcrumbs) Before**

image

**After**

image

---

**(Rate-limit/Auth text) Before**

image

**After**

image

---

This change does not require a newsfragment.

Preview: https://pr1295--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -121,6 +121,12 @@ jobs:
         run: |
           npm i
           npm run get-proposals
+      - name: "⚙️ npm — docsy"
+        # Install dependencies according to https://github.com/google/docsy/#prerequisites
+        run: |
+          cd themes/docsy
+          npm i
+          npm i --save-dev autoprefixer postcss-cli postcss
       - name: "⚙️ hugo"
         run: hugo --baseURL "${{ needs.calculate-baseurl.outputs.baseURL }}" -d "spec"
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -121,12 +121,6 @@ jobs:
         run: |
           npm i
           npm run get-proposals
-      - name: "⚙️ npm — docsy"
-        # Install dependencies according to https://github.com/google/docsy/#prerequisites
-        run: |
-          cd themes/docsy
-          npm i
-          npm i --save-dev autoprefixer postcss-cli postcss
       - name: "⚙️ hugo"
         run: hugo --baseURL "${{ needs.calculate-baseurl.outputs.baseURL }}" -d "spec"
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -115,8 +115,6 @@ jobs:
           extended: true
       - name: "📥 Source checkout"
         uses: actions/checkout@v2
-        with:
-          submodules: 'recursive'
       - name: "⚙️ npm"
         run: |
           npm i
@@ -161,8 +159,6 @@ jobs:
           extended: true
       - name: "📥 Source checkout"
         uses: actions/checkout@v2
-        with:
-          submodules: 'recursive'
       - name: "⚙️ npm"
         run: |
           npm i
--- a/README.md
+++ b/README.md
@@ -65,15 +65,14 @@ place after an MSC has been accepted, not as part of a proposal itself.
    Alternatively, use the Docker image at
    https://hub.docker.com/r/klakegg/hugo/. (The "extended edition" is required
    to process the SCSS.)
-2. Run `git submodule update --init --recursive` for good measure.
-3. Run `npm i` to install the dependencies. Note that this will require NodeJS to be installed.
-4. Run `npm run get-proposals` to seed proposal data. This is merely for populating the content of the "Spec Change Proposals"
+2. Run `npm i` to install the dependencies. Note that this will require NodeJS to be installed.
+3. Run `npm run get-proposals` to seed proposal data. This is merely for populating the content of the "Spec Change Proposals"
    page and is not required.
-5. Run `hugo serve` (or `docker run --rm -it -v $(pwd):/src -p 1313:1313
+4. Run `hugo serve` (or `docker run --rm -it -v $(pwd):/src -p 1313:1313
    klakegg/hugo:ext serve`) to run a local webserver which builds whenever a file
    change is detected. If watching doesn't appear to be working for you, try
    adding `--disableFastRender` to the commandline.
-6. Edit the specification 🙂
+5. Edit the specification 🙂
 
 We use a highly customized [Docsy](https://www.docsy.dev/) theme for our generated site, which uses Bootstrap and Font
 Awesome. If you're looking at making design-related changes to the spec site, please coordinate with us in
--- a/README.md
+++ b/README.md
@@ -65,7 +65,8 @@ place after an MSC has been accepted, not as part of a proposal itself.
    Alternatively, use the Docker image at
    https://hub.docker.com/r/klakegg/hugo/. (The "extended edition" is required
    to process the SCSS.)
-2. Run `npm i` to install the dependencies. Note that this will require NodeJS to be installed.
+2. Run `npm i` to install the dependencies and fetch the docsy git submodule.
+   Note that this will require NodeJS to be installed.
 3. Run `npm run get-proposals` to seed proposal data. This is merely for populating the content of the "Spec Change Proposals"
    page and is not required.
 4. Run `hugo serve` (or `docker run --rm -it -v $(pwd):/src -p 1313:1313
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -36,19 +36,16 @@ $table-border-color: rgba(black, .125);
 $table-row-alternate: $secondary-lightest-background;
 $table-row-default: $secondary-lighter-background;
 
-/*
- Disable loading any fonts from Google Fonts.
- */
+/* Disable loading any fonts from Google Fonts. */
 $td-enable-google-fonts: false;
 
 /*
- Replace the default font with Inter - we load it from a local copy, which is downloaded from
- Google Fonts manually via a script:
- https://github.com/matrix-org/matrix-spec/tree/main/static/css/fonts
-
- The $font-family-sans-serif definition here overrides the default value set by docsy:
- https://github.com/matrix-org/docsy/blob/66a4e61d2d34edc7196b9df83a7d09cd4af14b47/assets/scss/_variables.scss#L68
- and adds "Inter" to the front.
- */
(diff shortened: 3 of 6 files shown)


turt2live (commented):
clearing review while draft

anoadragon453:
This has been updated to the latest https://github.com/matrix-org/docsy/ (as of today). https://github.com/matrix-org/matrix-spec/pull/1295/commits/343edc7a126a77a43ee7326e5da373fed4d339f0 prevents all google fonts from loading, which also allowed removing [one downstream change](https://github.com/google/docsy/commit/5023a2914528e012ecf3ec85a56028c00ee97dd2#diff-e5ddb28a5a4ed5b4709d39378bbed38ecc9ae9986b7c3001116fd0cff023d05dR66) in our fork of docsy.

.github/workflows/main.yml
```diff
@@ -121,6 +121,12 @@ jobs:
         run: |
           npm i
           npm run get-proposals
+      - name: "⚙️ npm — docsy"
+        # Install dependencies according to https://github.com/google/docsy/#prerequisites
+        run: |
+          cd themes/docsy
+          npm i
+          npm i --save-dev autoprefixer postcss-cli postcss
```
richvdh, line 129: why is this needed, when it wasn't before?

richvdh, line 129: >           npm i --save-dev autoprefixer postcss-cli postcss

this doesn't look right at all. This is modifying the `package.json` of docsy to say that these libraries are dev-time dependencies of docsy. If that's true, then it should be checked into git rather than done on the fly... but I don't think it's true.

The suggestion at https://github.com/google/docsy/#prerequisites is to do it in the Hugo site which uses docsy (ie, `matrix-spec`), not docsy itself. And it's already been done in matrix-spec: https://github.com/matrix-org/matrix-spec/blob/main/package.json#L21-L26

assets/scss/_variables_project.scss
```diff
@@ -37,8 +37,17 @@ $table-row-alternate: $secondary-lightest-background;
 $table-row-default: $secondary-lighter-background;
 
 /*
- Opt to serve fonts locally by overriding web-font-path to be a non-google fonts URL.
- This is only possible with our modified docsy theme: https://github.com/matrix-org/docsy
+ Disable loading any fonts from Google Fonts.
```
richvdh, line 40:
```suggestion
 * Configure docsy to use the default bootstrap fonts instead of Google Fonts.
 * See https://www.docsy.dev/docs/adding-content/lookandfeel/#fonts
```

assets/scss/_variables_project.scss
```diff
@@ -37,8 +37,17 @@ $table-row-alternate: $secondary-lightest-background;
...
+/*
+ Replace the default font with Inter - we load it from a local copy, which is downloaded from
+ Google Fonts manually via a script:
+ https://github.com/matrix-org/matrix-spec/tree/main/static/css/fonts
+
+ The list of fonts (with Inter prepended) is taken from docsy's source code:
+ https://github.com/matrix-org/docsy/blob/66a4e61d2d34edc7196b9df83a7d09cd4af14b47/assets/scss/_variables.scss#L68
+ */
+@import url("../css/fonts/Inter.css");
+$font-family-sans-serif: "Inter", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
```
richvdh, line 53: is this `$font-family-sans-serif` ours? or bootstrap's? Or docsy's? Or something else?

assets/scss/custom.scss
```diff
@@ -340,6 +340,7 @@ footer {
   }
 
   table {
+    display: table;
```
richvdh, line 343: somewhat surprised this is necessary.

richvdh (changes requested):

richvdh:
Also: changelog please

assets/scss/_variables_project.scss
anoadragon453: Note that no fonts are imported by bootstrap - bootstrap tries to load fonts from the system: https://getbootstrap.com/docs/5.2/content/reboot/#native-font-stack

`$td-enable-google-fonts` will disable fetching fonts from Google Fonts, and use system fonts instead. The list of system fonts is configured by `$font-family-sans-serif` below. The comment about "The list of fonts ..." tried to explain this, but I'll clarify it to mention `$font-family-sans-serif` explicitly. While the contents of `$font-family-sans-serif` looks like what bootstrap defines, it's originally from docsy's source code.

We have to override it (with Inter at the front) in order to load Inter.

assets/scss/custom.scss
anoadragon453: I've traced this back to https://github.com/twbs/bootstrap/blob/e5643aaa89eb67327a5b4abe7db976f0ea276b70/scss/_tables.scss#L173 causing tables to have `display: block`. Docsy makes all tables responsive: https://github.com/google/docsy/blob/d9bf98b876d74772b850ef910c6f59b898f61ccf/assets/scss/_content.scss#L41

Bootstrap 5.0.0+ has removed the `display: block` bit: https://github.com/twbs/bootstrap/pull/30482. However, Docsy is currently stuck on Bootstrap 4.6.2: https://github.com/google/docsy/issues/470.

https://github.com/google/docsy/commit/1e7e32934f520cfe1baaecfe6576db42addc89d1 did indeed cause this, as when adding a `>` back in front of `table` again, the `display: table` is no longer needed. I'm not sure if that's unintended behaviour for the PR, so I've left a comment.

Including `display: table` here seems to be an ample workaround for right now though. I'll leave a comment mentioning why it's needed.

assets/scss/_variables_project.scss
anoadragon453: Hopefully answered above.

.github/workflows/main.yml
anoadragon453: Done in 3313b6a0a5a8fafae0479b8513859e0201d62ff0.

assets/scss/_variables_project.scss
anoadragon453: Done in a101159d96f684ae9b6563c4fb479ce81797b60e.

anoadragon453 (commented):
> Also: changelog please

[CONTRIBUTING.md mentions](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog) that:

> Changes that do not change the spec, such as changes to the build script, formatting, CSS, etc should not get a news fragment.

So I'm not sure if this PR needs one?

package.json
```diff
@@ -5,6 +5,9 @@
   "main": "none.js",
   "scripts": {
     "get-proposals": "node ./scripts/proposals.js",
+    "get:submodule": "git submodule update --init --depth 1",
+    "_prepare:docsy": "cd themes/docsy && npm install",
+    "prepare": "npm run get:submodule && npm run _prepare:docsy",
```
anoadragon453, line 10: This is taken from docsy's docs: https://www.docsy.dev/docs/get-started/other-options/#for-an-existing-site

Unfortunately, we can't add comments to package.json files...

richvdh:
> > Also: changelog please
> 
> [CONTRIBUTING.md mentions](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst?rgh-link-date=2022-11-01T18%3A37%3A24Z#adding-to-the-changelog) that:
> 
> > Changes that do not change the spec, such as changes to the build script, formatting, CSS, etc should not get a news fragment.
> 
> So I'm not sure if this PR needs one?

CONTRIBUTING.md lies. We have an `internal` changelog section now for this sort of thing.

README.md
```diff
@@ -65,15 +65,14 @@ place after an MSC has been accepted, not as part of a proposal itself.
    Alternatively, use the Docker image at
    https://hub.docker.com/r/klakegg/hugo/. (The "extended edition" is required
    to process the SCSS.)
-2. Run `git submodule update --init --recursive` for good measure.
```
richvdh, line 68: do we not still need a `git submodule update --init` here, even if `--recursive` is no longer necessary? docsy is still a submodule which needs cloning/

assets/scss/_variables_project.scss
richvdh: Ok, the thing I really wanted to see here is a link to the place where `td-enable-google-fonts` is defined, which we still don't have. The interesting point here is that it's exposed by *docsy* rather than bootstrap.

I think my original suggestion is ok, with s/bootstrap/system/ ?

assets/scss/custom.scss
richvdh: sounds good. I don't see a comment yet though?

richvdh (changes requested):
thanks for doing all that digging. still a couple of nits.

assets/scss/_variables_project.scss
anoadragon453: Ah, yes we definitely should add a link to that piece of documentation.

> I think my original suggestion is ok, with s/bootstrap/system/ ?

Yes, this looks good on a re-read.

anoadragon453: Added the modified suggestion in 4473805c2743544912076162098a5e31666c8da0.

assets/scss/custom.scss
anoadragon453: Oops. High aspirations. Have left both comments now, the one here in f261919feebad7e591ed08770c354ef2a1f3305c.

anoadragon453 (commented):
I've added a changelog entry as well (as `.feature`, which I thought fit best amongst [the options](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog), but feedback welcome!).

richvdh (approved):
:shipit:

anoadragon453:
image

assets/scss/custom.scss
chalin: Great in-depth analysis and explanation. Thanks for letting us know on the Docsy side as well. As you write, hopefully this will be resolved once Docsy upgrades to Bootstrap 5.

PR #1303, opened 2022-10-25
Improve error messages emitted by `resolve-additional-types`

richvdh:
I forgot to set the `items` on an array definition, and got an extremely
opaque error. Hopefully this will improve the lives of anyone who makes a
similar mistake in future.

Preview: https://pr1303--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/json-schema/resolve-additional-types.html
+++ b/layouts/partials/json-schema/resolve-additional-types.html
@@ -3,7 +3,8 @@
   Finds and returns all nested objects, given a dict containing:
    * `schema`: a JSON schema object
    * `anchor_base`: a prefix to add to the HTML anchors generated for each object. If nil, no anchors are generated.
-   * `name`: optionally, a name to use for this object in error/warning messages.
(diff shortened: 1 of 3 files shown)


richvdh:
I don't think this needs a changelog?

layouts/partials/json-schema/resolve-additional-types.html
```diff
@@ -3,6 +3,7 @@
   Finds and returns all nested objects, given a dict containing:
    * `schema`: a JSON schema object
    * `anchor_base`: a prefix to add to the HTML anchors generated for each object. If nil, no anchors are generated.
+   * `name`: optionally, a name to use for this object in error/warning messages.
```
anoadragon453, line 6: You may wish to mention that if `name` is not set, the object's `title` property (if present) is used instead.

richvdh:
```suggestion
   * `name`: optionally, a name to use for this object in error/warning messages. If left unset,
      the object's `title` property is used (if present).
```

PR #1305, opened 2022-10-25
Clarifications to AS spec, including MSC3905

richvdh:
Primarily this is the spec for MSC3905, but I've also taken the opportunity to clean up the section a bit and move the definition out to a `.yaml` file.

Suggest reviewing commit-by-commit.

Preview: https://pr1305--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/application_service/newsfragments/1305.clarification
+++ b/changelogs/application_service/newsfragments/1305.clarification
@@ -1,0 +1,1 @@
+Clarify that application services can only register an interest in local users.
--- a/changelogs/application_service/newsfragments/1305.clarification
+++ b/changelogs/application_service/newsfragments/1305.clarification
@@ -1,1 +1,1 @@
-Clarify that application services can only register an interest in local users.
+Clarify that application services can only register an interest in local users, as per [MSC3905](https://github.com/matrix-org/matrix-spec-proposals/issues/3905).
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -81,13 +81,15 @@ namespaces:
 ```
 
 {{% boxes/note %}}
-Application services can only register interest in *local* users (i.e., users
-whose IDs end with the `server_name` of the local homeserver). Events affecting
-users on other homeservers are not sent to an application service, even if the
-user happens to match the one of the `users` namespaces (unless, of course, the
-event affects a room that the application service is interested in for another
-room - for example, because there is another user in the room that the
+For the `users` namespace, application services can only register interest in
+*local* users (i.e., users whose IDs end with the `server_name` of the local
+homeserver). Events affecting users on other homeservers are not sent to an application
+service, even if the user happens to match the one of the `users` namespaces (unless,
+of course, the event affects a room that the application service is interested in
+for another room - for example, because there is another user in the room that the
 application service is interested in).
+
+For the `rooms` and `aliases` namespaces, all events in a matching room will be sent to the application service.
 {{% /boxes/note %}}
 
 {{% boxes/warning %}}
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -37,14 +37,8 @@ registration for suspicious regex strings.
 {{% /boxes/note %}}
 
 Application services register "namespaces" of user IDs, room aliases and
-room IDs. These namespaces are represented as regular expressions. An
-application service is said to be "interested" in a given event if one
-of the IDs in the event match the regular expression provided by the
-application service, such as the room having an alias or ID in the
-relevant namespaces. Similarly, the application service is said to be
-interested in a given event if one of the application service's
-namespaced users is the target of the event, or is a joined member of
-the room where the event occurred.
+room IDs. An application service is said to be "interested" in a given event
+if it matches any of the namespaces.
 
 An application service can also state whether they should be the only
 ones who can manage a specified namespace. This is referred to as an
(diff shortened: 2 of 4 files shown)


changelogs/application_service/newsfragments/1305.clarification
```diff
@@ -0,0 +1 @@
+Clarify that application services can only register an interest in local users.
```
turt2live, line 1:
```suggestion
Clarify that application services can only register an interest in local users, as per [MSC3905](https://github.com/matrix-org/matrix-spec-proposals/issues/3905).
```

The changelog needs to reference the MSC, always.

data/api/application-service/definitions/registration.yaml
```diff
@@ -0,0 +1,74 @@
...
+              service is interested in. Events will be sent to the AS if a
+              local user matching one of the namespaces is the target of the event,
+              or is a joined member of the room where the event occurred.
+      rooms:
+        allOf:
+          - $ref: namespace_list.yaml
+          - description: |-
+              A list of namespaces defining the room IDs that the application
+              service is interested in. Events will be sent to the AS if they
+              take place in a room with an ID matching one of the namespaces.
```
MadLittleMods, line 52:
```suggestion
              A list of namespaces defining the room IDs that the application
              service is interested in. All events sent in a matching room ID
              will be sent to the AS.
```

data/api/application-service/definitions/registration.yaml
```diff
@@ -0,0 +1,74 @@
...
+      aliases:
+        allOf:
+          - $ref: namespace_list.yaml
+          - description: |-
+              A list of namespaces defining the room aliases that the application
+              service is interested in. Events will be sent to the AS if they
+              take place in a room with an alias matching one of the namespaces.
```
MadLittleMods, line 59:
```suggestion
              A list of namespaces defining the room aliases that the application
              service is interested in. All events sent in a matching room alias
              will be sent to the AS.
```

content/application-service-api.md
```diff
@@ -83,38 +61,34 @@ they represent in the reserved namespace. For example, `@_irc_.*` would
...
+```
+
+{{% boxes/note %}}
+Application services can only register interest in *local* users (i.e., users
+whose IDs end with the `server_name` of the local homeserver). Events affecting
+users on other homeservers are not sent to an application service, even if the
+user happens to match the one of the `users` namespaces (unless, of course, the
+event affects a room that the application service is interested in for another
+room - for example, because there is another user in the room that the
+application service is interested in).
```
MadLittleMods, line 90: I think the previous text is too confusing in regards to `rooms`/`aliases`

```suggestion
For the `users` namespace, application services can only register interest in
*local* users (i.e., users whose IDs end with the `server_name` of the local
homeserver). Events affecting users on other homeservers are not sent to an application
service, even if the user happens to match the one of the `users` namespaces (unless,
of course, the event affects a room that the application service is interested in
for another room - for example, because there is another user in the room that the
application service is interested in).

For the `rooms` and `aliases` namespaces, all events in a matching room will be sent to the application service.
```

content/application-service-api.md
```diff
@@ -53,28 +47,12 @@ application services from creating/deleting entities in that namespace.
 Typically, exclusive namespaces are used when the rooms represent real
 rooms on another service (e.g. IRC). Non-exclusive namespaces are used
 when the application service is merely augmenting the room itself (e.g.
-providing logging or searching facilities). Namespaces are represented
-by POSIX extended regular expressions and look like:
```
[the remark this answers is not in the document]
richvdh: yes it matters, but it's included as part of the definition of the Namespace object.

content/application-service-api.md
richvdh: yes fair, thank you.

data/api/application-service/definitions/registration.yaml
richvdh:
```suggestion
              A list of namespaces defining the room IDs that the application
              service is interested in. All events sent in a room with an ID
              which matches one of the namespaces will be sent to the AS.
```

content/application-service-api.md
```diff
@@ -37,14 +37,8 @@ registration for suspicious regex strings.
...
-room IDs. These namespaces are represented as regular expressions. An
-application service is said to be "interested" in a given event if one
-of the IDs in the event match the regular expression provided by the
-application service, such as the room having an alias or ID in the
-relevant namespaces. Similarly, the application service is said to be
-interested in a given event if one of the application service's
-namespaced users is the target of the event, or is a joined member of
-the room where the event occurred.
+room IDs. An application service is said to be "interested" in a given event
+if it matches any of the namespaces.
```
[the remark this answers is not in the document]
richvdh: Not entirely convinced this is an improvement in clarity, personally. The fact that namespaces are defined via regular expressions is a detail which I think is better left for the formal definition below rather in this overview.

(In any case, we'd want to say "namespace regular expressions" rather than "namespace regex patterns").

content/application-service-api.md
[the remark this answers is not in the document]
MadLittleMods: (feel free to resolve)

content/application-service-api.md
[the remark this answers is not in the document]
MadLittleMods: "namespace regular expressions" sounds good to me 👍

I think it makes sense to mention when we're talking about matching. It's immediately apparent and understandable for anyone slightly familiar with regex how it matches then.

dbkr (approved):
Looks good as far as I can see

PR #1306, opened 2022-10-26
Replace misuses of 'plaintext' with 'cleartext'

### LINKED ISSUE #1223: Spec refers to "plaintext" when it means "cleartext"
**Link to problem area**:

https://spec.matrix.org/unstable/client-server-api/#forming-relationships-between-events

**Issue**

>  the `m.relates_to` key of an event MUST be included in the plaintext copy of the event

"plaintext" is simply wrong here: plaintext is stuff that was/will be encrypted; cleartext is stuff that is never encrypted; cf https://www.hypr.com/cleartext/. The `m.relates_to` property is in the part of the event that is never encrypted.

(while we're there: our [style guide](https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst) says 'avoid the term "key" unless you are specifically talking about the *name* of a property'. The text above is referring to the *property*, not the key.)

anoadragon453:
Closes #1223.

Also replaces the term in the documentation of spoilers, where 'spoilered text' makes more sense in my opinion.

### DIFF
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -327,11 +327,12 @@ If a reason were to be supplied, it would look like:
 }
 ```
 
-When sending a spoiler, clients SHOULD provide the plain text fallback in the `body`
+When sending a spoiler, clients SHOULD provide the spoilered text fallback in the `body`
 as shown above (including the reason). The fallback SHOULD omit the spoiler text verbatim
 since `body` might show up in text-only clients or in notifications. To prevent spoilers
-showing up in such situations, clients are strongly encouraged to first upload the plaintext
-to the media repository then reference the MXC URI in a markdown-style link, as shown above.
+showing up in such situations, clients are strongly encouraged to first upload the spoilered
+text to the media repository then reference the MXC URI in a markdown-style link, as shown
+above.
 
 Clients SHOULD render spoilers differently with some sort of disclosure. For example, the
 client could blur the actual text and ask the user to click on it for it to be revealed.
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -1945,14 +1945,14 @@ parent event, for example.
 {{% /boxes/note %}}
 
 To allow the server to aggregate and find child events for a parent, the `m.relates_to`
-key of an event MUST be included in the plaintext copy of the event. It cannot be
+key of an event MUST be included in the cleartext portion of the event. It cannot be
 exclusively recorded in the encrypted payload as the server cannot decrypt the event
 for processing.
 
 {{% boxes/warning %}}
 If an encrypted event contains an `m.relates_to` in its payload, it should be
-ignored and instead favour the plaintext `m.relates_to` copy (including when there
-is no plaintext copy). This is to ensure the client's behaviour matches the server's
+ignored and instead favour the cleartext `m.relates_to` copy (including when there
+is no cleartext copy). This is to ensure the client's behaviour matches the server's
 capability to handle relationships.
 {{% /boxes/warning %}}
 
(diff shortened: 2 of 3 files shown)


content/client-server-api/modules/instant_messaging.md
```diff
@@ -327,11 +327,12 @@ If a reason were to be supplied, it would look like:
 }
 ```
 
-When sending a spoiler, clients SHOULD provide the plain text fallback in the `body`
+When sending a spoiler, clients SHOULD provide the spoilered text fallback in the `body`
```
richvdh, line 330: The term "spoilered" is ambiguous for me. Does it mean the un-obfuscated text (ie, containing spoilers), or does it mean text which has been though an anti-spoilering process?

anoadragon453: The text containing spoilers, but good point. Replacing simply with "the text containing spoilers" is much clearer.

content/client-server-api/modules/instant_messaging.md
```diff
@@ -327,11 +327,11 @@ If a reason were to be supplied, it would look like:
...
-When sending a spoiler, clients SHOULD provide the plain text fallback in the `body`
-as shown above (including the reason). The fallback SHOULD omit the spoiler text verbatim
-since `body` might show up in text-only clients or in notifications. To prevent spoilers
-showing up in such situations, clients are strongly encouraged to first upload the plaintext
-to the media repository then reference the MXC URI in a markdown-style link, as shown above.
+When sending a spoiler, clients SHOULD provide the fallback in the `body` as shown above
+(including the reason). The fallback SHOULD omit the spoiler text verbatim since `body`
+might show up in text-only clients or in notifications. To prevent spoilers showing up in
+such situations, clients are strongly encouraged to first upload the text containing spoilers
+to the media repository, then reference the MXC URI in a markdown-style link, as shown above.
```
anoadragon453, line 334: Apologies for the poor diff due to formatting. The real diff is:

> When sending a spoiler, clients SHOULD provide **the fallback** in the `body` as shown above
> (including the reason). The fallback SHOULD omit the spoiler text verbatim since `body`
> might show up in text-only clients or in notifications. To prevent spoilers showing up in
> such situations, clients are strongly encouraged to first upload **the text containing spoilers**
> to the media repository, then reference the MXC URI in a markdown-style link, as shown above.

richvdh: I don't know what "omit the spoiler text verbatim" means, but that's not new.

This looks like a net clarification to me.

changelogs/client_server/newsfragments/1306.clarification
```diff
@@ -0,0 +1 @@
+Replace various instances of the word "plaintext" with "cleartext".
```
richvdh, line 1: I'd go with a more generic

```suggestion
Various clarifications throughout the specification.
```

richvdh (approved):

changelogs/client_server/newsfragments/1306.clarification
anoadragon453: I like that better than the generic "Fix various typos throughout the specification." we've been using elsewhere.

content/client-server-api/modules/instant_messaging.md
anoadragon453: Perhaps we can drop the "verbatim". In any case, it just means: don't include the text including spoilers in `body`.

anoadragon453: I've tried to clean this sentence up a bit in https://github.com/matrix-org/matrix-spec/pull/1306/commits/c24fb1d0e0bae15d40a6411230351809425eb4f8.

PR #1310, opened 2022-10-28
Upgrade Swagger data to OpenAPI 3.1

### LINKED ISSUE #331: Upgrade OpenAPI

zecakeh:
This is still a rough draft.

#### What was done to get here:

1. Ran [swagger2openapi](https://github.com/Mermade/oas-kit/tree/main/packages/swagger2openapi) to upgrade to OpenAPI 3.0.
2. Made a few edits manually to follow the [migration guide] between OpenAPI 3.0 and 3.1.
3. Updated and ran `scripts/validator.js` on the folders in `data/api` and fixed any issue that was reported.
4. The converter merged the server URLs into an array of `servers`. I edited them to split the URLs into variables so we can access the `basePath` easily for the other tools.
5. Built the spec before the upgrade, and then again after the upgrade, and fixed the differences between the HTML files. Most of those needed either:
    - Editing the templates to make them compatible with the new format,
    - Restoring data that was removed by the converter (some examples and fields at the same level as a `$ref`).
6. Updated `scripts/dump-swagger.py` to adapt to the changes in the data and generate valid OpenAPI 3.1 data.
7. Added `scripts/swagger-preview.html` to allow to preview the generated API with RapiDoc, and to make sure that it works. Also updated the README to replace instructions for previewing with Swagger UI (which is not compatible with OpenAPI 3.1).

#### What's left to do:
- [ ] The converter added a lot of unnecessary newlines in the descriptions, they should probably be cleaned up.
- [ ] Most of the JSON examples were converted to YAML. Maybe convert all the examples to YAML? Or revert this to make the diff smaller?
- [ ] 4 files do not pass validation in `data/api/server-server`:
    - `keys_query.yaml` and `keys_server.yaml`: the `keyId` property in the path cannot be `required: false`, in the OpenAPI 3.1 spec, all properties in path must be `required: true`.
    - `invites-v1.yaml` and `invites-v2.yaml`: the validator complains that it cannot resolve the reference in the example of the 200 response. It's not true because the same reference is used at another place in the file without any issue, so there must be something wrong with the syntax around the `$ref`.
- [ ] Check every single file that no data was lost (including comments)

I'll open discussions on the code for feedback for the remaining tasks.

Closes #331.


data/api/application-service/definitions/location.yaml
```diff
@@ -15,18 +15,21 @@ properties:
   alias:
     description: An alias for a matrix room.
     type: string
-    example: "#freenode_#matrix:matrix.org"
+    examples:
+      - "#freenode_#matrix:matrix.org"
```
zecakeh, line 19: Switching from `example` to `examples` in schemas was not strictly necessary, because the field is only deprecated in `OpenAPI 3.1`, but I thought it made sense to already switch to the recommended format.

data/api/application-service/query_room.yaml
```diff
@@ -12,79 +12,99 @@
...
+        This endpoint is invoked by the homeserver on an application service to
+        query
+
         the existence of a given room alias. The homeserver will only query room
+
         aliases inside the application service's `aliases` namespace. The
+
         homeserver will send this request when it receives a request to join a
+
         room alias within the application service's namespace.
```
zecakeh, line 33: The converter switched to the "folded" style which means newlines are only added to text when there's an empty line, which is similar to Markdown. However it expected the description to be in "literal" style before conversion, which means all newlines are kept, so it added empty lines to keep all the newlines from before, which is wrong, imo.

It should be noted that this doesn't have any effect on the rendering of the spec or in RapiDoc.

data/api/client-server/admin.yaml
```diff
@@ -41,75 +30,82 @@ paths:
...
+                                      type: string
+                                      description: Most recently seen IP address of the session.
+                                    last_seen:
+                                      type: integer
+                                      format: int64
+                                      description: Unix timestamp that the session was last active.
+                                    user_agent:
+                                      type: string
+                                      description: User agent string last seen in the session.
+              examples:
```
zecakeh, line 82: Most of JSON examples were converted to YAML during conversion. Should we do that for all examples? Revert this to make the diff smaller?

data/api/server-server/invites-v1.yaml
```diff
@@ -11,157 +11,186 @@
...
+                        "event":  {
+                          "$ref": "examples/minimal_pdu.json",
+                          "type": "m.room.member",
+                          "state_key": "@someone:example.org",
+                          "origin": "example.org",
+                          "origin_server_ts": 1549041175876,
+                          "sender": "@someone:example.org",
+                          "unsigned": {
+                            "invite_room_state": {
+                              "$ref": "../../event-schemas/examples/invite_room_state.json"
```
zecakeh, line 152: The validator complains about not being able to load this reference, although it is able to resolve the exact same reference in the `requestBody`, so there must be something wrong with the syntax.

data/api/server-server/invites-v2.yaml
```diff
@@ -11,176 +11,189 @@
...
+                      "event":  {
+                        "$ref": "examples/minimal_pdu.json",
+                        "type": "m.room.member",
+                        "state_key": "@someone:example.org",
+                        "origin": "example.org",
+                        "origin_server_ts": 1549041175876,
+                        "sender": "@someone:example.org",
+                        "unsigned": {
+                          "invite_room_state": {
+                            "$ref": "../../event-schemas/examples/invite_room_state.json"
```
zecakeh, line 130: Same issue as in `invites-v1.yaml`

data/api/server-server/keys_query.yaml
```diff
@@ -35,99 +26,131 @@ paths:
...
         - in: path
           name: keyId
-          type: string
           description: |-
             **Deprecated**. Servers should not use this parameter and instead
             opt to return all keys, not just the requested one. The key ID to
             look up.
 
             When excluded, the trailing slash on this endpoint is optional.
           required: false
```
zecakeh, line 42: This is invalid in OpenAPI 3.x, all path parameters must have `required:true`.

Should we define two separate paths and deprecate the one with the `keyId`?

data/api/server-server/keys_server.yaml
```diff
@@ -46,18 +39,26 @@ paths:
...
         - in: path
           name: keyId
-          type: string
           description: |-
             **Deprecated**. Servers should not use this parameter and instead
             opt to return all keys, not just the requested one. The key ID to
             look up.
 
             When excluded, the trailing slash on this endpoint is optional.
           required: false
```
zecakeh, line 48: Same issue as in `keys_query.yaml`

data/api/client-server/appservice_room_directory.yaml
```diff
@@ -11,78 +11,91 @@
...
-  - https
-  - http
-basePath: /_matrix/client/v3
-consumes:
-  - application/json
-produces:
-  - application/json
-securityDefinitions:
-  # Note: this is the same access_token definition used elsewhere in the client
-  # server API, however this expects an access token for an application service.
```
zecakeh, line 29: Comments like that were lost by the converter, and maybe other data. Every single file needs to be checked for that.

richvdh:
I'm going to mark this as "ready for review" to gather some feedback. It's likely to get overlooked while it's in draft.

data/api/application-service/definitions/location.yaml
richvdh: Not a hard objection, but anything we can do to reduce the size of this diff to make it easier to review would be a good thing. If it's not a huge amount of work, I would defer this for a future PR.

data/api/application-service/query_room.yaml
richvdh: yeah, we should clean these up.

I think the problem is that the definitions were using literal style to start with. If we went through and replaced `|-` with `>-` before doing the conversion, would that solve this problem?

data/api/server-server/invites-v1.yaml
richvdh: hrm, odd. Let's park this for now and come back to it.

data/api/server-server/keys_query.yaml
richvdh: To be honest we could just drop `keyId` from the spec. It's been deprecated for years, and nobody should be relying on it.

richvdh:
@zecakeh thank you so much for picking this up - there is a lot of great work here and I am very excited to see it land!

I've made a few comments in reply to your questions.

Generally, I'd love it if we could find ways to pull out bits of this PR to reduce the size of it. For example:

* > Added scripts/swagger-preview.html to allow to preview the generated API with RapiDoc, and to make sure that it works. Also updated the README to replace instructions for previewing with Swagger UI (which is not compatible with OpenAPI 3.1).

  Given RapiDoc is compatible with both OpenAPI 2 and OpenAPI 3, let's do this now as a separate PR, before we change anything else?

* If there are changes that we can make to the existing definitions (such as replacing `|-` with `>-` for the descriptions) which will make them easier to convert, let's do that?

Another question: How much manual editing did you do? Should we aim to script the changes? It might be easier to review a set of conversion scripts than 167 manually-edited files!

data/api/application-service/query_room.yaml
zecakeh: Sure, I'll test that

data/api/client-server/admin.yaml
zecakeh: As far as I understand, the script parses everything, edits it and then serializes it according to the file extension, that's why comments are gone and everything is converted to YAML.

One thing we could do is to see if it's possible to copy examples and comments and link them to a field, and we reintroduce them in the proper place after the conversion.

zecakeh:
Thanks for the feedback.
    
>       Given RapiDoc is compatible with both OpenAPI 2 and OpenAPI 3, let's do this now as a separate PR, before we change anything else?

Done in #1331.

>     * If there are changes that we can make to the existing definitions (such as replacing `|-` with `>-` for the descriptions) which will make them easier to convert, let's do that?

Sure, I'll look into it.

> Another question: How much manual editing did you do? Should we aim to script the changes? It might be easier to review a set of conversion scripts than 167 manually-edited files!

I think I did more manual editing than necessary, mostly because I didn't know what I was getting into, and what a valid result would look like. I'll look into writing scripts to automate most (or all) of it.

data/api/client-server/admin.yaml
richvdh: > As far as I understand, the script parses everything, edits it and then serializes it according to the file extension, that's why comments are gone and everything is converted to YAML.

Makes sense. I don't think it's worth spending a lot of effort to work around this.

data/api/application-service/query_room.yaml
zecakeh: Actually we have to keep the literal style, otherwise we lose new paragraphs with Hugo's markdownify (or we need to add an extra newline to get the new paragraph).

However, it should be possible to control the output serialization to make it use the same format.

zecakeh:
I split the different steps needed for conversion, between automatic and manual steps. I also removed the changes that are not strictly necessary.

With these commits:
- The definitions pass the `validator.js` script.
- The generated spec is unchanged except for the same 4 definitions files that wouldn't pass validation otherwise.

I found out another small thing that could reduce the diff if done in a separate PR: switching the definitions and tools to use [`x-servers`](https://github.com/mermade/swaggerplusplus#features). Then during conversion it would just be a `x-servers` `servers` change.

zecakeh:
I tried with another validator (namely [Redocly CLI](https://github.com/Redocly/redocly-cli)) and it complained about a few things.

I trust these changes because most of them are required by swagger-parser when declaring the files as OpenAPI 3.0.0, but it stops complaining for some reason when the version is OpenAPI 3.1.0. They also make sense.

Redocly CLI also complains about `$ref` in examples which doesn't seem to be actually [supported by the OpenAPI spec](https://spec.openapis.org/oas/v3.1.0#example-object): it expects either the whole example as a `$ref`, or the complete content of the example.

data/api/server-server/invites-v1.yaml
zecakeh: The reason why it doesn't work might be because `$ref` are not supposed to be supported in examples, however that doesn't explain why it doesn't complain about other references.

scripts/package.json
```diff
@@ -9,7 +9,7 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "nopt": "^3.0.2",
-    "swagger-parser": "^10.0.3"
+    "@apidevtools/swagger-parser": "^10.1.0",
```
richvdh, line 12: this thing seems to work on swagger 2 as well? let's land it separately if so.

richvdh:
> Redocly CLI also complains about $ref in examples which doesn't seem to be actually [supported by the OpenAPI spec](https://spec.openapis.org/oas/v3.1.0#example-object): it expects either the whole example as a $ref, or the complete content of the example.

Ok. As I understand it, there are only a couple of places where this is a problem? If so, let's fix it by just copying the complete content into the right place instead of using `$ref`. Could you open a separate PR?

Likewise, would you be able to open a separate PR for getting rid of `{keyId}` on the `/keys` endpoints?

richvdh (changes requested):

scripts/package.json
zecakeh: It complains in the Identity Service API about the `3xx` and `4xx` responses status codes. I'm not sure what other syntax could be used for the same thing that would be considered valid swagger2.

zecakeh:
> > Redocly CLI also complains about $ref in examples which doesn't seem to be actually [supported by the OpenAPI spec](https://spec.openapis.org/oas/v3.1.0#example-object): it expects either the whole example as a $ref, or the complete content of the example.
> 
> Ok. As I understand it, there are only a couple of places where this is a problem? If so, let's fix it by just copying the complete content into the right place instead of using `$ref`. Could you open a separate PR?

Done in #1349.

> Likewise, would you be able to open a separate PR for getting rid of `{keyId}` on the `/keys` endpoints?

Done in #1350.

data/api/server-server/user_keys.yaml
```diff
@@ -12,243 +12,221 @@
...
-                              type: object
-                              additionalProperties:
+                      See the [Client-Server Key Algorithms](/client-server-api/#key-algorithms) section for more information on
+                      the Key Object format.
+                    additionalProperties:
+                      type: object
+                      additionalProperties:
+                        oneOf:
+                          - type: string
+                          - type: object
```
zecakeh, line 71: I don't think it's the proper fix that should be applied. See #1351.

richvdh:
@zecakeh it's gone a bit quiet here, so I just wanted to say I'm still very excited about seeing this land. I'm trying to push through MSC3938 so we can fix the `/keys` endpoints, and I believe @KitsuneRal is working his way through a more comprehensive review on this. Please bear with us!

scripts/package.json
KitsuneRal: Nxx notation seems to [only be available in OpenAPI 3](https://swagger.io/docs/specification/describing-responses/#status-codes) - cf. the "HTTP Status Codes" section in the [guide for Swagger 2](https://swagger.io/docs/specification/2-0/describing-responses/).

KitsuneRal: ...however, we do (or at least did) use this notation as our custom extension, as documented in [openapi_extensions.md](https://github.com/matrix-org/matrix-spec/blob/main/openapi_extensions.md)

data/api/server-server/user_keys.yaml
KitsuneRal: As I also commented there - it's almost proper, except that it should be one level of `additionalProperties` deeper.

scripts/package.json
zecakeh: If it's a custom extension, shouldn't it start with `x-`? Maybe the validator won't complain then.

KitsuneRal (commented):
Starting slow: please see some comments from reviewing `application-service/`, meanwhile I'm in for CS API (but that will likely be my Christmas reading).

data/api/application-service/protocols.yaml
```diff
@@ -40,42 +30,50 @@ paths:
...
-            $ref: ../client-server/definitions/errors/error.yaml
-        404:
+          content:
+            application/json:
+              schema:
+                $ref: ../client-server/definitions/errors/error.yaml
+              examples:
+                response:
+                  value:
+                    errcode: COM.EXAMPLE.MYAPPSERVICE_UNAUTHORIZED
```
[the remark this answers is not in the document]
zecakeh: This is the output of the conversion. The current definitions are completely parsed by the converter, then edited, then serialized again in the format of the file. So sadly there is not much to do.

data/api/application-service/protocols.yaml
```diff
@@ -89,102 +87,120 @@ paths:
...
           name: protocol
-          type: string
           description: The protocol ID.
           required: true
-          x-example: irc
+          example: irc
+          schema:
+            type: string
         - in: query
           name: fields...
```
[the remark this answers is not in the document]
zecakeh: The [OpenAPI 3.1 spec](https://spec.openapis.org/oas/v3.1.0#fixed-fields-9) says that `explode` defaults to `true` only if `style` is `form`, which is the default for `query`.

The example sets `style` to `matrix`, which is why it must explicitely set `explode` to `true`.

However I see another problem here. The type of `fields` shouldn't be `string`, it should probably be `object`.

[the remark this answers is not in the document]
KitsuneRal: The piece with `style: matrix` is the one for `path`; whereas the one for `query` (with `schema: { type: boolean }`) has a comment that spells out its defaults and it says `explode=false`. And no, I don't think that `fields` should be an `object`, as I look at the same example.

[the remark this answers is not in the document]
KitsuneRal: Hm, ok, after all the sources combined (I have just looked at the example in the spec itself), it seems the example in the guide is really misleading if not incorrect, and your conclusions are accurate: the schema type for `fields` should be `object`.

Taking it formally we should also put `additionalProperties { type: string }` in the schema; omitting this allows the custom fields to have arbitrary types and structure. I'm not sure how much such relaxing would constitute a spec change. @richvdh or @turt2live, do you know if there are already practical cases where non-string parameters are used in the context of `/thirdparty/user/` requests?

data/api/client-server/content-repo.yaml
```diff
@@ -12,154 +12,164 @@
...
-          schema:
-            "$ref": "definitions/errors/error.yaml"
-        429:
+              description: The name of the file that was previously uploaded, if set.
+              schema:
+                type: string
+          content:
+            application/octet-stream:
+              schema:
+                type: string
```
KitsuneRal, line 154:
```suggestion
                type: string
                format: binary
```

data/api/client-server/content-repo.yaml
```diff
@@ -170,68 +180,77 @@ paths:
...
-            }
-          schema:
-            "$ref": "definitions/errors/error.yaml"
-        429:
+              schema:
+                type: string
+          content:
+            application/octet-stream:
+              schema:
+                type: string
```
KitsuneRal, line 235:
```suggestion
                type: string
                format: binary
```

data/api/client-server/content-repo.yaml
```diff
@@ -12,154 +12,164 @@
...
+          application/json:
+            schema:
+              type: string
+              example: <bytes>
+              format: byte
+          "*/*":
+            schema:
+              type: string
+              example: <bytes>
+              format: byte
```
KitsuneRal, line 50: AIUI, the OpenAPI v3 spec [recommends](https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#data-types) using `format: byte` for base64-encoded payloads and `format: binary` for "any sequence of octets". The latter is more appropriate across content-repo.
```suggestion
              format: binary
```

KitsuneRal (changes requested):
A few more from the first part of `client-server/`

data/api/client-server/content-repo.yaml
zecakeh: This has actually changed in the [OpenAPI 3.1 spec](https://spec.openapis.org/oas/v3.1.0#considerations-for-file-uploads).
The spec says to write it like this:

```yaml
application/octet-stream: {}
```

data/api/client-server/content-repo.yaml
zecakeh: Same as above

data/api/client-server/content-repo.yaml
zecakeh: Same as above.

data/api/client-server/content-repo.yaml
KitsuneRal: Gotta say, I like that new notation; "binary strings" made me cringe when I first saw that...

data/api/application-service/protocols.yaml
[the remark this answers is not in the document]
richvdh: (some history on this: https://github.com/matrix-org/matrix-spec-proposals/pull/1414)

data/api/client-server/typing.yaml
```diff
@@ -40,48 +29,68 @@ paths:
...
+        required: true
       responses:
-        200:
+        "200":
           description: The new typing state was set.
-          examples:
-            application/json: {
-              }
-          schema:
-            type: object  # empty json object
```
KitsuneRal, line 81: Same as above, guess we can use `maxProperties: 0` here.

data/api/client-server/wellknown.yaml
```diff
@@ -11,18 +11,12 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
-swagger: '2.0'
+openapi: 3.1.0
 info:
-  title: "Matrix Client-Server Server Discovery API"
-  version: "1.0.0"
-host: localhost:8008
-schemes:
-  - https
```
KitsuneRal, line 20: And again `http` missing.

KitsuneRal (commented):
Okay, `client-server/` done.

I realise that the whole body will probably have to be re-converted from the most recent API definitions, so applying those suggestions now would not be prudent - but at least those can be used to apply changes after (or even before, like adding missing `http` schemes).

KitsuneRal (commented):
The rest except `server-server` is covered. Got a couple of ideas after reviewing the scripts - please find below.

scripts/openapi-convert/object-fixes.mjs
```diff
@@ -0,0 +1,47 @@
...
+        server.variables.protocol = {
+            enum: ["http", "https"],
+            default: "https",
+        }
+        obj.servers.pop();
+
+        server.url = "{protocol}://" + url.host + "{basePath}";
+    } else {
+        server.url = url.origin + "{basePath}";
+    }
```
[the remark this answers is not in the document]
KitsuneRal: Answering to myself - insecure HTTP would not be wanted in server-server federation. Still, it seems to me that we should rather impose one specific set of schemes for each API, instead of hunting after deviations in individual files.

data/api/server-server/leaving-v2.yaml
```diff
@@ -55,86 +43,90 @@ paths:
...
-            "$ref": "examples/minimal_pdu.json",
-            "type": "m.room.member",
-            "state_key": "@someone:example.org",
-            "origin": "example.org",
-            "origin_server_ts": 1549041175876,
-            "sender": "@someone:example.org",
-            "content": {
-                "membership": "leave"
-            }
-          }
```
KitsuneRal, line 129: Seems that this example is completely lost after conversion.

data/api/server-server/third_party_invite.yaml
```diff
@@ -41,166 +30,142 @@ paths:
...
-                  "token": "abc123",
-                  "signatures": {
-                    "magic.forest": {
-                      "ed25519:3": "fQpGIW1Snz+pwLZu6sTy2aHy/DYWWTspTJRPyNp0PKkymfIsNffysMl6ObMMFdIJhk6g6pwlIqZ54rxo8SLmAg"
-                    }
-                  }
-                }
-              }
-            }
-          }
```
KitsuneRal, line 193: This example seems to have been lost, too.

data/api/server-server/third_party_invite.yaml
```diff
@@ -210,116 +175,127 @@ paths:
...
-                    room_id:
-                      type: string
-                      description: The room ID the invite is valid for.
-                      example: "!somewhere:example.org"
-                    sender:
-                      type: string
-                      description: The user ID that sent the invite.
-                      example: "@bob:matrix.org"
-                    # TODO (TravisR): Make this reusable when doing IS spec changes
-                    #   also make sure it isn't lying about anything, like the key version
```
KitsuneRal, line 268: This comment needs salvaging as well.

data/api/server-server/third_party_invite.yaml
```diff
@@ -210,116 +175,127 @@ paths:
...
-                      properties:
-                        mxid:
-                          type: string
-                          description: |-
-                            The user ID that has been bound to the third party
-                            identifier.
-                          example: "@alice:matrix.org"
-                        token:
-                          type: string
-                          # TODO: What is this actually?
```
KitsuneRal, line 284: This is still a valuable TODO.

data/api/server-server/transactions.yaml
```diff
@@ -46,64 +35,70 @@ paths:
...
-                    type: array
-                    description: |-
-                      List of ephemeral messages. May be omitted if there are no ephemeral
-                      messages to be sent. Must not include more than 100 EDUs.
-                    items:
-                      $ref: "definitions/edu.yaml"
-          example: {
-            "$ref": "examples/transaction.json",
-            "edus": [{"$ref": "examples/edu.json"}]
-          }
```
KitsuneRal, line 73: Another lost example. Also, `$ref` cannot be combined with `edus` this way but this is already fixed in the most recent `main`.

data/api/server-server/user_devices.yaml
```diff
@@ -38,81 +27,89 @@ paths:
...
-              master_key:
+          content:
+            application/json:
+              schema:
                 type: object
-                description: |-
-                  The user\'s master cross-signing key.
-                allOf:
-                  - $ref: ../client-server/definitions/cross_signing_key.yaml
-                  # FIXME: why isn't the doc generator picking up this example?
```
KitsuneRal, line 91: I haven't checked whether the doc generator is still not picking up the example but it might well not because `allOf` should be wrapped in `schema` here.

data/api/server-server/user_devices.yaml
```diff
@@ -38,81 +27,89 @@ paths:
...
-                        "ed25519:base64+master+public+key": "base64+master+public+key",
-                      }
-                    }
-              self_signing_key:
-                type: object
-                description: |-
-                  The user\'s self-signing key.
-                allOf:
-                  - $ref: ../client-server/definitions/cross_signing_key.yaml
-                  # FIXME: why isn't the doc generator picking up this example?
```
KitsuneRal, line 105: In the same vein as above.

KitsuneRal (changes requested):
The last piece, `server-server/` completed. This one has been more problematic - aside from things I could debug (see the comments), most concerning is that the conversion has apparently lost a few examples. I guess it might be because of some weird `example`/`examples` confusion or the wrong level of nesting but didn't dig into this, honestly.

scripts/openapi-convert/string-fixes.mjs
```diff
@@ -0,0 +1,38 @@
...
+    console.log("%s", short_path);
+
+    let content = await fs.readFile(path, "utf-8");
+
+    // Fix occurrences of `*xx` status codes to `*XX`.
+    content = content.replaceAll("3xx:", "\"3XX\":");
+    content = content.replaceAll("4xx:", "\"4XX\":");
+
+    // Fix occurrences of `_ref` to `$ref`.
+    content = content.replaceAll("_ref:", "$ref:");
```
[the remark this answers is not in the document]
zecakeh: It seems to be a bug during the conversion. All the `securitySchemes` turn up like this:

```yaml
components:
  securitySchemes:
    _ref: definitions/security.yaml
```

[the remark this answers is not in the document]
KitsuneRal: Ah, this is likely because `securityDefinitions` is not supposed to include Reference objects as of Swagger 2.0.

scripts/openapi-convert/convert-definitions.mjs
```diff
@@ -0,0 +1,82 @@
...
+        // Don't deduplicate requestBodies.
+        resolveInternal: true,
+        // Write OpenAPI version 3.1.0, even if it's not completely true, it'll
+        // be fixed later.
+        targetVersion: '3.1.0',
+    });
+
+    // Apply fixes on object.
+    const obj = applyObjectFixes(options.openapi);
+
```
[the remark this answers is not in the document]
zecakeh: I had to modify this a bit to make it work:

```js
    const doc = new yaml.Document(obj);
    yaml.visit(doc, {
        Pair(_, p, path) {
            const parent = path[path.length - 2];

            if (parent.key && p.key && p.value && ((parent.key.value === 'response' && p.key.value === 'value') ||
                (parent.key.value === 'schema' && p.key.value === 'example'))) {
                if (!p.value.flow) {
                    p.value.flow = true
                    return p
                }
            }
        }
    });
```
The output does look more like JSON, except the strings are not double-quoted.

scripts/openapi-convert/object-fixes.mjs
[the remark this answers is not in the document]
richvdh: Yes, I think there may have been some idea that we would never want to run the identity service, or the federation protocol, over http. Still, it's not obvious to me that the api files are a good place to try to enforce that.

KitsuneRal:
What I would suggest after figuring out what to do on each item from the laundry lists above:
- apply preparatory commits where necessary (maybe even take them out to separate PRs, as before); I suspect we got examples defined in the wrong place or form in `server-server/`, fixing that looks like one case here.
- re-apply the conversion over the most recent `main`
- apply all the manual post-fixes
- force-push the branch _under the same name_
- I review the changes across the force-push (GitHub allows to do that)
- we plan follow-up fixes if any remain

zecakeh:
>  apply preparatory commits where necessary (maybe even take them out to separate PRs, as before); I suspect we got examples defined in the wrong place or form in `server-server/`, fixing that looks like one case here.

Thanks for taking the time to review everything, I'll try to open PRs for those in the next days.

data/api/client-server/administrative_contact.yaml
```diff
@@ -303,59 +317,63 @@ paths:
...
             The homeserver has disassociated the third party identifier from the
             user.
-          schema:
-            type: object
-            properties:
-              id_server_unbind_result:
-                type: string
-                enum:
-                  # XXX: I don't know why, but the order matters here so that "no-support"
-                  # doesn't become "no- support" by the renderer.
```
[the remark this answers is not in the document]
zecakeh: I tested building the spec locally and I don't see anything weird when changing the order so I guess this is no longer applicable

data/api/server-server/keys_query.yaml
```diff
@@ -35,99 +26,108 @@ paths:
...
-          type: string
           description: |-
             **Deprecated**. Servers should not use this parameter and instead
             opt to return all keys, not just the requested one. The key ID to
             look up.
 
             When excluded, the trailing slash on this endpoint is optional.
-          required: false
-          x-example: "ed25519:abc123"
+          required: true
```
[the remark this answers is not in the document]
zecakeh: `keyId` wqas removed in #1350.

data/api/server-server/keys_server.yaml
```diff
@@ -46,18 +39,26 @@ paths:
...
           description: |-
             **Deprecated**. Servers should not use this parameter and instead
             opt to return all keys, not just the requested one. The key ID to
             look up.
 
             When excluded, the trailing slash on this endpoint is optional.
-          required: false
-          x-example: "ed25519:abc123"
+          required: true
           deprecated: true
```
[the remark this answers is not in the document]
zecakeh: Same as above, `keyId` was removed in #1350.

data/api/client-server/create_room.yaml
```diff
@@ -76,176 +65,188 @@ paths:
...
+                      id_access_token:
+                        type: string
+                        description: |-
+                          An access token previously registered with the identity server. Servers
+                          can treat this as optional to distinguish between r0.5-compatible clients
+                          and this specification version.
+                      medium:
+                        type: string
+                        description: The kind of address being passed in the address field, for example
+                          `email`.
```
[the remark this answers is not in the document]
zecakeh: Opened #1417.

data/api/client-server/third_party_membership.yaml
```diff
@@ -71,70 +60,96 @@ paths:
...
+                id_access_token:
+                  type: string
+                  description: |-
+                    An access token previously registered with the identity server. Servers
+                    can treat this as optional to distinguish between r0.5-compatible clients
+                    and this specification version.
+                medium:
+                  type: string
+                  description: The kind of address being passed in the address field, for example
+                    `email`.
```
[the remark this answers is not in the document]
zecakeh: Same as above, opened #1417.

data/api/server-server/invites-v1.yaml
```diff
@@ -49,119 +38,111 @@ paths:
...
+                        event:
+                          $ref: definitions/invite_event.yaml
+                      required:
+                        - event
+              examples:
+                response:
+                  value:
+                    - 200
+                    - event:
+                        $ref: examples/minimal_pdu.json
```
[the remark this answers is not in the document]
zecakeh: Done in #1454.

data/api/server-server/user_devices.yaml
zecakeh: The comment is gone on `main` and it seems the example is picked up in the spec.

I don't think `allOf` needs to be wrapped in a `schema` since this already is in the `schema` of the response. However it seems wrong to me that the `type` and `description` are outside of the `allOf`.

scripts/package.json
zecakeh: I am marking this as resolved because updating this dependency right now complains about errors that seem to be gone after the upgrade.

zecakeh:
So we should be good to re-apply the upgrade after #1454 is merged and we agree on the correct fix for https://github.com/matrix-org/matrix-spec/pull/1310#discussion_r1070674654, unless I forgot something else that can be fixed before the upgrade.

data/api/server-server/user_devices.yaml
KitsuneRal: Unlike `$ref`, `allOf` doesn't forbid sibling keys; but I agree that it would be better to put the entire schema inside `allOf` for integrity. Also, `type: object` is not needed then as it's already inherited from (in this case) `cross_signing_key.yaml`.

scripts/openapi-convert/object-fixes.mjs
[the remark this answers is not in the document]
KitsuneRal: I think that the whole authority+base path piece should be out of the API definition scope: the thing is supposed to be generic. I wonder if there's a mechanism to have that part of the URL as some sort of a variable that is substituted per installation.

data/api/identity/v2_terms.yaml
```diff
@@ -11,104 +11,89 @@
...
-                    The policy. Includes a map of language (ISO 639-2) to language-specific
-                    policy information.
-                  properties:
-                    version:
-                      type: string
-                      description: |-
-                        The version for the policy. There are no requirements on what this
-                        might be and could be "alpha", semantically versioned, or arbitrary.
-                  required: ['version']
-                  # TODO: TravisR - Make this render
```
KitsuneRal, line 91: Missed one more comment to salvage.

data/api/client-server/capabilities.yaml
```diff
@@ -34,79 +25,91 @@ paths:
...
+                  capabilities:
                     type: object
+                    title: Capabilities
                     description: |-
-                      Capability to indicate if the user can change their password.
-                    title: ChangePasswordCapability
+                      The custom capabilities the server supports, using the
+                      Java package naming convention.
+                    additionalProperties:
+                      type: object
```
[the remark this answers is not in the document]
KitsuneRal: The PR to fix things: https://github.com/matrix-org/matrix-spec/pull/1455

data/api/client-server/typing.yaml
KitsuneRal: Proposed in #1455.

data/api/server-server/user_devices.yaml
KitsuneRal: Looking at examples in other places, having `type` and `description` outside of `allOf` is actually common and overrides whatever `allOf` comes with. I would not keep `type` because one should already come from `$ref` but that's not a big deal (and I see quite a bit of it across our APIs... so worth a separate cleanup PR, before or after the migration).

KitsuneRal:
I tried to tighten the definitions here and there a bit more - let's have another run once #1454 and #1455 are in.

scripts/openapi-convert/object-fixes.mjs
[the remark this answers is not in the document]
zecakeh: We could definitely do like the published API definitions, where the URL is defined as `https://{homeserver_url}` with a default value of `matrix-client.matrix.org` for the variable (I think it's mostly a convenience to have a default server to try requests in the API viewer). The scheme could also be left out as a choice entirely.

It seems the viewer at https://matrix.org/docs/api/ enforces the server to be ` https://matrix-client.matrix.org/`, I don't know if it's on purpose. If I try the RapiDoc viewer [here](https://rapidocweb.com/examples/example1.html) with the [latest client-server API definition](https://spec.matrix.org/latest/client-server-api/api.json) it offers me to change the default URL.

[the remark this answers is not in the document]
KitsuneRal: Ok, I'll make a separate suggestion in a minute.

scripts/openapi-convert/object-fixes.mjs
```diff
@@ -0,0 +1,47 @@
...
+
+    server.variables.basePath.default = url.pathname;
```
KitsuneRal, line 35: I'm not really good at JavaScript but here's the idea that leaves key parts up for individual tweaking while having reasonable defaults applied across all definitions.
```suggestion
    let server = {
        url: "",
        variables: {
            protocol: {},
            hostname: {
                default: ""
            },
            basePath: {
                default: ""
            }
        }
    }

    const url = new URL(obj.servers[0].url);

    if (obj.servers.length > 1) {
        // In our case several URLs always mean both http and https for the same
        // host.
        obj.servers.pop()
    }

    server.url = "{protocol}://{hostname}{basePath}"
    server.variables.protocol = {
        enum: ["http", "https"],
        default: "https",
    }
    server.variables.hostname.default = url.host
    server.variables.basePath.default = url.pathname
```

KitsuneRal:
Those PRs are all merged - we should be good to go?

zecakeh:
Upgrade re-done on main.

I have added back all comments and examples that were gone. I checked that it passed two validators (swagger-parser and redocly-cli) and that the generated spec has no unwanted changes.

KitsuneRal (commented):
Sorry for the delayed review. Things are mostly in order now, save for one place with the user agent string sliced into parts in some weird (and actually not correct) way, and examples being kind-of-but-not-quite JSON now. I suggest we go forward with the currently formatted examples, but please address the first comment. Then (after yet another rebase... sorry for that) it should be good enough for merging.

zecakeh:
Should I wait until #1495 is merged to rebase, to avoid a conflict?

uhoreg:
> Should I wait until #1495 is merged to rebase, to avoid a conflict?

It has been merged, so should be safe to rebase. :wink:

zecakeh:
Rebased on latest main.

I was unhappy to leave the examples as is so I wrote a dirty little JS script that converts them back to JSON. Hopefully that reduces the diff.

KitsuneRal:
Thanks for your patience, again. I think it is as good as it can be for now. The remaining comments can be addressed later. @matrix-org/spec-core-team, I guess it's quite a good moment to merge this, while Matrix v1.8 is still kind of away and there's still time to iron parts that still stick out?

KitsuneRal (approved):

zecakeh:
So I should rebase one last time to solve the conflicts so it can be merged?

richvdh:
> So I should rebase one last time to resolve the conflicts so it can be merged?

yes please - either rebase, or better yet, just *merge* main into your branch and fix up the conflicts by hand, so the changes since @KitsuneRal's review are easy to see. Once you've done so I will press the merge button.

zecakeh:
I improved the script to be able to run it on a single file at a time so I was able to make a merge commit. Once again I checked that they pass the validation and that the generated Matrix spec doesn't have any changes.

richvdh (approved):
:tada:

richvdh:
Thank you so so much to @zecakeh for all your hard work on this, and to @KitsuneRal for reviewing it. It's fantastic to finally have this landed

PR #1317, opened 2022-11-02
Stop rendering CS modules and room version fragments as standalone pages

richvdh:
This is actually doing two things:

 * creating `{fragments,modules}/index.md` turns the fragments and modules into page resources, rather than pages in their own right. We have to update the shortcodes to match, but we can also remove the filtering from the sidebar and the frontmatter from the module definitions.

 * adding `headless: true` means that we don't render the pages.

The net effect is that we don't render pages like https://spec.matrix.org/v1.4/rooms/fragments/v1-auth-rules/ and https://spec.matrix.org/v1.4/client-server-api/modules/account_data/.

Preview: https://pr1317--matrix-spec-previews.netlify.app


dbkr (approved):
Seems plausible

Issue #647, opened 2020-06-12
how much can default push rules be reconfigured?


richvdh:
evidently I am allowed to change the `actions` and the `enabled`-ness of a default push rule, but am I allowed to change the conditions it matches on? If not, what happens if I try to do so?

richvdh:
I think this is closed thanks to #1319

PR #1319, opened 2022-11-03
Clarify the behavior of `PUT /pushrules/{scope}/{kind}/{ruleId}`

### LINKED ISSUE #645: PushRule endpoint has wrong description
https://matrix.org/docs/spec/client_server/r0.6.1#put-matrix-client-r0-pushrules-scope-kind-ruleid looks like it was meant for /pushers

zecakeh:
This is based on the behavior of Synapse and Dendrite. Conduit's implementation is already non-compliant in regards to what was already defined in the spec.

Closes #645.
Related to #647 (probably closes it too, unless we want to be more explicit somewhere about what can be changed on default push rules).

Related PR in ruma that would allow to fix Conduit's implementation: https://github.com/ruma/ruma/pull/1364

Preview: https://pr1319--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/pushrules.yaml
+++ b/data/api/client-server/pushrules.yaml
@@ -365,9 +365,20 @@ paths:
     put:
       summary: Add or change a push rule.
       description: |-
-        This endpoint allows the creation, modification and deletion of pushers
-        for this user ID. The behaviour of this endpoint varies depending on the
-        values in the JSON body.
+        This endpoint allows the creation and modification of user defined push
+        rules.
+
+        If a rule with the same `rule_id` already exists among rules of the same
+        kind, it is updated with the new parameters, otherwise a new rule is
+        created.
+
+        If both `after` and `before` are provided, the new or updated rule must
+        be the next most important rule with respect to the rule identified by
+        `before`.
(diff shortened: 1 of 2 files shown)


data/api/client-server/pushrules.yaml
```diff
@@ -365,9 +365,20 @@ paths:
...
+        This endpoint allows the creation and modification of user defined push
+        rules.
+
+        If a rule with the same `rule_id` already exists among rules of the same
+        kind, it is updated with the new parameters, otherwise a new rule is
+        created.
+
+        If both `after` and `before` are provided, the new or updated rule must
+        be the next most important rule with respect to the rule identified by
+        `before`.
```
zecakeh, line 377: In this case, Synapse ignore `after` completely, while Dendrite checks that the `after` rule exists and is before `before`.

Dendrite's behavior seems more sensible here, even if it seems there's no real value in setting them both anyway.

dbkr (approved):
I think this seems fine - thanks. Perhaps someone else wants to double check.

richvdh:
lgtm

Issue #233, opened 2017-03-05
Improve delete semantics for pushers


mujx:

richvdh:
Another reason for this is that there's basically nowhere else in the protocol that we allow `null` (cf #472)

PR #1321, opened 2022-11-03
Clarify what fields are required when deleting a pusher

zecakeh:
The required fields match those used by Synapse, Dendrite and Conduit to delete existing pushers.

Related: #233

Preview: https://pr1321--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/pusher.yaml
+++ b/data/api/client-server/pusher.yaml
@@ -182,7 +182,9 @@ paths:
                   If the `kind` is `"email"`, this is the email address to
                   send notifications to.
               kind:
-                type: string
+                type:
+                  - "string"
+                  - "null"
                 description: |-
                   The kind of pusher to configure. `"http"` makes a pusher that
                   sends HTTP pokes. `"email"` makes a pusher that emails the
--- a/data/api/client-server/pusher.yaml
+++ b/data/api/client-server/pusher.yaml
@@ -139,6 +139,11 @@ paths:
         This endpoint allows the creation, modification and deletion of [pushers](/client-server-api/#push-notifications)
         for this user ID. The behaviour of this endpoint varies depending on the
         values in the JSON body.
+
+        If `kind` is not `null`, the pusher with this `app_id` and `pushkey`
+        for this user is updated, or it is created if it doesn't exist. If
+        `kind` is `null`, the pusher with this `app_id` and `pushkey` for this
+        user is deleted.
       operationId: postPusher
       security:
         - accessToken: []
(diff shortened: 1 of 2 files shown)


data/api/client-server/pusher.yaml
```diff
@@ -209,14 +214,15 @@ paths:
...
               data:
                 type: object
                 description: |-
-                  A dictionary of information for the pusher implementation
-                  itself. If `kind` is `http`, this should contain `url`
-                  which is the URL to use to send notifications to.
+                  Required if `kind` is not `null`. A dictionary of information
+                  for the pusher implementation itself. If `kind` is `http`,
+                  this should contain `url` which is the URL to use to send
+                  notifications to.
```
dbkr, line 225: I'm not sure this is necessarily required in principle, although in practice all (both) types of pusher need some information here.

zecakeh: Synapse requires the field to be present: https://github.com/matrix-org/synapse/blob/develop/synapse/rest/client/pusher.py#L95-L106

Is there really something required for an `email` pusher? The spec doesn't say so.

zecakeh: Does that mean that it should also not be required on the `GET /_matrix/client/v3/pushers` endpoint?

dbkr: Hmm - probably: I'd expect most to include it anyway to guard against lazy clients, but that doesn't mean the spec should force them to. What do others think?

data/api/client-server/pusher.yaml
```diff
@@ -139,6 +139,11 @@ paths:
         This endpoint allows the creation, modification and deletion of [pushers](/client-server-api/#push-notifications)
         for this user ID. The behaviour of this endpoint varies depending on the
         values in the JSON body.
+
+        If `kind` is not `null`, the pusher with this `app_id` and `pushkey`
```
richvdh, line 143: we're probably supposed to change the `type` of `kind`. In OpenAPIv3 I think that means adding `nullable: true`, but it might be harder in v2? Maybe just add a comment for now?

data/api/client-server/pusher.yaml
richvdh: Mostly, I think that making `data` optional here is even more of a divergence from the established spec, and harder to justify without an MSC. We can make an exception if `kind` is `null`, because (a) it's clearly nonsense to require a `data` in that case and (b) that's what synapse has done forever, but going beyond that seems a stretch.

If we want to relax the requirements for `data` when `kind` is *not* `null`, let's make it a separate change.

richvdh:
> This is technically not concerning this msc

It's not an MSC.

> but the topic: is there a reason why this weird way of unsetting/explicit null is user over a REST DELETE endpoint?

No. That is the topic of https://github.com/matrix-org/matrix-spec/issues/233.

data/api/client-server/pusher.yaml
zecakeh: There are already some uses of `null` in the current definitions, like this:

```yaml
type:
    - "string"
    - "null"
```

OpenAPI v3 does use `nullable: true`, but v3.1 uses the same syntax as above.

richvdh (approved):
lgtm. Thanks!

Issue #1336, opened 2022-11-14
CI currently failing with broken links


turt2live:
eg https://github.com/matrix-org/matrix-spec/actions/runs/3423504915/jobs/5702191261

richvdh:
that seems to be because the CI for main sticks an `/unstable` on the front of all the links, which I think is easily accounted for in the link checker

richvdh:
There are some actual broken links; Hugo seems to be generating links to RSS with too many `/unstable`s in the URL:

`<link rel="alternate" type="application/rss&#43;xml" href="/unstable/unstable/index.xml">`

PR #1338, opened 2022-11-15
Unpack spec into right place for htmltest

richvdh:
Hopefully, fixes https://github.com/matrix-org/matrix-spec/issues/1336

Preview: https://pr1338--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Changes look plausible and the build for this PR worked. Let's merge and see what happens!

PR #1340, opened 2022-11-15
Generate unstable changelogs using towncrier

richvdh:
Replace the current stack of hugo templates with a towncrier invocation. The main advantage of this is that it means that the "Changes since last release" section is consistent with the changelogs for the actual releases.

This also changes the release process so that the changelog is generated *before* tagging, which means that the thing tagged v1.5 is actually the v1.5 spec.

Fixes https://github.com/matrix-org/matrix-spec/issues/908.

Suggest reviewing commit-by-commit.

Preview: https://pr1340--matrix-spec-previews.netlify.app


turt2live (commented):
clearing code review while draft, also just a note to please update the release documentation to accomodate the change.

layouts/shortcodes/changelog/changelogs.html
```diff
@@ -0,0 +1,10 @@
+{{/*
+   This template is used to render all of the changelog sections under
+   "content/changelogs"
+*/}}
+
+{{ with .Page.Resources.Match "*.md" }}
+{{ range ((sort . "Name" "desc")) }}
```
turt2live, line 7: will this cause a problem when we get to v1.10 or v10.0?

turt2live (commented):
otherwise lgtm

layouts/shortcodes/changelog/changelogs.html
richvdh: bleh. yes.

richvdh: I've fixed this by including the timestamp of when the changelog was generated in the file, and sorting by that instead.

turt2live (approved):
thanks :) 

and doubly thanks for going back in time to find the dates of each release.

turt2live:
> > and doubly thanks for going back in time to find the dates of each release.
> 
> I mean they are written three lines below in each changelog, but I'll take the thanks :)

considering that's more effort than I put into reading the changelogs, thank you 😇

Issue #1296, opened 2022-10-19
Update feature profiles table


turt2live:
see https://github.com/matrix-org/matrix-spec/pull/1206#discussion_r998394786 - tldr is a bunch of recent code has landed but the table hasn't been updated.

Everything that has landed is optional for all profiles, to my knowledge.

PR #1344, opened 2022-11-15
Add Event Replacements, Threading and Reference Relations modules to module summary table

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/1296.

The issue mentions that all new modules are optional. Threading is of course not a required feature. Event replacements are not required as we have fallbacks for those. Reference relations are optional too...?

Preview: https://pr1344--matrix-spec-previews.netlify.app


uhoreg (approved):
Looks like it does what it says it does

PR #1349, opened 2022-11-17
Include examples inline instead of using a reference for invite endpoint definitions

zecakeh:
The OpenAPI 3.x spec doesn't allow building examples by composition. Either the whole example must be a reference, or all of it has to be included.

Preview: https://pr1349--matrix-spec-previews.netlify.app


richvdh (approved):
lgtm. Thanks!

PR #1350, opened 2022-11-17
Remove `keyId` from the server keys endpoints

zecakeh:
Spec PR for [MSC3938](https://github.com/matrix-org/matrix-spec-proposals/pull/3938).

Preview: https://pr1350--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/1350.breaking
+++ b/changelogs/server_server/newsfragments/1350.breaking
@@ -1,1 +1,1 @@
-Remove `keyId` from the server keys endpoints.
+Remove `keyId` from the server `/keys` endpoints, as per [MSC3938](https://github.com/matrix-org/matrix-spec-proposals/pull/3938).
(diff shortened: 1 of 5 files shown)


turt2live (changes requested):
Deprecated endpoints require an MSC to remove them.

richvdh:
> Deprecated endpoints require an MSC to remove them.

I mean, really? Isn't the fact that they are deprecated enough of an indication that if people care about them they need to make a case to *undeprecate* them?

richvdh:
ugh, what a pain in the ass. I think we should revise that.

turt2live:
it'd only take an MSC 😇 

I'm fairly opposed to changing the policy here, but that's a discussion for elsewhere.

richvdh:
I've opened https://github.com/matrix-org/matrix-spec-proposals/pull/3938.

zecakeh:
I updated the description and changelog entry to mention [MSC3938](https://github.com/matrix-org/matrix-spec-proposals/pull/3938).

turt2live (approved):
Thanks!

PR #1351, opened 2022-11-17
Fix `POST _matrix/federation/v1/user/keys/claim` response schema

zecakeh:
The syntax was not compliant with the Swagger spec (the `type` field cannot be an array).

It also lacked one level of nesting (or maybe that's what the syntax was trying to do).

Preview: https://pr1351--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/1351.clarification
+++ b/changelogs/server_server/newsfragments/1351.clarification
@@ -1,0 +1,1 @@
+Fix `POST _matrix/federation/v1/user/keys/claim` response schema.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
looks good. thanks!

KitsuneRal:
Sorry for taking time to catch up with this - actually, passing an array for `type` is yet another of our [extensions](https://github.com/matrix-org/matrix-spec/blob/main/openapi_extensions.md) for Swagger 2, as it doesn't have `oneOf`. The fix of nesting is correct (the added comments really help to navigate a stack of `additionalProperties`...) - but the definition still needs the `string` type alternative next to `KeyObject`. I guess it's best to just not forget to fix that in #1310.

PR #1355, opened 2022-11-22
Remove duplicated "Note:" from `m.room.power_levels`

johannescpk:
Follow-up to https://github.com/matrix-org/matrix-spec/pull/1169. While looking at the preview from #1169, I noticed that the section is already called "Note:".

Signed-off-by: Johannes Becker <j.becker@famedly.com>

Preview: https://pr1355--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1355.clarification
+++ b/changelogs/client_server/newsfragments/1355.clarification
@@ -1,1 +1,1 @@
-Remove duplicated "Note:" section prefix.
+Clarify the power levels integer range.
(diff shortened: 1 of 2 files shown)


changelogs/client_server/newsfragments/1355.clarification
```diff
@@ -0,0 +1 @@
+Remove duplicated "Note:" section prefix.
```
richvdh, line 1: Using the same changelog text will cause the changelog entries to be merged.
```suggestion
Clarify the power levels integer range.
```

richvdh (approved):

PR #1370, opened 2022-12-01
replace "<content>" with "content"

dylhack:
This parameter that's part of the content-repo OpenAPI spec causes generators to mess up.

### DIFF
--- a/data/api/client-server/content-repo.yaml
+++ b/data/api/client-server/content-repo.yaml
@@ -49,7 +49,7 @@ paths:
           name: filename
           description: The name of the file being uploaded
         - in: body
-          name: "<content>"
+          name: "content"
           description: The content to be uploaded.
           required: true
           x-example: "<bytes>" # so the spec shows "<bytes>" without quotes.


dylhack:
note: #1310 fixes this

turt2live:
if https://github.com/matrix-org/matrix-spec/pull/1310 fixes this, would it be better to pursue that option instead?

dylhack:
Yes if it were quicker to merge. In this #1310  can manage the conflicts as long as that means code generators work today.

turt2live:
For the speed the spec moves at, #1310 is likely to land before the next spec release at the current rate - this seems suitable for conde generators too, as using unstable API definitions is, well, unstable.

KitsuneRal (changes requested):
The changes looks good but please add a news fragment, as described here: https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog

KitsuneRal (approved):

KitsuneRal:
This snippet doesn't even land in the spec text, it seems - and #1310 will have to be updated (merged or rebased) to the latest `main` anyway, so I suggest merging it. Up to @turt2live though.

changelogs/internal/newsfragments/1370.clarification
```diff
@@ -0,0 +1,2 @@
+Replace "<content>" with "content" for the content repo OpenAPI
```
turt2live, line 1:
```suggestion
Rename "<content>" to "content" in the OpenAPI files for content uploads.
```

turt2live (commented):
otherwise lgtm, as it sounds like @KitsuneRal is okay with managing conflicts

dylhack:
Awesome thanks

turt2live:
For future changes, please note that we normally require [Sign Off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off) to ensure the change can be incorporated into the project. For extraordinarily trivial changes like this however, we tend to bypass this requirement.

PR #1384, opened 2022-12-21
Stop autogenerating examples where we already have one

richvdh:
If an object definition already has an example, we shouldn't try to extend that
definition by adding examples derived from the individual properties. Doing so
is confusing, and there is no way to inhibit it when it is not desired. It's
also not what the RapiDoc viewere does, so we end up with examples being
inconsistent.

Preview: https://pr1384--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/administrative_contact.yaml
+++ b/data/api/client-server/administrative_contact.yaml
@@ -210,14 +210,12 @@ paths:
               client_secret:
                 type: string
                 description: The client secret used in the session with the homeserver.
+                example: "d0nt-T3ll"
               sid:
                 type: string
                 description: The session identifier given by the homeserver.
+                example: "abc123987"
             required: ["client_secret",  "sid"]
-            example: {
-              "sid": "abc123987",
-              "client_secret": "d0nt-T3ll"
-            }
       responses:
         200:
           description: The addition was successful.
--- a/data/api/client-server/cross_signing.yaml
+++ b/data/api/client-server/cross_signing.yaml
@@ -75,6 +75,11 @@ paths:
                 allOf:
                 - $ref: "definitions/auth_data.yaml"
             example: {
+              "auth": {
+                "type": "example.type.foo",
+                "session": "xxxxx",
+                "example_credential": "verypoorsharedsecret"
+              },
               "master_key": {
                 "user_id": "@alice:example.com",
                 "usage": ["master"],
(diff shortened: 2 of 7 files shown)


data/api/client-server/administrative_contact.yaml
```diff
@@ -210,14 +210,12 @@ paths:
...
+                example: "d0nt-T3ll"
               sid:
                 type: string
                 description: The session identifier given by the homeserver.
+                example: "abc123987"
             required: ["client_secret",  "sid"]
-            example: {
-              "sid": "abc123987",
-              "client_secret": "d0nt-T3ll"
-            }
```
richvdh, line 220: this is so that we still get an example for `auth`.

data/api/client-server/cross_signing.yaml
```diff
@@ -75,6 +75,11 @@ paths:
                 allOf:
                 - $ref: "definitions/auth_data.yaml"
             example: {
+              "auth": {
+                "type": "example.type.foo",
+                "session": "xxxxx",
+                "example_credential": "verypoorsharedsecret"
+              },
```
richvdh, line 82: again, this means that we still get an example for `auth`.

data/api/identity/definitions/request_email_validation.yaml
```diff
@@ -12,11 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 type: object
-example: {
-  "client_secret": "monkeys_are_GREAT",
-  "email": "foo@example.com",
-  "send_attempt": 1
-}
```
richvdh, line 19: this is redundant (the individual fields all have examples), but also, this object is inherited elsewhere and this example was overriding the examples for the fields in the derived object.

data/api/identity/definitions/request_msisdn_validation.yaml
```diff
@@ -12,12 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 type: object
-example: {
-  "client_secret": "monkeys_are_GREAT",
-  "country": "GB",
-  "phone_number": "07700900001",
-  "send_attempt": 1
-}
```
richvdh, line 20: ditto

data/api/server-server/user_devices.yaml
```diff
@@ -88,7 +88,6 @@ paths:
                   The user\'s master cross-signing key.
                 allOf:
                   - $ref: ../client-server/definitions/cross_signing_key.yaml
-                  # FIXME: why isn't the doc generator picking up this example?
```
richvdh, line 91: ... this was why.

dbkr (approved):

PR #1393, opened 2023-01-03
Spec `omit_members` for `/v2/send_join`

richvdh:
Per MSC3706


richvdh:
I haven't written much about how this is meant to be used, since faster-joins are still somewhat in development, and that's really the topic of MSC3902. It's really just meant to lay down the mechanism on which faster-joins are based.

data/api/server-server/joins-v1.yaml
agschaid, line 1: did you forget to change the example of the sendJoinV1 response? 
The `event` should be gone, I suppose.

PR #1412, opened 2023-01-24
Define hkdf-hmac-sha256.v2 MAC method for SAS verification

uhoreg:
Spec PR for https://github.com/matrix-org/matrix-spec-proposals/pull/3783

Changes are in https://pr1412--matrix-spec-previews.netlify.app/client-server-api/#short-authentication-string-sas-verification

Preview: https://pr1412--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1412.feature
+++ b/changelogs/client_server/newsfragments/1412.feature
@@ -1,0 +1,1 @@
+Define `hkdf-hmac-sha256.v2` MAC method for SAS verification, as per [MSC 3783](Define hkdf-hmac-sha256.v2 MAC method for SAS verification).
--- a/changelogs/client_server/newsfragments/1412.feature
+++ b/changelogs/client_server/newsfragments/1412.feature
@@ -1,1 +1,1 @@
-Define `hkdf-hmac-sha256.v2` MAC method for SAS verification, as per [MSC 3783](Define hkdf-hmac-sha256.v2 MAC method for SAS verification).
+Define `hkdf-hmac-sha256.v2` MAC method for SAS verification, as per [MSC 3783](https://github.com/matrix-org/matrix-spec-proposals/pull/3783).
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -668,17 +668,20 @@ The process between Alice and Bob verifying each other would be:
     the users to select a method.
 14. Alice and Bob compare the strings shown by their devices, and tell
     their devices if they match or not.
-15. Assuming they match, Alice and Bob's devices calculate the HMAC of
-    their own device keys and a comma-separated sorted list of the key
-    IDs that they wish the other user to verify, using SHA-256 as the
-    hash function. The HMAC calculation is defined [below](#mac-calculation).
+15. Assuming they match, Alice and Bob's devices each calculate Message
+    Authentication Codes (MACs) for:
+    * the keys that they wish to verify (usually their device ed25519 key and
+      their master cross-signing key)
+    * the list of key IDs that they wish the other user to verify.
+
+    The MAC calculation is defined [below](#mac-calculation).
 16. Alice's device sends Bob's device an `m.key.verification.mac`
     message containing the MAC of Alice's device keys and the MAC of her
     key IDs to be verified. Bob's device does the same for Bob's device
     keys and key IDs concurrently with Alice.
 17. When the other device receives the `m.key.verification.mac` message,
-    the device calculates the HMAC of its copies of the other device's
-    keys given in the message, as well as the HMAC of the
+    the device calculates the MAC of its copies of the other device's
+    keys given in the message, as well as the MAC of the
     comma-separated, sorted, list of key IDs in the message. The device
     compares these with the HMAC values given in the message, and if
     everything matches then the device keys are verified.
@@ -766,11 +769,15 @@ following error codes are used in addition to those already specified:
 
 ###### MAC calculation
 
-For verification of each party's device keys, a MAC is calculated individually
-for each the keys that are to be verified. As well, a MAC is calculated for a
-list of the keys IDs. The MAC used is HMAC as defined in [RFC
-2104](https://tools.ietf.org/html/rfc2104), using SHA-256 as the hash function.
-The HMAC key is calculated using HKDF as defined in [RFC
+During the verification process, Message Authentication Codes (MACs) are calculated
+for keys and lists of key IDs.
+
+The method used to calculate these MACs depends upon the value of the
+`message_authentication_code` property in the [`m.key.verification.accept`](#mkeyverificationaccept)
+message. All current implementations should use the `hkdf-hmac-sha256.v2` method which is
+defined as follows:
+
+The MAC used is HMAC as defined in [RFC
 5869](https://tools.ietf.org/html/rfc5869), using SHA-256 as the hash
 function. The shared secret is supplied as the input keying material. No salt
 is used, and in the info parameter is the concatenation of:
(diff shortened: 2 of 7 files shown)


changelogs/client_server/newsfragments/1412.feature
```diff
@@ -0,0 +1 @@
+Define `hkdf-hmac-sha256.v2` MAC method for SAS verification, as per [MSC 3783](Define hkdf-hmac-sha256.v2 MAC method for SAS verification).
```
richvdh, line 1: This MSC link looks wrong

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -671,10 +671,7 @@ The process between Alice and Bob verifying each other would be:
 15. Assuming they match, Alice and Bob's devices calculate the HMAC of
     their own device keys and a comma-separated sorted list of the key
     IDs that they wish the other user to verify, using SHA-256 as the
-    hash function. HMAC is defined in [RFC
-    2104](https://tools.ietf.org/html/rfc2104). The key for the HMAC is
-    different for each item and is calculated by generating 32 bytes
-    (256 bits) using [the key verification HKDF](#hkdf-calculation).
+    hash function. The HMAC calculation is defined [below](#mac-calculation).
```
richvdh, line 674: I find it confusing the way we switch freely between "HMAC" and "MAC", and as far as I can tell we're talking about the same thing.

Also, I think we're calculating two separate MACs here? The current phrasing suggests that all the inputs are smashed together to calculate a single MAC.

I don't think we need to say we use SHA-256 here: that's part of the MAC calculation.

I also don't think we need to specify the format of the key IDs list; again that can go below.

```suggestion
15. Assuming they match, Alice and Bob's devices each calculate Message
    Authentication Codes (MACs) for:
    * their own device key [XXX: ed25519 key? curve25519 key? something else?], and:
    * the list of key IDs that they wish the other user to verify.
    The MAC calculation is defined [below](#mac-calculation).
```

Also s/HMAC/MAC/ in step 17.

richvdh: stupid question, but why are the key IDs to be verified different from the device key?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -767,7 +764,42 @@ following error codes are used in addition to those already specified:
...
+-   The Matrix ID of the other user.
+-   The Device ID of the device receiving the MAC.
+-   The `transaction_id` being used.
+-   The Key ID of the key being MAC-ed, or the string `KEY_IDS` if the
+    item being MAC-ed is the list of key IDs.
+
+If the key list is being MACed, the list is sorted lexicographically and
+comma-separated with no extra whitespace added. In this way, the recipient can
+reconstruct the list from the names in the `mac` property of the
+`m.key.verification.mac` message and ensure that no keys were added or removed.
```
richvdh, line 790: We say how to format a key list, but not a device key. Shouldn't we include how to format a device key?

data/event-schemas/schema/m.key.verification.start$m.sas.v1.yaml
```diff
@@ -42,7 +42,10 @@ properties:
         type: array
         description: |-
           The message authentication codes that the sending device understands.
```
richvdh, line 44: We're not negotiating codes per se, but rather methods for generating them.

```suggestion
          The message authentication code methods that the sending device understands.
```

similarly in `m.key.verification.accept`.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -767,7 +764,42 @@ following error codes are used in addition to those already specified:
 
 {{% event event="m.key.verification.mac" %}}
 
-###### HKDF calculation
+###### MAC calculation
+
+For verification of each party's device keys, a MAC is calculated individually
+for each the keys that are to be verified. As well, a MAC is calculated for a
+list of the keys IDs. The MAC used is HMAC as defined in [RFC
```
richvdh, line 771:
```suggestion
During the verification process, Message Authentication Codes (MACs) are calculated
for device keys and lists of key IDs.

The method used to calculate these MACs depends upon the value of the 
`message_authentication_code` property in the [`m.key.verification.accept`](#mkeyverificationaccept)
message. All current implementations should use the `hkdf-hmac-sha256.v2` method which is
defined as follows:

The MAC used is HMAC as defined in [RFC
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -767,7 +764,42 @@ following error codes are used in addition to those already specified:
...
+    item being MAC-ed is the list of key IDs.
+
+If the key list is being MACed, the list is sorted lexicographically and
+comma-separated with no extra whitespace added. In this way, the recipient can
+reconstruct the list from the names in the `mac` property of the
+`m.key.verification.mac` message and ensure that no keys were added or removed.
+
+{{% boxes/note %}}
+The MAC method `hkdf-hmac-sha256` used an incorrect base64 encoding, due to a
+bug in the original implementation in libolm.  To remedy this,
```
richvdh, line 794: The algorithm above doesn't seem to use base64 anywhere, so I'm a bit confused about how we could be using an incorrect encoding.

In other words: I think the description of the method is missing something about encoding. Presumably it's how we take the output of the HMAC and stuff it into a JSON text property?

richvdh (changes requested):
I'm afraid I am struggling to keep up.

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: > stupid question, but why are the key IDs to be verified different from the device key?

The key IDs to be verified may include the cross-signing keys

content/client-server-api/modules/end_to_end_encryption.md
uhoreg: It's mentioned in the `m.key.verification.mac` definition. But I'll add a mention here too.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -767,7 +767,51 @@ following error codes are used in addition to those already specified:
...
+-   The `transaction_id` being used.
+-   The Key ID of the key being MAC-ed, or the string `KEY_IDS` if the
+    item being MAC-ed is the list of key IDs.
+
+If the key list is being MACed, the list is sorted lexicographically and
+comma-separated with no extra whitespace added, with each key written in the
+form `{algorithm}:{keyId}`. For example, the key list could look like:
+`ed25519:Cross+Signing+Key,ed25519:DEVICEID`. In this way, the recipient can
+reconstruct the list from the names in the `mac` property of the
+`m.key.verification.mac` message and ensure that no keys were added or removed.
```
richvdh, line 799: we still don't seem to say how to feed a key into the MAC.

Do we use the 43(?) bytes of Base64-encoded public key? Or the 32 bytes of binary public key? Or other?

richvdh (changes requested):

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -668,20 +668,20 @@ The process between Alice and Bob verifying each other would be:
...
-    IDs that they wish the other user to verify, using SHA-256 as the
-    hash function. HMAC is defined in [RFC
-    2104](https://tools.ietf.org/html/rfc2104). The key for the HMAC is
-    different for each item and is calculated by generating 32 bytes
-    (256 bits) using [the key verification HKDF](#hkdf-calculation).
+15. Assuming they match, Alice and Bob's devices each calculate Message
+    Authentication Codes (MACs) for:
+    * the keys that they wish to verify (usually their device ed25519 key and
+      their master cross-signing key)
+    * the list of key IDs that they wish the other user to verify.
```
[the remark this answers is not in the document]
uhoreg: Yes, the key list is to prevent people from adding or removing keys

richvdh (approved):
looks great now, thanks!

PR #1415, opened 2023-01-25
Improve formatting of definitions in the Push Notifications section

HarHarLinks:
Signed-off-by: Kim Brose <2803622+HarHarLinks@users.noreply.github.com>

I thought I skipped a beat when I tried to read this https://spec.matrix.org/unstable/client-server-api/#push-notifications:
image

The same idea might apply to the rest of the spec, but I don't have time to search for it.

Preview: https://pr1415--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1415.clarification
+++ b/changelogs/internal/newsfragments/1415.clarification
@@ -1,0 +1,1 @@
+Improve formatting of push.md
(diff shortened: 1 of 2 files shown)


uhoreg (changes requested):
Good catch.  Looks like this was [originally](https://www.matrix.org/docs/spec/client_server/r0.6.0.html#id89) a definition list, but it got lost in the conversion.  Could you try switching it use the [Markdown definition list markup](https://michelf.ca/projects/php-markdown/extra/#def-list).  There's an example a bit later, under the Push Rules > Actions section (line ~190 of push.md).

uhoreg (approved):

uhoreg:
Thanks!

PR #1422, opened 2023-01-31
Clarify room version grammar

### LINKED ISSUE #1408: Room version grammar is inconsistent
**Link to problem area**: https://spec.matrix.org/v1.5/rooms/#room-version-grammar

**Issue**
> Room versions MUST NOT be empty and SHOULD contain only the characters `a-z`, `0-9`, `.`, and `-`.

but then we go on to describe a grammar which doesn't honour the "SHOULD": is it legal to put emoji (for example) in room versions or not?

richvdh:
It is not legal to use characters outside `a-z`, `0-9`, `.`, and `-`.

Fixes #1408 

Preview: https://pr1422--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1422.clarification
+++ b/changelogs/room_versions/newsfragments/1422.clarification
@@ -1,0 +1,1 @@
+Clarify the grammar for room versions.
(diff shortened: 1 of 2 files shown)


turt2live (approved):
see https://github.com/matrix-org/matrix-spec/issues/1408#issuecomment-1410920156 for rationale.

tldr is this would normally require an MSC to change, but given the history of this issue that feels overly silly.

PR #1424, opened 2023-01-31
Clarify terminology around aggregations

### LINKED ISSUE #1405: Aggregations are confusing
An `m.replace` event is aggregated by a server to (sometimes) create a bundle: none of these words are intuitive. 

See https://spec.matrix.org/v1.4/client-server-api/#server-side-aggregation-of-mreplace-relationships and related sections.

richvdh:
I've done my best to remove the word "bundle", because I feel like it causes more confusion than clarification. Instead I have favoured "aggregated child events" which I think is clearer.

Some general clarification around these parts of the spec.

Hopefully, fixes #1405 

Preview: https://pr1424--matrix-spec-previews.netlify.app


data/api/client-server/threads_list.yaml
```diff
@@ -32,8 +32,7 @@ paths:
       x-addedInMatrixVersion: "1.4"
       summary: Retrieve a list of threads in a room, with optional filters.
       description: |-
-        Paginates over the thread roots in a room, ordered by the `latest_event` of each thread root
-        in its bundle.
+        Paginates over the thread roots in a room.
```
turt2live, line 35: The description appears to have lost some context here?

meta/documentation_style.rst
```diff
@@ -44,33 +44,36 @@ Stylistic notes
...
-*name* of a property - and be mindful of the scope for confusion with cryptographic
-keys.
+* When talking about properties in JSON objects, prefer the word "property" to "field",
+  "member", or various other alternatives. For example: "this property will be set to
+  X if ...". Also avoid the term "key" unless you are specifically talking about the
+  *name* of a property - and be mindful of the scope for confusion with cryptographic
+  keys.
+
+* "i.e." (*id est*) is an abbreviation and hence is written with two full
+  stops. So is "e.g." (*exempli gratia*).
```
turt2live, line 76: This should probably be its own PR so it doesn't get buried in our future search history.

Also I'm not currently convinced it's a thing we [should] do - given the first point is that the language should be unambiguous, shouldn't we be using complete words/sentences instead? ("For example", "As demonstrated below", "..., such as when", etc)

turt2live (commented):
Generally this lgtm with a couple (hopefully easy) comments to resolve.

data/api/client-server/threads_list.yaml
richvdh: It seemed redundant to specify the sort order here as well as in the response.

meta/documentation_style.rst
richvdh: I'll move it out of here.

data/api/client-server/threads_list.yaml
turt2live: it adds a bit of clarity without having to scroll past a whole request schema, though not all that bothered.

The `summary` and `description` should be switched if we are losing the text though: it's more awkward if the `summary` is longer than the `description`.

richvdh: Have updated a bit more, with reference to https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#openapi.

If you still don't like it, am happy to put it back how it was.

turt2live (approved):

PR #1433, opened 2023-02-13
Improve distinction between tags and their attributes in the rich text section

deepbluev7:
Originally the tags used to be bold, followed by a colon and separated into two columns in a table. This at least restores the table aspect, which makes it clear, that font is not an attribute (and similar).

This seems to have gotten lost in the transition to the new design.

Preview: https://pr1433--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1433.clarification
+++ b/changelogs/client_server/newsfragments/1433.clarification
@@ -1,0 +1,1 @@
+Improve distinction between tags and their attributes in the rich text section. Contributed by Nico.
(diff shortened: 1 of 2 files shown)


deepbluev7:
Very old:

image

Current:

image

With this change:

image

(it's growing!)

uhoreg (approved):
The gray background on the tag/attribute names on top of the gray background on the odd rows is a bit suboptimal, but it's good enough. Thanks for the patch.

PR #1439, opened 2023-02-16
Clarify that reply chain fallback for threads may not be present

jplatte:
Discussed [in #matrix-spec](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$u987mZ4qSTDa3_pERBap2VQPWvJrVeyQm1wXZ7Uda_I?via=matrix.org&via=libera.chat&via=element.io).

Preview: https://pr1439--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1439.clarification
+++ b/changelogs/client_server/newsfragments/1439.clarification
@@ -1,1 +1,1 @@
-Clarify that reply chain fallback for threads is not required.
+Clarify that reply chain fallback for threads may not be present.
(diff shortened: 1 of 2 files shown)


changelogs/client_server/newsfragments/1439.clarification
```diff
@@ -0,0 +1 @@
+Clarify that reply chain fallback for threads is not required.
```
richvdh, line 1:
```suggestion
Clarify that reply chain fallback for threads may not be present.
```

richvdh (approved):
thanks!

Issue #688, opened 2020-09-01
Spec writing: put common values like senders, room IDs, aliases, etc in a shared location 


turt2live:
At the moment each example is responsible for trying to be consistent with all its `$ref`s, which is a bit annoying. It'd be nice to be able to say `%(roomId)` or similar and have the spec generator automatically populate that field with a known value, allowing for consistency.

PR #1440, opened 2023-02-21
Change the server aggregation for edits

richvdh:
per https://github.com/matrix-org/matrix-spec-proposals/pull/3925

### DIFF
--- a/changelogs/client_server/newsfragments/1440.feature
+++ b/changelogs/client_server/newsfragments/1440.feature
@@ -1,0 +1,1 @@
+Changes to the server-side aggregation of `m.replace` (edit) events, per [MSC3925](https://github.com/matrix-org/matrix-spec-proposals/pull/3925).
--- a/content/client-server-api/modules/event_replacements.md
+++ b/content/client-server-api/modules/event_replacements.md
@@ -133,13 +133,6 @@ being overwritten entirely by `m.new_content`, with the exception of `m.relates_
 which is left *unchanged*. Any `m.relates_to` property within `m.new_content`
 is ignored.
 
-{{% boxes/note %}}
-Note that server implementations must not *actually* overwrite
-the original event's `content`: instead the server presents it as being overwritten
-when it is served over the client-server API. See [Server-side replacement of content](#server-side-replacement-of-content)
-below.
-{{% /boxes/note %}}
-
 For example, given a pair of events:
 
 ```json
@@ -199,10 +192,11 @@ Note that there can be multiple events with an `m.replace` relationship to a
 given event (for example, if an event is edited multiple times). These should
 be [aggregated](#aggregations) by the homeserver.
 
-The aggregation format of `m.replace` relationships gives the `event_id`,
-`origin_server_ts`, and `sender` of the **most recent** replacement event. The
-most recent event is determined by comparing `origin_server_ts`; if two or more
-replacement events have identical `origin_server_ts`, the event with the
+The aggregation format of `m.replace` relationships gives the **most recent**
+replacement event, formatted as normal for the client-server API (see [Room Event Format](#room-event-format)).
+
+The most recent event is determined by comparing `origin_server_ts`; if two or
+more replacement events have identical `origin_server_ts`, the event with the
 lexicographically largest `event_id` is treated as more recent.
 
 This aggregation is bundled under the `unsigned` property as `m.relations` for any
@@ -211,16 +205,35 @@ event that is the target of an `m.replace` relationship. For example:
 ```json
 {
   "event_id": "$original_event_id",
-  // irrelevant fields not shown
+  "type": "m.room.message",
+  "content": {
+    "body": "I really like cake",
+    "msgtype": "m.text",
+    "formatted_body": "I really like cake"
+  },
   "unsigned": {
     "m.relations": {
       "m.replace": {
         "event_id": "$latest_edit_event_id",
         "origin_server_ts": 1649772304313,
         "sender": "@editing_user:localhost"
+        "type": "m.room.message",
+        "content": {
+          "body": "* I really like *chocolate* cake",
+          "msgtype": "m.text",
+          "m.new_content": {
+            "body": "I really like *chocolate* cake",
+            "msgtype": "m.text"
+          },
+          "m.relates_to": {
+            "rel_type": "m.replace",
+            "event_id": "$original_event_id"
+          }
+        }
       }
     }
   }
+  // irrelevant fields not shown
 }
 ```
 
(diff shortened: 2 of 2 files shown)


content/client-server-api/modules/event_replacements.md
```diff
@@ -231,15 +244,14 @@ subsequent replacements are themselves redacted). Note that this behaviour is
...
+`event_id`, `origin_server_ts`, and `sender` of the replacement event were included
+in the aggregation. v1.7 extends the aggregation to the full event.
 
-An exception applies to [`GET /_matrix/client/v3/rooms/{roomId}/event/{eventId}`](#get_matrixclientv3roomsroomideventeventid),
-which should return the unmodified event (though the relationship should still
-be bundled, as described above).
+{{< changed-in v="1.7" >}} In previous versions of the specification, servers
+were expected to replace the content of the original event with that of the
+replacement event. However that behaviour made reliable client-side
+implementation difficult, and servers should no longer make this replacement.
```
turt2live, line 254: As far as spec language is concerned, this should be something like:

```
{{< changed-in v="1.7" >}} Servers MUST NOT return the most recent edit from [/event].

{{ boxes/rationale }}
In previous versions...
{{ /boxes/rationale }}
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -231,15 +244,14 @@ subsequent replacements are themselves redacted). Note that this behaviour is
...
 events](#redactions-of-edited-events) below.
 
-##### Server-side replacement of content
-
-Whenever an `m.replace` is to be bundled with an event as above, the server
-should also modify the content of the original event according to the
-`m.new_content` of the most recent replacement event (determined as above).
+{{< changed-in v="1.7" >}} In previous versions of the specification, only the
+`event_id`, `origin_server_ts`, and `sender` of the replacement event were included
+in the aggregation. v1.7 extends the aggregation to the full event.
```
turt2live, line 249: Like the comment I've left on the paragraph below, the spec doesn't prefer to have inline changelogs like this (shouldn't say "In previous versions" if at all possible). Instead, this sort of text requirement can get rolled into a much earlier paragraph as a rule for implementations to follow - I'll leave a comment demonstrating that.

content/client-server-api/modules/event_replacements.md
```diff
@@ -211,16 +205,35 @@ event that is the target of an `m.replace` relationship. For example:
 ```json
 {
   "event_id": "$original_event_id",
-  // irrelevant fields not shown
+  "type": "m.room.message",
```
turt2live, line 208: fwiw, I'm still very sad that https://github.com/matrix-org/matrix-spec/issues/688 isn't a thing, which would ideally make this sort of example easy to update in the future.

Non-blocking for this review.

content/client-server-api/modules/event_replacements.md
```diff
@@ -199,10 +192,11 @@ Note that there can be multiple events with an `m.replace` relationship to a
 given event (for example, if an event is edited multiple times). These should
 be [aggregated](#aggregations) by the homeserver.
 
-The aggregation format of `m.replace` relationships gives the `event_id`,
-`origin_server_ts`, and `sender` of the **most recent** replacement event. The
-most recent event is determined by comparing `origin_server_ts`; if two or more
-replacement events have identical `origin_server_ts`, the event with the
+The aggregation format of `m.replace` relationships gives the **most recent**
+replacement event, formatted as normal for the client-server API (see [Room Event Format](#room-event-format)).
```
turt2live, line 196:
```suggestion
replacement event. {{< changed-in v="1.7" >}} That event is formatted as normal for the client-server API (see [Room Event Format](#room-event-format)).
```

arguably we can also use less words because we're already in the CS API:
```suggestion
replacement event. {{< changed-in v="1.7" >}} That event is formatted [as normal](#room-event-format) for clients.
```

content/client-server-api/modules/event_replacements.md
```diff
@@ -133,13 +133,6 @@ being overwritten entirely by `m.new_content`, with the exception of `m.relates_
 which is left *unchanged*. Any `m.relates_to` property within `m.new_content`
 is ignored.
 
-{{% boxes/note %}}
-Note that server implementations must not *actually* overwrite
-the original event's `content`: instead the server presents it as being overwritten
```
turt2live, line 138: We might want to stick the first part of this note somewhere still, just in case servers do super strange things with their events? 

Possibly at the end of the server implementation section with a bit more detail?
```
{{% boxes/note %}}
Note that server implementations must not *actually* overwrite
the original event's `content`. Doing so would cause the original
event to fail signature checks.
{{% /boxes/note %}}

File: changelogs/client_server/newsfragments/1440.feature [hunk 6]
```diff
@@ -0,0 +1 @@
+Changes to the server-side aggregation of `m.replace` (edit) events, per [MSC3925](https://github.com/matrix-org/matrix-spec-proposals/pull/3925).
```
turt2live, line 1: consistency nit:
```suggestion
Changes to the server-side aggregation of `m.replace` (edit) events, as per [MSC3925](https://github.com/matrix-org/matrix-spec-proposals/pull/3925).
```

turt2live (commented):
generally lgtm - mostly nits about documentation style in this one.

content/client-server-api/modules/event_replacements.md
richvdh: This feels like it is more confusing than helpful. As a server implementer, I am led to wonder wtf this "*actually*" means.

I have added some more words to the "server-side aggregation" section which might help here (https://github.com/matrix-org/matrix-spec/pull/1440/commits/97b7544f2c212ef19323c1f835607e3b57331593#diff-7791fd17e89af8337a50aaa7e5fdf6cb1c20013d38cd491cda852e696884be76R248).

changelogs/client_server/newsfragments/1440.feature
richvdh: applied manually

content/client-server-api/modules/event_replacements.md
richvdh: > ```
> {{< changed-in v="1.7" >}} Servers MUST NOT return the most recent edit from [/event].
> ```

this seems to say the opposite to what is intended. 

I've moved the `changed-in` to the top of the section, which seems to be what we've done elsewhere, and moved this explanation into a "rationale" as you suggest.

content/client-server-api/modules/event_replacements.md
richvdh: resolving this thread in favour of https://github.com/matrix-org/matrix-spec/pull/1440#discussion_r1119582258

content/client-server-api/modules/event_replacements.md
richvdh: Now that I've moved the `changed-in` to the top of the section I don't think this extra changed-in is helpful. I've applied the wording changes, with tweaks.

richvdh:
@turt2live PTAL.

I also realised I'd left in some untruths in the "client behaviour section" - cf https://github.com/matrix-org/matrix-spec/pull/1440/commits/05d1779d1fa6e9e56bc32c20e6ca20aeb9dee762

turt2live (approved):
thanks :)

Issue #1401, opened 2023-01-12
Spec doesn't describe how push rules are to be processed in E2E rooms client side


clokep:
**Link to problem area**: https://spec.matrix.org/v1.5/client-server-api/#push-rules

**Issue**
What is wrong? This section, which is under "Client behaviour", seems to describe *Server* behavior (see #600), and fails to describe what clients are supposed to do with this information.

In addition, some of the push rules conditions talk about "unencrypted" events, I don't think this is accurate -- I think the homeserver runs those rules against the encrypted event, and clients are expected to decrypt the event first and then run the push rules. I'm not as confident about this, however.

**Expected behaviour**
How can the issue be fixed? Links to implementations/documents which prove the spec is wrong are appreciated.

The spec should describe how clients are expected to handle push rules.

[matrix-js-sdk's implementation](https://github.com/matrix-org/matrix-js-sdk/blob/7b5e137ec085a78dde17ccc4352b12cf674c1971/src/pushprocessor.ts#L157)

anoadragon453:
I asked about this and other general push implementation questions [in the Matrix Client Developers matrix room](https://matrix.to/#/!XXSJTvRPInupfUgQVb:matrix.org/$NOirbbX6kd7z1fJcPZgHzywPR7aJ3hF1iySSzVxYinI?via=pixie.town&via=matrix.org&via=element.io) and had some great feedback. Leaving the conclusions here as a reference until we can clean up the spec.

The point of running push rules client side is:

* for clients that don't process push rules, and
* processing notifications in encrypted rooms.

There is agreement among clients that push rules should stop executing after a match is found (https://github.com/matrix-org/matrix-spec/issues/638) (as otherwise they wouldn't work as intended!).

Nheko ([implementation](https://github.com/Nheko-Reborn/mtxclient/blob/master/lib/structs/pushrules.cpp#L558)):
  - decrypts events before running push rules over them
  - processes push rules in the order given by the account data event

mautrix-go ([implementation](https://github.com/mautrix/go/tree/master/pushrules), powers gomuks, beeper server-side evaluation):
  - decrypts events before running push rules over them
  - processes push rules in the order given by the account data event

Hydrogen doesn't evaluate push rules client-side yet.

The spec doesn't currently clarify whether you should run the rules over your own events (turned this into https://github.com/matrix-org/matrix-spec/issues/1404).

Do clients run push rules over encrypted versions of events too?

[...]

PR #1441, opened 2023-02-21
Clarify what key content-specific rules match against.

clokep:
It wasn't clear to me what event property content-specific rules actually match against, I had to check Synapse's implementation:

https://github.com/matrix-org/synapse/blob/8219525b66a2ffa9a9f1ed6e5c716fcc1146469e/rust/src/push/base_rules.rs#L268-L281

https://github.com/matrix-org/synapse/blob/8219525b66a2ffa9a9f1ed6e5c716fcc1146469e/synapse/rest/client/push_rule.py#L228-L233

Preview: https://pr1441--matrix-spec-previews.netlify.app

### DIFF
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -936,7 +936,7 @@ E.164 numbering plan. Note that MSISDNs do not include a leading '+'.
 
 ## Glob-style matching
 
-It is useful to match strings via globbing in some situations, globbing in Matrix
+It is useful to match strings via globbing in some situations. Globbing in Matrix
 uses the following rules:
 
 * The character `*` matches zero or more characters.
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -934,6 +934,14 @@ The `address` is the telephone number represented as a MSISDN (Mobile
 Station International Subscriber Directory Number) as defined by the
 E.164 numbering plan. Note that MSISDNs do not include a leading '+'.
 
+## Glob-style matching
+
+It is useful to match strings via globbing in some situations, globbing in Matrix
+uses the following rules:
+
+* The character `*` matches zero or more characters.
+* `?` matches exactly one character.
+
 ## Security Threat Model
 
 ### Denial of Service
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -155,8 +155,12 @@ The different `kind`s of rule, in the order that they are checked, are:
 1.  **Content-specific rules (`content`).**
     These configure behaviour for (unencrypted) messages that match certain
     patterns. Content rules take one parameter: `pattern`, that gives the
-    glob pattern to match against. This is treated in the same way as
-    `pattern` for `event_match` against a `key` of `content.body`.
+    [glob-style pattern](/appendices#glob-style-matching) to match against.
+    The match is performed case-insensitively, and must match any substring of
+    the `content.body` property which starts and ends at a word boundary. A word
+    boundary is defined as the start or end of the value, or any character not
+    in the sets `[A-Z]`, `[a-z]`, `[0-9]` or `_`.The exact meaning of
+    "case insensitive" is defined by the implementation of the homeserver.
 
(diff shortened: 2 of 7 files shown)


content/client-server-api/modules/push.md
```diff
@@ -156,7 +156,7 @@ The different `kind`s of rule, in the order that they are checked, are:
     These configure behaviour for (unencrypted) messages that match certain
     patterns. Content rules take one parameter: `pattern`, that gives the
     glob pattern to match against. This is treated in the same way as
-    `pattern` for `event_match`.
+    `pattern` for `event_match` against a `key` of `content.body`.
```
richvdh, line 159: I mean it's true (I think), and certainly an improvement over what was there before, but it seems a very complicated way of defining it, particularly given the behaviour of `content.body` is itself defined as a special-case of the more general `event_match` behaviour.

Could we define the behaviour from first principles rather than by reference to `event_match`, including the definition of a word boundary and the behaviour of `content.body` is absent or not a string? Possibly the definition of a `glob` should be factored out to a separate section after "Conditions" that can be linked to where we mention globs.

Let me know what you think: if it's a pain in the butt that you don't have time for then we can land this as-is as an incremental improvement.

richvdh: > Possibly the definition of a "glob-style pattern" should be factored out to a separate section after "Conditions" that can be linked to whenever we mention globs.

or even to an appendix so that (now, or in the future) it can be referred to from https://pr1441--matrix-spec-previews.netlify.app/client-server-api/#events-19 and https://pr1441--matrix-spec-previews.netlify.app/client-server-api/#mroomserver_acl

clokep: I can give it a go. 👍 I didn't realize the globbing was used somewhere else, an appendix sounds reasonable.

content/client-server-api/modules/push.md
```diff
@@ -155,8 +155,12 @@ The different `kind`s of rule, in the order that they are checked, are:
 1.  **Content-specific rules (`content`).**
     These configure behaviour for (unencrypted) messages that match certain
```
clokep, line 156: I wonder if we should remove this "(unencrypted)" in this section -- I think this isn't really true since clients also run push rules? (Although see #1401)

content/client-server-api/modules/push.md
clokep: I've attempted this -- the appendix section is rather short, but it is nice to have the same reference from a bunch of places. Let me know what you think!

content/appendices.md
```diff
@@ -934,6 +934,14 @@ The `address` is the telephone number represented as a MSISDN (Mobile
 Station International Subscriber Directory Number) as defined by the
 E.164 numbering plan. Note that MSISDNs do not include a leading '+'.
 
+## Glob-style matching
+
+It is useful to match strings via globbing in some situations, globbing in Matrix
```
richvdh, line 939: Comma splice!
```suggestion
It is useful to match strings via globbing in some situations. Globbing in Matrix
```

content/client-server-api/modules/push.md
richvdh: Indeed.

richvdh (approved):
looks great otherwise.

content/client-server-api/modules/push.md
clokep: Is that a "Yes, remove this please"? 😄

clokep: There seems to be a few more instances of it, I think it'll be cleaner to do it in a follow-up (some of them are quite unrelated to this PR).

clokep: I started doing this in #1461.

clokep:
To reiterate -- I think this is ready, I'd like to do some more clean-up here (and kind of have to for MSC3758, MSC3873, and MSC3966) but will do it in follow-ups.

richvdh:
Thanks!

PR #1444, opened 2023-02-21
Load Inter CSS directly

richvdh:
Move the load of the Inter font CSS from `_variables_project.scss` to `head-end.html`.

Empirically, this seems to fix https://github.com/matrix-org/matrix-spec/issues/965


richvdh:
I'm still quite tempted just to rip out everything Inter-related and use the system font, but let's see if this works first.

Issue #1358, opened 2022-11-23
/createRoom styling is borked


DMRobertson:
**Link to problem area**:  https://spec.matrix.org/v1.5/client-server-api/#post_matrixclientv3createroom

**Issue**
![image](https://user-images.githubusercontent.com/8614563/203576411-75394599-81a4-48f1-b688-3744949bc611.png)

DMRobertson:
Regressed in 1.4; fine in 1.3.

DMRobertson:
Two oddities:
- the text of each list item is inside a `<p>`, which seems like overkill
- the list items inherit the following style which seems to bork the rendering:
  ```css
   summary, slot[name="internal-main-summary"]::slotted(summary) {
     list-style: disclosure-closed inside;
   }
   ```

DMRobertson:
Oh I see: all of the guff at the top is inside the <summary>, and the inheritance of styles is meant to give it a more compact display.

If it were up to me I'd move everything except the HTTP Method and Path outside of the details. (It's frequently irritating that collapsing an endpoint you don't care about still leaves you with a big chunk of description that you still don't care about.)

turt2live:
related: https://github.com/matrix-org/matrix-spec/issues/1352

richvdh:
also [`POST /_matrix/client/v3/rooms/{roomId}/invite`](https://spec.matrix.org/v1.5/client-server-api/#third-party-invites)

DMRobertson:
And also https://spec.matrix.org/v1.6/client-server-api/#mroommember

DMRobertson:
Pretty sure this is a duplicate of #1352.

PR #1446, opened 2023-02-22
Only include endpoint path in `<summary>`

### LINKED ISSUE #1352: Borked styling of list items under /register definition
**Link to problem area**: https://spec.matrix.org/v1.4/client-server-api/#post_matrixclientv3register

**Issue**
image

Was fine in 1.3, borked in 1.4 and 1.5.

DMRobertson:
As threatened on https://github.com/matrix-org/matrix-spec/issues/1358#issuecomment-1325220859 and https://github.com/matrix-org/matrix-spec/issues/1352#issuecomment-1439033334. (If nothing else, I seem to be consistent.)

Fixes #1352.

Preview: https://pr1446--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -288,15 +288,19 @@ footer {
 /* Styles for sections that are rendered from data, such as HTTP APIs and event schemas */
 .rendered-data {
   background-color: $secondary-lightest-background;
-  padding: 1rem;
-  margin: 1rem 0;
+  padding: 0.85rem;
+  margin: 0.85rem 0;
 
   details {
     summary {
-      padding: .5rem 0;
-      p {
-        max-width: 80%;
+      h1 {
+        margin: 0;
+        /* Ensure the disclosure control is vertically centred with the header text. */
+        vertical-align: middle;
       }
+    }
+    p {
+      max-width: 80%;
     }
   }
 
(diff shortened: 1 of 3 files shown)


DMRobertson:
Seems to fix #1352. I can't remember the details but I vaguely remember from the last time I looked into this that there's something about being inside a `<summary>` that changes how lists are laid out.

DMRobertson:
> I do seem to have lost the borders

This seems to be local to me and was because I had uncommitted changes in themes/docsy.

assets/scss/custom.scss
```diff
@@ -288,16 +288,20 @@ footer {
...
   details {
     summary {
-      padding: .5rem 0;
-      p {
-        max-width: 80%;
+      h1 {
+        margin: 0;
+        /* Ensure the disclosure control is vertically centred with the header text. */
+        vertical-align: middle;
       }
```
richvdh, line 300: I guess this relies on the first line of the summary being an `<h1>`, but that's probably an assumption we can live with.

richvdh (approved):
thanks!

assets/scss/custom.scss
DMRobertson: Could maybe do `summary > *` or `summary > :first-child` if this bites us in the future/

DMRobertson:
Other details elements have this wierdly oversized summary too, e.g https://spec.matrix.org/v1.7/client-server-api/#mroomcanonical_alias

PR #1447, opened 2023-02-22
Add a hyphen between third and party when used as an adjective

anoadragon453:
'third-party' should have a hyphen when used as an adjective, such as in 'third-party network'. It should not have a dash when used as a noun, such as in 'sharing with a third party'.

All uses in the spec of third-party were as an adjective. Thus they have all been updated to include a hyphen where they didn't already.

Additionally, any instance of a capital "Third-Party" has had the word "Party" lowercased. This is mostly personal preference. We use both second-word capitalised and non-capitalised throughout the second (see e-mail, well-known, single sign-on, etc.)

Preview: https://pr1447--matrix-spec-previews.netlify.app

### DIFF
--- a/attic/drafts/model/rooms.rst
+++ b/attic/drafts/model/rooms.rst
@@ -246,7 +246,7 @@ anyway.
 In this arrangement there is now a room with both users may join but neither has
 the power to invite any others. Both users now have the confidence that (at
 least within the messaging system itself) their messages remain private and
-cannot later be provably leaked to a third party. They can freely set the topic
+cannot later be provably leaked to a third-party. They can freely set the topic
 or name if they choose and add or edit any other state of the room. The update
 powerlevel of each of these fixed properties should be 1, to lock out the users
 from being able to alter them.
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -1115,7 +1115,7 @@ the homeserver using the [`/account/3pid`](#get_matrixclientv3account3pid) API r
 {
   "type": "m.login.password",
   "identifier": {
-    "medium": "<The medium of the third-party-identifier>",
+    "medium": "<The medium of the third-party identifier>",
     "address": "<The canonicalised third-party address of the user>"
   },
   "password": "<password>"
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -150,15 +150,15 @@ Sent when a threepid given to an API cannot be used because no record
 matching the threepid was found.
 
 `M_THREEPID_AUTH_FAILED`
-Authentication could not be performed on the third party identifier.
+Authentication could not be performed on the third-party identifier.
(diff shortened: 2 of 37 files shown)


attic/drafts/model/rooms.rst
```diff
@@ -246,7 +246,7 @@ anyway.
 In this arrangement there is now a room with both users may join but neither has
 the power to invite any others. Both users now have the confidence that (at
 least within the messaging system itself) their messages remain private and
-cannot later be provably leaked to a third party. They can freely set the topic
+cannot later be provably leaked to a third-party. They can freely set the topic
```
KitsuneRal, line 249: This is not an adjective?

content/client-server-api/modules/third_party_networks.md
```diff
@@ -1,18 +1,18 @@
...
 communication platforms, with messages ferried back and forth by the
 application service. A single application service may bridge multiple
-third party networks, and many individual locations within those
-networks. A single third party network location may be bridged to
+third-party networks, and many individual locations within those
+networks. A single third-party network location may be bridged to
 multiple Matrix rooms.
 
-#### Third Party Lookups
+#### Third-party Lookups
```
KitsuneRal, line 12: I'm not sure "lookup" should be qualified this way. To me, "third-party lookup" is a lookup done by a third party; without a hyphen, it's a lookup of that third party.

KitsuneRal (commented):
I'm not a native speaker but I think it goes a bit too far in a couple of caces.

content/client-server-api/modules/third_party_networks.md
anoadragon453: Aren't third-party lookups something that a Matrix user does to figure out what third-party locations (IRC rooms, Slack channels, etc.) exist for a given Matrix room?

This lookup isn't done by the third party - they don't know how to speak Matrix.

content/client-server-api/_index.md
```diff
@@ -1115,8 +1115,8 @@ the homeserver using the [`/account/3pid`](#get_matrixclientv3account3pid) API r
 {
   "type": "m.login.password",
   "identifier": {
-    "medium": "<The medium of the third party identifier>",
-    "address": "<The canonicalised third party address of the user>"
+    "medium": "<The medium of the third-party-identifier>",
```
uhoreg, line 1118: I don't think that second hyphen is supposed to be there

```suggestion
    "medium": "<The medium of the third-party identifier>",
```

content/client-server-api/modules/third_party_networks.md
uhoreg: The hyphenated version is correct here.  It is ambiguous, but hopefully context will make it clear.  If you want to blame it on English being a terrible language, then yes, it is.

KitsuneRal (approved):

PR #1454, opened 2023-02-26
Remove overridden references to `examples/minimal_pdu.json` and add the missing `room_id` field.

zecakeh:
From https://github.com/matrix-org/matrix-spec/pull/1310#discussion_r1070549892.

The syntax is incorrect and would require to use `allOf` because a `$ref` can't have siblings.
However the only field not overwritten of that definition is `room_id`, so we include it instead of the `$ref`.

Pre-requisite for #1310.

Preview: https://pr1454--matrix-spec-previews.netlify.app


KitsuneRal (approved):

richvdh (approved):

changelogs/server_server/newsfragments/1454.clarification
```diff
@@ -0,0 +1 @@
+Remove overridden references to `examples/minimal_pdu.json` and add the missing `room_id` field.
```
richvdh, line 1:
```suggestion
Fix invalid OpenAPI specifications caused by overridden references to `examples/minimal_pdu.json`.
```

PR #1456, opened 2023-02-28
Minor updates to documentation style

richvdh:
i.e. has full-stops

Preview: https://pr1456--matrix-spec-previews.netlify.app

### DIFF
--- a/meta/documentation_style.rst
+++ b/meta/documentation_style.rst
@@ -44,33 +44,39 @@ Stylistic notes
 General
 ~~~~~~~
 
-Try to write clearly and unambiguously. Remember that many readers will not
-have English as their first language.
+* Try to write clearly and unambiguously. Remember that many readers will not
+  have English as their first language.
 
-Prefer British English (colour, -ise) to American English.
+* Prefer British English (colour, -ise) to American English.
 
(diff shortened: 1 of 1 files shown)


meta/documentation_style.rst
```diff
@@ -44,33 +44,39 @@ Stylistic notes
...
+  "member", or various other alternatives. For example: "this property will be set to
+  X if ...". Also avoid the term "key" unless you are specifically talking about the
+  *name* of a property - and be mindful of the scope for confusion with cryptographic
+  keys.
+
+* "i.e." (*id est*) is an abbreviation and hence is written with two full
+  stops. Prefer full sentences where possible without loss of clarity.
+
+* Prefer "for example" over "e.g.". If you *must* use "e.g.", remember it too
+  is an abbreviation (*exempli gratia*).
```
richvdh, line 79: Carrying over some comments from https://github.com/matrix-org/matrix-spec/pull/1424#discussion_r1119573812:

@turt2live says:

> ... I'm not currently convinced it's a thing we [should] do - given the first point is that the language should be unambiguous, shouldn't we be using complete words/sentences instead? ("For example", "As demonstrated below", "..., such as when", etc)

turt2live (approved):
thanks - this feels like a reasonable compromise given our usage.

PR #1461, opened 2023-03-07
Define how clients process push rules

### LINKED ISSUE #1401: Spec doesn't describe how push rules are to be processed in E2E rooms client side

clokep:
Fixes #1401: I'm not 100% sure this is accurate, but I think it is? It doesn't answer all the questions in that issue, but gets a lot closer.

~~Based on #1441.~~

Preview: https://pr1461--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -833,6 +833,13 @@ Servers MUST include the number of unread notifications in a client's
 `/sync` stream, and MUST update it as it changes. Notifications are
 determined by the push rules which apply to an event.
 
+For encrypted events, the homeserver has limited access to the event content
+and properly processing push rules falls on the client. Clients should process
+push rules for each incoming event *after decrypting* them. This may result in
+needing to modify the number of unread notifications received from the homeserver.
+
+##### Marking notifications as read
+
 When the user updates their read receipt (either by using the API or by
 sending an event), notifications prior to and including that event MUST
 be marked as read. Which specific events are affected can vary depending
@@ -863,6 +870,15 @@ to ensure that future events, like `m.reaction`, are correctly considered
 "part of" a given thread.
 
 #### Server behaviour
+
+When receiving a new event homeservers process push rules for all other local
+users in the room, this may result in:
+
+* Generating a new number of unread notifications for the user;
+* Making a request to the configured push gateway.
+
+The updated notification count from a new event MUST appear in the same `/sync`
+response as the event itself.
 
 #### Push Gateway behaviour
 
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -296,7 +296,7 @@ For an example of this, see the default rule
 **`contains_display_name`**
 
 This matches messages where `content.body` contains the owner's display name in
-that room. This is a separate rule because display names may change and as such
+that room. This is a separate condition because display names may change and as such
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/push.md
```diff
@@ -346,10 +295,10 @@ For an example of this, see the default rule
 
 **`contains_display_name`**
```
clokep, line 296: I wonder if we should make each push rule condition a header now?

content/client-server-api/modules/push.md
```diff
@@ -346,10 +295,10 @@ For an example of this, see the default rule
...
 
-This matches unencrypted messages where `content.body` contains the
-owner's display name in that room. This is a separate rule because
-display names may change and as such it would be hard to maintain a rule
-that matched the user's display name. This condition has no parameters.
+This matches messages where `content.body` contains the owner's display name in
+that room. This is a separate rule because display names may change and as such
```
richvdh, line 299:
```suggestion
that room. This is a separate condition because display names may change and as such
```

content/client-server-api/modules/push.md
```diff
@@ -495,8 +444,8 @@ Definition:
 
 **`.m.rule.contains_display_name`**
 
-Matches any message whose content is unencrypted and contains the user's
-current display name in the room in which it was sent.
+Matches any message whose content is contains the user's current display name
```
richvdh, line 447:
```suggestion
Matches any message whose content contains the user's current display name
```

content/client-server-api/modules/push.md
```diff
@@ -618,12 +568,12 @@ Definition:
...
 ```
 
-###### Default Content Rules
+##### Default Content Rules
 
 **`.m.rule.contains_user_name`**
 
-Matches any message whose content is unencrypted and contains the local
-part of the user's Matrix ID, separated by word boundaries.
+Matches any message whose content is contains the local part of the user's
```
richvdh, line 575:
```suggestion
Matches any message whose content contains the local part of the user's
```

richvdh (changes requested):
looks good, a couple of nits

content/client-server-api/modules/push.md
clokep: > We could park this for now though.

This sounds good to me, I don't want to bitrot my follow-up anyway. 👍

clokep:
Thanks for the review! I think this is ready for another look! 👍

richvdh (approved):

Issue #200, opened 2016-09-11
Spec the implicit limit on /syncs (SPEC-453)


matrixbot:
If you call `/sync` without a filter, or with a filter without a limit, it will use a default limit. We should spec this.

(Imported from https://matrix.org/jira/browse/SPEC-453)

richvdh:
fixed by #1463

PR #1463, opened 2023-03-14
Spec implicit filter event limit

Cadair:
This is an attempt to address #200 

I think I am understanding the issue correctly. ~~I took the default out of [synapse](https://github.com/matrix-org/synapse/blob/a1c986939444db9bbf2343224091c88d18b3f598/synapse/api/filtering.py#L336) I haven't looked in any other implementations to see if they also use the default of 10.~~

Preview: https://pr1463--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/event_filter.yaml
+++ b/data/api/client-server/definitions/event_filter.yaml
@@ -14,7 +14,7 @@
 title: EventFilter
 properties:
   limit:
-    description: The maximum number of events to return.
+    description: The maximum number of events to return. If not specified defaults to 10.
     type: integer
   not_senders:
     description: A list of sender IDs to exclude. If this list is absent then no senders
--- a/data/api/client-server/definitions/event_filter.yaml
+++ b/data/api/client-server/definitions/event_filter.yaml
@@ -14,7 +14,11 @@
 title: EventFilter
 properties:
   limit:
-    description: The maximum number of events to return. If not specified defaults to 10.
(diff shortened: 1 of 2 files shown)


data/api/client-server/definitions/event_filter.yaml
```diff
@@ -14,7 +14,7 @@
 title: EventFilter
 properties:
   limit:
-    description: The maximum number of events to return.
+    description: The maximum number of events to return. If not specified defaults to 10.
```
richvdh, line 17: I'm not entirely sure we should mandate this, or that we *can* mandate it, without going through the MSC process: it may well be that non-synapse implementations use a different default.

I think what we need to say instead is that if it's not specified, servers should use an implementation-specific default.

The `limit` param on [`GET /threads`](https://spec.matrix.org/v1.6/client-server-api/#get_matrixclientv1roomsroomidthreads) has sensible wording here.

richvdh (changes requested):
@Cadair thanks for picking this up!

richvdh:
Also: please could you take a look at https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst, particularly with respect to the need for a [changelog file](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog)?

richvdh (approved):
Great, thank you!

PR #1464, opened 2023-03-14
Document MSC3758, MSC3966, MSC3873, and MSC3980.

clokep:
This documents the new push rule conditions:

* `event_property_is` ([MSC3758](https://github.com/matrix-org/matrix-spec-proposals/pull/3758))
* `event_property_contains` ([MSC3966](https://github.com/matrix-org/matrix-spec-proposals/pull/3966))

As well as the dotted-path escape rules from [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873) and [MSC3980](https://github.com/matrix-org/matrix-spec-proposals/pull/3980).

~~Based on #1461.~~

Preview: https://pr1464--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1464.clarification
+++ b/changelogs/client_server/newsfragments/1464.clarification
@@ -1,0 +1,1 @@
+Disambiguate using property names with dots in them during push rule processiong, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
--- a/changelogs/client_server/newsfragments/1464.clarification
+++ b/changelogs/client_server/newsfragments/1464.clarification
@@ -1,1 +1,1 @@
-Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
+Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873) and [MSC3980](https://github.com/matrix-org/matrix-spec-proposals/pull/3980).
--- a/changelogs/client_server/newsfragments/1464.clarification
+++ b/changelogs/client_server/newsfragments/1464.clarification
@@ -1,1 +1,1 @@
-Disambiguate using property names with dots in them during push rule processiong, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
+Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -951,10 +951,12 @@ event's `content`.
 To handle ambiguity when a property name contains a dot, any literal dot or
 backslash found in a property name should be escaped with a backslash. E.g. a
 property `m.relates_to` in the `content` would be expressed as
-`content.m\.relates_to`.
+`content.m\.relates_to`. Similarly, a `content` property named `m\foo` would be
+expressed as `content.m\\foo`.
 
-It is recommended that implementations do not redundantly escape characters, as
-other escape sequences are reserved for future use.
+Other escape sequences are left as-is, e.g. a `\x` would be treated as a literal
+backslash followed by 'x'. It is recommended that implementations do not redundantly
+escape characters, as other escape sequences are reserved for future use.
 
 ## Security Threat Model
 
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -942,6 +942,20 @@ uses the following rules:
 * The character `*` matches zero or more characters.
 * `?` matches exactly one character.
 
+## Dot-separated property paths
+
+It is useful to express the "path" to an event property by concatenating property
+names with dots, e.g. `content.body` would represent the a `body` property in the
+event's `content`.
+
+To handle ambiguity when a property name contains a dot, any literal dot or
+backslash found in a property name should be escaped with a backslash. E.g. a
+property `m.relates_to` in the `content` would be expressed as
+`content.m\.relates_to`.
+
+It is recommended that implementations do not redundantly escape characters, as
+other escape sequences are reserve for future use.
+
 ## Security Threat Model
 
 ### Denial of Service
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -945,7 +945,7 @@ uses the following rules:
 ## Dot-separated property paths
 
 It is useful to express the "path" to an event property by concatenating property
-names with dots, e.g. `content.body` would represent the a `body` property in the
+names with dots, e.g. `content.body` would represent a `body` property in the
 event's `content`.
 
 To handle ambiguity when a property name contains a dot, any literal dot or
@@ -954,7 +954,7 @@ property `m.relates_to` in the `content` would be expressed as
 `content.m\.relates_to`.
 
 It is recommended that implementations do not redundantly escape characters, as
-other escape sequences are reserve for future use.
+other escape sequences are reserved for future use.
 
 ## Security Threat Model
 
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -214,7 +214,7 @@ The following conditions are defined:
 
 This is a glob pattern match on a property of the event. Parameters:
 
--   `key`: The [dot-separated path of the property](/appendices#dot-separated property paths)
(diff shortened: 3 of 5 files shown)


data/api/client-server/definitions/sync_filter.yaml
```diff
@@ -16,10 +16,9 @@ title: Filter
 properties:
   event_fields:
     description: List of event fields to include. If this list is absent then all
-      fields are included. The entries may include '.' characters to indicate sub-fields.
-      So ['content.body'] will include the 'body' field of the 'content' object. A
-      literal '.' character in a field name may be escaped using a '\\'. A server may
-      include more fields than were requested.
```
clokep, line 22: The MSC doesn't *quite* say this since this implicitly adds escaping a backslash with another backslash, which I don't think was previously valid. I can back this out if you would like.

content/client-server-api/modules/push.md
```diff
@@ -293,6 +293,94 @@ For an example of this, see the default rule
...
+
+The match is performed exactly and only supports strings, integers, booleans, and
+`null`.
+
+If the property specified by `key` is completely absent from the event, or does
+not have a string, integer, boolean, or `null` value, then the condition will not
+match.
+
+{{% boxes/note %}}
+For example, if `key` is `content.m.federate`, and `value` is `true`, then
```
clokep, line 313: This isn't escaped properly:

```suggestion
For example, if `key` is `content.m\\.federate`, and `value` is `true`, then
```

I wanted to use an example of a real field -- if there are better ones than this, please let me know of ideas!

content/client-server-api/modules/push.md
```diff
@@ -293,6 +293,50 @@ For an example of this, see the default rule
...
+
+This is an exact value match on a property of the event. Parameters:
+
+-   `key`: The dot-separated path of the property of the event to match, e.g.
+    `content.body`.
+
+-   `value`: The value to match against.
+
+The match is performed exactly and only supports strings, integers, booleans, and
+`null`.
```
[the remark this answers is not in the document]
anoadragon453: Likewise for `event_property_contains`.

content/appendices.md
```diff
@@ -942,6 +942,20 @@ uses the following rules:
 * The character `*` matches zero or more characters.
 * `?` matches exactly one character.
 
+## Dot-separated property paths
+
+It is useful to express the "path" to an event property by concatenating property
+names with dots, e.g. `content.body` would represent the a `body` property in the
```
anoadragon453, line 948:
```suggestion
names with dots, e.g. `content.body` would represent a `body` property in the
```

content/appendices.md
```diff
@@ -942,6 +942,20 @@ uses the following rules:
...
+event's `content`.
+
+To handle ambiguity when a property name contains a dot, any literal dot or
+backslash found in a property name should be escaped with a backslash. E.g. a
+property `m.relates_to` in the `content` would be expressed as
+`content.m\.relates_to`.
+
+It is recommended that implementations do not redundantly escape characters, as
+other escape sequences are reserve for future use.
```
anoadragon453, line 957:
```suggestion
other escape sequences are reserved for future use.
```

content/appendices.md
```diff
@@ -942,6 +942,20 @@ uses the following rules:
...
+## Dot-separated property paths
+
+It is useful to express the "path" to an event property by concatenating property
+names with dots, e.g. `content.body` would represent the a `body` property in the
+event's `content`.
+
+To handle ambiguity when a property name contains a dot, any literal dot or
+backslash found in a property name should be escaped with a backslash. E.g. a
+property `m.relates_to` in the `content` would be expressed as
+`content.m\.relates_to`.
```
anoadragon453, line 954: I think it would help to also include an example of escaping a backslash as well (for a theoretical field, as none today contain backslashes).

changelogs/client_server/newsfragments/1464.clarification
```diff
@@ -0,0 +1 @@
+Disambiguate using property names with dots in them during push rule processiong, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
```
anoadragon453, line 1:
```suggestion
Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
```

data/api/client-server/definitions/sync_filter.yaml
```diff
@@ -16,10 +16,9 @@ title: Filter
...
+      fields are included. The entries are [dot-separated paths for each property](/appendices#dot-separated property paths)
```
anoadragon453, line 19: This markdown link doesn't appear to work. I think you might need to include some more hyphens?

```suggestion
      fields are included. The entries are [dot-separated paths for each property](/appendices#dot-separated-property-paths)
```

Likewise for the other places the Appendices is referenced in this PR.

data/api/client-server/definitions/sync_filter.yaml
anoadragon453: Hmm, since MSC3873 doesn't propose modifying sync filters to support escaping backslashes (instead just taking the dot-escaping capability of sync filter), I believe the *correct* approach to unify the two would be to have another MSC about Sync filters to include support for escaping backslashes.

I think such an MSC would be met without any contention, as everyone just wants consistency at the end of the day.

However, while that separate MSC is in review and MSC3873 is merged, it does mean that clients/servers will have to match backslashes differently depending on whether they're processing push rules or sync filters, which is a bit of a pain. So this may be a case of the process getting in the way, and we simply edit MSC3873 after the fact to mention that sync filters should also escape backslashes.

Finally, this is all a bit moot as no fields in events currently contain backslashes anyway (except possibly some displaynames?).

What do other @matrix-org/spec-core-team members think?

anoadragon453 (changes requested):
This looks great, and I appreciate your clear wording.

A few bits and bobs to fix up, and a leading question.

content/client-server-api/modules/push.md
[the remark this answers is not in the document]
clokep: The MSC discussed "non-compound values from Canonical JSON" (or something similar to that), is that phrasing useful or just list them out?

content/client-server-api/modules/push.md
clokep: > Why don't we escape it properly here, since we're also introducing [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/issues/3873)

Because I forgot to do it in the PR so was making a note to myself -- I was curious if a reviewer had a better property to check against.

data/api/client-server/definitions/sync_filter.yaml
clokep: > Finally, this is all a bit moot as no fields in events currently contain backslashes anyway (except possibly some displaynames?).

It only applies to property *names*, not values.

data/api/client-server/definitions/sync_filter.yaml
clokep: Oops. 👍

data/api/client-server/definitions/sync_filter.yaml
clokep: I decided to just write the MSC: see [MSC3980](https://github.com/matrix-org/matrix-spec-proposals/pull/3980).

anoadragon453: I'll try to push for review of the MSC on Tuesday with the rest of the Spec Core Team as it should be universally agreeable and a very short read. Thank you for writing it :slightly_smiling_face:

changelogs/client_server/newsfragments/1464.clarification
```diff
@@ -0,0 +1 @@
+Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873).
```
clokep, line 1:
```suggestion
Disambiguate using property names with dots in them during push rule processing, per [MSC3873](https://github.com/matrix-org/matrix-spec-proposals/pull/3873) and [MSC3980](https://github.com/matrix-org/matrix-spec-proposals/pull/3980).
```

data/api/client-server/definitions/sync_filter.yaml
clokep: @anoadragon453 Re-requested review, although this is blocked on the FCP timer of MSC3980.

clokep: This should be ready to go now. 👍

PR #1465, opened 2023-03-14
Correct a small typo in the push rules spec (#1465

anoadragon453:
This PR fixes a field name in an example that does not make sense.

The example does not talk about a "membership" field in any case. It looks like this was instead supposed to be "topic", as the previous paragraph also talks about.

Preview: https://pr1465--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1465.clarifications
+++ b/changelogs/client_server/newsfragments/1465.clarifications
@@ -1,0 +1,1 @@
+Fix various typos throughout the specification.
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -249,7 +249,7 @@ Other `topic` values which will match are:
 
  * `"LUNCH"` (case-insensitive; `*` may match zero characters)
 
-The following `membership` values will NOT match:
+The following `topic` values will NOT match:
  * `" lunch"` (note leading space)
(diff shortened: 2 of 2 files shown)


clokep (approved):
Oh I noticed in this #1464 and failed to update it. 😢

anoadragon453:
> Oh I noticed in this #1464 and failed to update it. cry

Your PR made me notice it, so in the end it still helped fixed it :sparkles: 

I'm going to Patrick's review as an approval of this typo fix.

turt2live:
(ftr, force-merge bypassed the changelog check: `.clarification` is not plural. Will fix out of band)

anoadragon453:
@turt2live I didn't perform a force merge, and the CI passed: https://github.com/matrix-org/matrix-spec/actions/runs/4417512313/jobs/7743320855#step:3:7

Though it looks like the CI only checks that the changelog file is in a valid directory, and has the right number. It does not check that it has a valid extension: https://github.com/matrix-org/matrix-spec/blob/db0fd307447edeb5724c9897e07e6c56d87123f6/scripts/check-newsfragments#L8

turt2live:
I swear yesterday it was saying the ci partially passed :(

Sorry, and good spot: we should fix that workflow.

PR #1466, opened 2023-03-14
Clarify that Persistent data unit is PDU

### LINKED ISSUE #360: persistent data unit or persisted data unit?
we have both in the s2s spec

Cadair:
Should close #360, this was the only occurrence of "Persisted Data Unit" I could find with grep.

Preview: https://pr1466--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/1466.clarification
+++ b/changelogs/server_server/newsfragments/1466.clarification
@@ -1,1 +1,1 @@
-Make sure that PDU is referred to as "Persistent Data Unit" everywhere.
+Fix various typos throughout the specification.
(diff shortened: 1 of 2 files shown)


changelogs/server_server/newsfragments/1466.clarification
```diff
@@ -0,0 +1 @@
+Make sure that PDU is referred to as "Persistent Data Unit" everywhere.
```
richvdh, line 1: for these sorts of really trivial fixes we normally just do:

```suggestion
Fix various typos throughout the specification.
```

richvdh (approved):
Thanks!

PR #1475, opened 2023-03-28
Spec annotations

richvdh:
Spec for MSC2677

~~Includes #1476~~

Preview: https://pr1475--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/event_annotations.md
+++ b/content/client-server-api/modules/event_annotations.md
@@ -58,7 +58,8 @@ When performing this count:
  * Implementations should ignore any annotation event which refers to an event
    which itself has an `m.relates_to` with `rel_type: m.annotation` or
    `rel_type: m.replace`. In other words, it is not possible to annotate a
-   [replacement event](#event-replacements) or an annotation.
+   [replacement event](#event-replacements) or an annotation.  Annotations should
+   instead refer to the original event.
 
  * When an annotation is redacted, it is removed from the count.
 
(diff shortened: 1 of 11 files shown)


content/client-server-api/modules/event_annotations.md
```diff
@@ -0,0 +1,92 @@
...
+#### Server behaviour
+
+##### Avoiding duplicate annotations
+
+Homeservers should prevent users from sending a second annotation for a given
+event with identical event `type` and annotation `key` (unless the first event
+has been redacted).
+
+Attempts to send such an annotation should be rejected with a 400 error and an
+error code of `M_DUPLICATE_ANNOTATION`.
```
richvdh, line 82: wonder if I should add something about this to the docs on `/send`

uhoreg: Yes, I think so.  That endpoint seems to need more documentation about errors in general, but starting with this would be a good step at least.

content/client-server-api/modules/event_annotations.md
```diff
@@ -0,0 +1,92 @@
...
+   excluded from the count.
+
+ * Multiple identical annotations (i.e., with the same event `type` and
+   annotation `key`) from the same user (i.e., events with the same `sender`)
+   should be treated as a single annotation.
+
+ * Implementations should ignore any annotation event which refers to an event
+   which itself has an `m.relates_to` with `rel_type: m.annotation` or
+   `rel_type: m.replace`. In other words, it is not possible to annotate a
+   [replacement event](#event-replacements) or an annotation.
```
uhoreg, line 61: Maybe add something like:

```suggestion
   [replacement event](#event-replacements) or an annotation.  Annotations should
   instead refer to the original event.
```

?

uhoreg (approved):

PR #1478, opened 2023-03-30
Mention that `state_default` can be zero by default.

turt2live:
Preview: https://pr1478--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/schema/m.room.power_levels.yaml
+++ b/data/event-schemas/schema/m.room.power_levels.yaml
@@ -76,7 +76,7 @@ properties:
             The default level required to send state events. Can be overridden
             by the `events` key. Defaults to 50 if unspecified.
 
-            **Note**: When there is no power levels event in the room, this defaults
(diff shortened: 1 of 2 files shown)


data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -75,6 +75,9 @@ properties:
         description: |-
             The default level required to send state events. Can be overridden
             by the `events` key. Defaults to 50 if unspecified.
+
+            **Note**: When there is no power levels event in the room, this defaults
```
uhoreg, line 79:
```suggestion
            **Note**: When there is no `m.room.power_levels` event in the room, this defaults
```

My nitpicky comment: when we talk about the event, we tend to use the full name.  There does seem to be some places where we call it the "power level event", but it looks like most of the places use the full "`m.room.power_levels` event", so I think we should stick to that for consistency.

uhoreg (approved):

PR #1479, opened 2023-03-31
Remove outdated text saying that state_default defaults to 0

richvdh:
There was substantial confusion around this, but I've done some archaeology.

Basically, this was changed back in r0.5.0 by MSC1304 and https://github.com/matrix-org/synapse/pull/3397. Before that, it was indeed the case that `state_default` was 0 if there was no `m.room.power_levels` event, but that was confusing and a source of security holes, so we changed it.

https://github.com/matrix-org/matrix-doc/pull/1656 changed the spec, but apparently overlooked the text in the description.

Reverts: #1478.
Fixes: #861.

Preview: https://pr1479--matrix-spec-previews.netlify.app


data/event-schemas/schema/m.room.power_levels.yaml
```diff
@@ -75,9 +65,6 @@ properties:
         description: |-
             The default level required to send state events. Can be overridden
             by the `events` key. Defaults to 50 if unspecified.
-
-            **Note**: When there is no `m.room.power_levels` event in the room, this defaults
-            to 0. See description for more information.
```
turt2live, line 80: I probably should have done it in my original PR, but while you're here it'd be good to have a similar note for the "creator gets 100" clause on `users_default` (effectively).

turt2live (approved):
lgtm either way though

data/event-schemas/schema/m.room.power_levels.yaml
richvdh: Added a note in 4a90545e3da7084415710e08aeb676c9191af1cb (and clarified the rest of the description of `users_default` too

PR #1488, opened 2023-04-06
Fix generation of anchors for additional properties

richvdh:
\#1174 added HTML anchors for object definitions, but objects defined under `additionalProperties` missed out on them. This fixes that.

Preview: https://pr1488--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1488.clarification
+++ b/changelogs/internal/newsfragments/1488.clarification
@@ -1,0 +1,1 @@
+Fix generation of anchors for additional properties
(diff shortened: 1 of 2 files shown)


uhoreg:
I don't understand how the change fixes the issue.  I'm inclined to believe that it does, but could you provide a link to a place in the spec where we can see where the change makes a difference?

richvdh:
Sorry yes, this is a bit obscure. Here's one example: https://pr1488--matrix-spec-previews.netlify.app/client-server-api/#_matrixclientv3roomsroomidjoined_members_roommember.

uhoreg (approved):
Thanks.  I will take that as evidence that the PR does what it is supposed to do.

PR #1495, opened 2023-04-18
Clarify signatures upload endpoint

uhoreg:
Preview: https://pr1495--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1495.clarification
+++ b/changelogs/client_server/newsfragments/1495.clarification
@@ -1,0 +1,1 @@
+Clarify parts of the end-to-end encryption sections.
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
This text change looks great, thank you! While you're here, could you also fix the Request body parameters section just below? Looking at https://pr1495--matrix-spec-previews.netlify.app/client-server-api/#post_matrixclientv3keyssignaturesupload, I currently see:

image

The contents under "Request" -> "Request body" are missing.

richvdh:
> The contents under "Request" -> "Request body" are missing.

I think that's somewhat expected, and a result of the fact that the top-level key is a user ID rather than a fixed string.

https://github.com/matrix-org/matrix-spec/pull/1162 is related, but sadly stuck.

I think we should merge this as-is and consider the poor formatting of requests like this a separate problem.

PR #1501, opened 2023-04-25
Remove the `dont_notify` and `coalesce` push rule actions.

clokep:
Per MSC3987, these should both be considered no-ops.

### DIFF
--- a/content/client-server-api/modules/push.md
+++ b/content/client-server-api/modules/push.md
@@ -147,17 +147,6 @@ actions are defined:
 
 :   This causes each matching event to generate a notification.
 
-`dont_notify`
-
-:   This prevents each matching event from generating a notification.
-
-`coalesce`
-
-:   This enables notifications for matching events but activates homeserver
-    specific behaviour to intelligently coalesce multiple events into a
-    single notification. Not all homeservers may support this. Those that do
-    not support it should treat it as the `notify` action.
-
 `set_tweak`
 
 :   Sets an entry in the `tweaks` dictionary key that is sent in the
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/push.md
```diff
@@ -195,6 +184,12 @@ they are represented as a dictionary with a key equal to their name and
 other keys as their parameters, e.g.
 `{ "set_tweak": "sound", "value": "default" }`.
 
+{{% boxes/note %}}
+Older versions of the Matrix specification included the `dont_notify` and
+`coalesce` actions. These should both be considered no-ops (ignored, not
+rejected) if received from a client.
+{{% /boxes/note %}}
```
clokep, line 191: My fear is that someone will reject a push rule with actions of `["dont_notify"]`, which is equivalent to `[]`.

clokep:
(This is a draft since the MSC hasn't actually passed FCP yet.)

clokep:
> (This is a draft since the MSC hasn't actually passed FCP yet.)

FCP has finished, putting up for review.

uhoreg (changes requested):
The change that you've made looks good, but I think we should also be removing the places where the rules are used, e.g. in https://spec.matrix.org/v1.6/client-server-api/#predefined-rules

A grep says that there's 5 more references to `dont_notify` in `content/client-server-api/modules/push.md`, and it's also in `data/api/client-server/pushrules.yaml` (3x) and in `data/event-schemas/examples/m.push_rules.yaml` (2x)

clokep:
> I think we should also be removing the places where the rules are used, e.g. in [spec.matrix.org/v1.6/client-server-api/#predefined-rules](https://spec.matrix.org/v1.6/client-server-api/#predefined-rules)

Oops, I forgot to do that. Thanks for noticing! 👍

uhoreg (approved):

content/client-server-api/modules/push.md
Johennes: I'm trying to evaluate this change for a system where 1.3 and 1.11 servers will co-exists and am wondering: In terms of RFC2119-language, would the "ignoring without rejecting" behavior be a MUST or a SHOULD?

clokep: Probably a MUST, old clients might still use `dont_notify`?

Johennes: Yeah, I think that would make sense. Contrary to what I wrote in https://github.com/matrix-org/matrix-spec/issues/1782#issuecomment-2188074209 this morning this might actually be a case where keyword usage in an info box would be helpful.

PR #1506, opened 2023-05-01
Clarify that arbitrary unicode is allowed in user/room IDs and room aliases

tulir:


anoadragon453 (commented):
This appears like the right thing to do to me, but I'd like a second opinion.

richvdh (changes requested):

content/appendices.md
```diff
@@ -598,6 +598,13 @@ character set:
 
     extended_user_id_char = %x21-39 / %x3B-7E  ; all ASCII printing chars except :
 
+##### User IDs over federation
+
+Due to a lack of validation in original Matrix homeserver implementations,
+the localpart of user IDs over federation may contain any valid unicode
+codepoints except `:`. A future spec change may create a new room version
+to disallow such user IDs.
```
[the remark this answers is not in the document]
tulir: > aaaaaaaargh everything is awful

:+1: 

> I can't see any difference between this and "Historical user IDs". They are both "IDs you're not supposed to create but might have to deal with because apparently validation is hard"

The only difference is that there are real users who have created historical user IDs on normal clients and servers. Arbitrary unicode ids exist in the wild too, but it has never been possible to create them without modifying the server.

> do they even have to be valid unicode codepoints?

I think server implementations try to decode utf-8 sufficiently early so that invalid unicode user IDs explode before being persisted anywhere, but I'm not completely sure.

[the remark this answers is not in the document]
richvdh: > I think server implementations try to decode utf-8 sufficiently early so that invalid unicode user IDs explode before being persisted anywhere, but I'm not completely sure.

ok, that seems fair. Let's assume they *do* have to be UTF-8 until someone proves otherwise :)

content/appendices.md
```diff
@@ -655,8 +666,11 @@ The `domain` of a room alias is the [server name](#server-name) of the
 homeserver which created the alias. Other servers may contact this
 homeserver to look up the alias.
 
-Room aliases MUST NOT exceed 255 bytes (including the `#` sigil and the
-domain).
+The localpart of a room alias may contain any valid unicode codepoints
+except `:`.
+
+Room aliases MUST NOT exceed 255 bytes as UTF-8 (including the `#` sigil
+and the domain).
```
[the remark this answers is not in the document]
turt2live: I think we should assume yes until proven otherwise, when we can update the spec again.

turt2live:
@richvdh I've updated this with permission from tulir - please take another look.

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
 users whose user IDs do not conform to the permitted character set, and
 a number of rooms whose history includes events with a `sender` which
 does not conform. In order to handle these rooms successfully, clients
-and servers MUST accept user IDs with localparts from the expanded
-character set:
+and servers MUST accept user IDs with localparts consisting of any legal
+unicode codepoint except for `:`, including zero characters. Localparts
```
richvdh, line 615: I spent a while wondering what a zero character was.

```suggestion
unicode codepoint except for `:`, including the empty string. Localparts
```

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
...
+MUST be valid UTF-8 sequences.
```
richvdh, line 616: to be clear: we allow control chars (U+0000, etc) here?

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
...
+
+Servers SHOULD NOT produce user IDs with localparts outside of the following
+character set, and SHOULD NOT forward such user IDs to clients when referenced
+outside the context of an event. For example, device list updates from "invalid"
+user IDs would be dropped by the receiving server.
 
     extended_user_id_char = %x21-39 / %x3B-7E  ; all ASCII printing chars except :
 
+A future room version may prevent users using a historical character set
+from participating. Use of the historical character set is *deprecated*.
```
richvdh, line 626: I'm still not entirely following this.

As far as I can tell, we are suggesting that there are three different classes:

1. "Conformant" user IDs, consisting only of `a-z`, `0-9`, `.`, `_`, `=`, `-`, `/`, and `+`, which we'd like servers to generate.
2. "Historical" user IDs set 1, consisting of any printing ASCII except `:`, which we have to treat the same as "Conformant" user IDs.
3. "Historical" user IDs set 2, consisting of any unicode codepoints except `:`, which we filter out. Sometimes.

I'm still not entirely following the logic that says we need three different classes: do we really?

If we're going to go with it, can we try to make it clearer that we have them? Maybe give each of them their own name rather than smashing two of them together as "historical". It reads rather confusingly at the moment.

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
 users whose user IDs do not conform to the permitted character set, and
 a number of rooms whose history includes events with a `sender` which
 does not conform. In order to handle these rooms successfully, clients
-and servers MUST accept user IDs with localparts from the expanded
-character set:
+and servers MUST accept user IDs with localparts consisting of any legal
+unicode codepoint except for `:`, including zero characters. Localparts
+MUST be valid UTF-8 sequences.
+
+Servers SHOULD NOT produce user IDs with localparts outside of the following
```
richvdh, line 618: surely servers SHOULD NOT produce user IDs with localparts outside of `a-z`, `0-9`,
`.`, `_`, `=`, `-`, `/`, and `+` ?

content/appendices.md
```diff
@@ -663,6 +672,10 @@ Room IDs are case-sensitive. They are not meant to be
 human-readable. They are intended to be treated as fully opaque strings
 by clients.
 
+The localpart of a room ID (`opaque_id` above) may contain any valid
+unicode codepoints except `:`, but it is recommended to only include
+ASCII letters and digits when generating them.
```
richvdh, line 677: Let's be clearer about what's allowed: 

```suggestion
unicode codepoints except `:`, but it is recommended to only include
ASCII letters and digits (`A-Z`, `a-z`, `0-9`) when generating them.
```

content/appendices.md
```diff
@@ -663,6 +672,10 @@ Room IDs are case-sensitive. They are not meant to be
 human-readable. They are intended to be treated as fully opaque strings
 by clients.
 
+The localpart of a room ID (`opaque_id` above) may contain any valid
+unicode codepoints except `:`, but it is recommended to only include
```
richvdh, line 676: again, we allow ctrl chars?

richvdh (changes requested):

content/appendices.md
tulir: NUL is probably a special case, but other control characters are definitely allowed (I've created `@\x07:maunium.net`, `@\x08:maunium.net` and `@\x0B:maunium.net` at least)

changelogs/appendices/newsfragments/1506.clarification
```diff
@@ -0,0 +1 @@
+Clarify that arbitrary unicode is allowed in user/room IDs and room aliases.
```
kegsay, line 1: Strongly disagree on this.

The original vintage 2014 Synapse implementation [only allowed non URL quotable characters in user localparts](https://github.com/matrix-org/synapse/blob/4f475c7697722e946e39e42f38f3dd03a95d8765/synapse/rest/register.py#L39C29-L39C29) when registering for an account. Unfortunately, federation did not apply the same check, alluded to by other comments here:

> Arbitrary unicode ids exist in the wild too, but it has never been possible to create them without modifying the server.

If you are playing silly games (removing validation checks on your server) we should NOT set the precedent that we are just going to accept your games imo, otherwise where do you draw the line? Some comments in this proposal mention about excluding the null byte, but why? Sure, postgres cannot represent it when used with TEXT columns, but hey I am using a modified Synapse which doesn't have this problem, so why are we allowing some silly games but not others?

The result is an inconsistent mess, and it was _never designed_ to be that way. This isn't like other cases where "hey this is what synapse does, in the spec it goes" because in order to get the failure mode of unicode characters _you need a malicious and/or buggy actor_.

The consequences of making this the rule in the specification, and hence removing these checks in Dendrite, Conduit, et al is an increased risk of homograph attacks. I cannot and will not support increasing the attack surface of Matrix just because a few people back in 2014 removed validation and sent unicode user IDs into a room, which synapse accepted.

It is worth emphasising that room versions ARE NOT a get-out-of-jail-free card here, as user IDs are outside the scope of rooms. For example, the sliding sync proxy recently had an issue with unicode user IDs in device list changes. It's not hard to see how this can also be an issue with the user directory and to-device msgs, both of which sit outside of rooms.


Counter proposal: sorry folks with smiley poos as user localparts, you're going to be broken in the next release of Synapse, and we remove / subsequently ignore events with malformed user IDs aka what we should have done in 2014.

kegsay (changes requested):

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
...
-and servers MUST accept user IDs with localparts from the expanded
-character set:
+and servers MUST accept user IDs with localparts consisting of any legal
+unicode codepoint except for `:`, including zero characters. Localparts
+MUST be valid UTF-8 sequences.
+
+Servers SHOULD NOT produce user IDs with localparts outside of the following
+character set, and SHOULD NOT forward such user IDs to clients when referenced
+outside the context of an event. For example, device list updates from "invalid"
+user IDs would be dropped by the receiving server.
```
kegsay, line 621: This is worse than just removing events by these users, as things will only half work (E2EE will entirely break for them in weird and wonderful ways).

changelogs/appendices/newsfragments/1506.clarification
neilalexander: While I agree that this only really happens because of people modifying their servers, the result of enforcing those checks in one implementation and not in another can end up being outright disastrous for innocent users.

For example, if Dendrite enforces these checks but Synapse doesn't, then all it takes is for a Synapse user with an emoji localpart to make a power level change or perform any kind of power action to break the room for Dendrite users probably irreversibly.

Hell, they might not even have to perform a power action, because we might drop auth events (such as the user joining to begin with) or refuse to pull in prev events (from normal timeline events), which in turn causes us to run state resolution with a different set of input events (which can result in a different output state set or a complete state reset) or to accidentally propagate broken state to other servers when they ask us for `/state_ids` or `/state`.

This situation 100% sucks but Matrix only functions if implementations agree to handle these things in the same way, otherwise different implementations will never arrive at a consistent state. We already see this happen due to other corner cases and it just ends up feeling terrible for all users involved.

As for room versions, it is true that it's not a great solution to the problem due to the fact that it still leaks into device updates or other areas, but at least it's possible in a future room version to make sure that users with invalid localparts can't join the rooms to begin with. That is a _huge_ step towards stamping out invalid localparts.

richvdh:
@tulir, @turt2live: would one of you be able to address the questions in my review https://github.com/matrix-org/matrix-spec/pull/1506#pullrequestreview-1571564751?

content/appendices.md
tulir: If things like dropping device list updates from users in set 2 is done, then it must be distinct from set 1, because set 1 contains real users. Some appservices still create uppercase users too (e.g. matrix-appservice-irc: https://github.com/matrix-org/matrix-appservice-irc/issues/1399).

Other than that I guess they don't really need to be separate.

content/appendices.md
tulir: Yeah, probably best to define everything except NUL as allowed like with user IDs

turt2live:
> [...], @turt2live: would one of you be able to address the questions in my review [#1506 (review)](https://github.com/matrix-org/matrix-spec/pull/1506#pullrequestreview-1571564751)?

Which questions? The link hits a generic review. If there's outstanding comments needed, please ping inline on the threads.

richvdh:
> Which questions? The link hits a generic review. If there's outstanding comments needed, please ping inline on the threads.

Looks like @tulir has updated the PR and addressed my questions since I wrote that. Thanks @tulir!

content/appendices.md
tulir: @turt2live this is the main outstanding thing that I'm not sure about

content/appendices.md
tulir: This part probably just needs some rewording, but it depends on how the above thread about the three user ID classes is resolved.

content/appendices.md
richvdh: Right, I get this now. The proposal is to define 3 different classes of user IDs:

1. "Conformant" user IDs, consisting only of `a-z`, `0-9`, `.`, `_`, `=`, `-`, `/`, and `+`. Servers SHOULD generate user-ids using only this set.
2. "Historical" user IDs, which contain characters outside the above set, but still in the range `%x21-39 / %x3B-7E` (ie, printing ASCII). There are real users in this set, so we have to treat them the same as "Conformant" user IDs, but servers SHOULD NOT generate such user IDs.
3. "Non-conformant" user IDs, which contain characters outside the printing ASCII range. Servers must accept these in event fields such as `sender` to avoid breaking the fabric of the room, but can otherwise ignore them, for example not sending device-list updates for such users.

content/appendices.md
richvdh: If you set your localpart to a smiley poo, E2EE will break for you. That's fine by me. What I care about is you not being able to DoS a room by setting your localpart to a smiley poo, which is why we have to accept events over federation whose senders are invalid; we don't have to do much else to help such users.

content/appendices.md
```diff
@@ -611,11 +611,20 @@ characters permitted in user ID localparts. There are currently active
...
-and servers MUST accept user IDs with localparts from the expanded
-character set:
+and servers MUST accept user IDs with localparts consisting of any legal
+unicode codepoint except for `:` and `NUL` (U+0000), including other control
+characters and the empty string. Localparts MUST be valid UTF-8 sequences.
+
+Servers SHOULD NOT produce user IDs with localparts outside of the following
+character set, and SHOULD NOT forward such user IDs to clients when referenced
+outside the context of an event. For example, device list updates from "invalid"
+user IDs would be dropped by the receiving server.
```
richvdh, line 621: How about this:

```suggestion
User IDs with localparts containing characters outside the range U+0021 to U+007E, or with
an empty localpart, are considered non-compliant. For current room versions, servers must
still accept events using such user IDs over federation; however they SHOULD NOT forward
such user IDs to clients when referenced outside the context of an event. For example,
device list updates from non-compliant user IDs would be dropped by the receiving server.
```

content/appendices.md
```diff
@@ -611,10 +611,18 @@ characters permitted in user ID localparts. There are currently active
 users whose user IDs do not conform to the permitted character set, and
 a number of rooms whose history includes events with a `sender` which
 does not conform. In order to handle these rooms successfully, clients
-and servers MUST accept user IDs with localparts from the expanded
-character set:
+and servers MUST accept user IDs with localparts consisting of any legal
+unicode codepoint except for `:` and `NUL` (U+0000), including other control
+characters and the empty string. Localparts MUST be valid UTF-8 sequences.
```
richvdh, line 616: we're talking characters rather than bytes here, so "MUST be valid UTF-8 sequences" seems confusing. I think we want:

```suggestion
and servers MUST accept user IDs with localparts consisting of any legal
non-surrogate Unicode code points except for `:` and `NUL` (U+0000), including other control
characters and the empty string.
```

richvdh: I did a bit of testing on Synapse. As far as I can tell, it rejects any attempt to use the surrogates. If you encode the surrogates as UTF-8 (`<ed> <a0> <80>`, say, which is the UTF-8 encoding of U+D800), and send that over HTTP, Synapse will reject the request when it tries to decode the UTF-8. 

Likewise, it refuses to persist any strings containing the surrogates to the database, or to put them in outgoing HTTP requests/responses, because it fails to encode them as UTF-8.

I'm therefore pretty happy to say that nobody has to deal with the surrogates.

content/appendices.md
```diff
@@ -663,6 +671,11 @@ Room IDs are case-sensitive. They are not meant to be
 human-readable. They are intended to be treated as fully opaque strings
 by clients.
 
+The localpart of a room ID (`opaque_id` above) may contain any valid
+unicode codepoints, including control characters, except `:` and `NUL`
```
richvdh, line 675: Again I think we can and should exclude the surrogates here

```suggestion
non-surrogate Unicode code points, including control characters, except `:` and `NUL`
```

content/appendices.md
```diff
@@ -676,6 +689,9 @@ The `domain` of a room alias is the [server name](#server-name) of the
 homeserver which created the alias. Other servers may contact this
 homeserver to look up the alias.
 
+The localpart of a room alias may contain any valid unicode codepoints
+except `:`.
```
richvdh, line 693: ... do we allow NUL in aliases?

richvdh:
```suggestion
The localpart of a room alias may contain any valid unicode codepoints
except `:` and `NUL`.
```

richvdh (approved):
lgtm

PR #1508, opened 2023-05-02
Update spec for MSC3952: intentional mentions

clokep:
See [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952).

Preview: https://pr1508--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1508.feature
+++ b/changelogs/client_server/newsfragments/1508.feature
@@ -1,0 +1,2 @@
+Document the `m.mentions` property; the `.m.rule.is_user_mention` and `.m.rule.is_room_mention` push rules; and other behaviour from [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952).
+
--- a/content/client-server-api/modules/event_replacements.md
+++ b/content/client-server-api/modules/event_replacements.md
@@ -310,7 +310,6 @@ for re-notifying if the sending client feels a large enough revision was made).
 For example, if there is an event mentioning Alice:
 
 ```json5
-```json
 {
     "event_id": "$original_event",
     "type": "m.room.message",
--- a/content/client-server-api/modules/event_replacements.md
+++ b/content/client-server-api/modules/event_replacements.md
@@ -293,6 +293,74 @@ events will not be included in the aggregation bundled with the original
 event. Note that the subsequent edits are not actually redacted themselves:
 they simply serve no purpose within the visible timeline.
 
+#### Edits of events with mentions
+
+When editing an event with [user and room mentions](#user-and-room-mentions) the
+replacement event will hae two `m.mentions` properties:
+
+* One at the top-level of the `content`, which should contain mentions due to
+  this edit revision.
+* One inside the `m.new_content` property, which should contain the resolved mentions
+  for the final version of the event.
+
+The difference between these properties ensures that users will not be notified
+for each edit revision of an event, but allows for new users to be mentioned (or
+for re-notifying if the sending client feels a large enough revision was made).
+
+For example, if there is an event mentioning Alice:
+
+```json5
+```json
+{
+    "event_id": "$original_event",
+    "type": "m.room.message",
+    "content": {
+        "body": "Hello Alice!",
+        "m.mentions": {
+            "user_ids": ["@alice:example.org"]
+        }
+    }
+}
+```
+
+And an edit to also mention Bob:
+
+```json5
+{
+  "content": {
+    "body": "* Hello Alice & Bob!",
+    "m.mentions": {
+      "user_ids": [
+        // Include only the newly mentioned user.
+        "@bob:example.org"
+      ]
+    },
+    "m.new_content": {
+      "body": "Hello Alice & Bob!",
+      "m.mentions": {
+        "user_ids": [
+          // Include all of the mentioned users.
+          "@alice:example.org",
+          "@bob:example.org"
+        ]
+      },
+    },
+    "m.relates_to": {
+      "rel_type": "m.replace",
+      "event_id": "$original_event"
+    }
+  },
+  // other fields as required by events
+}
+```
+
+If an edit revision removes a user's mention then that user's Matrix ID should be
+included in neither `m.mentions` property.
+
+Clients may also wish to modify the [client behaviour](#user-and-room-mentions) of
+determining if an event mentions the current user by checking the `m.mentions`
+property under `m.new_content`.
+
 #### Edits of replies
 
 Some particular constraints apply to events which replace a
--- a/content/client-server-api/modules/event_replacements.md
+++ b/content/client-server-api/modules/event_replacements.md
@@ -296,7 +296,7 @@ they simply serve no purpose within the visible timeline.
 #### Edits of events with mentions
 
 When editing an event with [user and room mentions](#user-and-room-mentions) the
-replacement event will hae two `m.mentions` properties:
+replacement event will have two `m.mentions` properties:
 
 * One at the top-level of the `content`, which should contain mentions due to
   this edit revision.
--- a/content/client-server-api/modules/mentions.md
+++ b/content/client-server-api/modules/mentions.md
@@ -43,6 +43,17 @@ the localpart of Matrix IDs and display names it is recommended that clients inc
 a `m.mentions` property on each event. If there are no mentions to include it can
 be an empty object.
 
+{% boxes/rationale %}
+In previous versions of the specification, mentioning users was done by
+including the user's display name or the localpart of their Matrix ID and room
+mentions were done by including the string "@room" in the plaintext `body` of
+the event. This was prone to confusing and buggy behaviour.
+
+See the [`.m.rule.contains_display_name`](#_m_rule_contains_display_name),
+[`.m.rule.contains_user_name`](#_m_rule_contains_user_name), and
+[`.m.rule.roomnotif`](#_m_rule_roomnotif) deprecated push rules.
+{% /boxes/rationale %}
+
 #### Client behaviour
 
 In addition to using the appropriate `Matrix URI` for the mention,
--- a/content/client-server-api/modules/mentions.md
+++ b/content/client-server-api/modules/mentions.md
@@ -83,8 +83,6 @@ background color to signify to the user that they were mentioned.
 When clicked, the mention should navigate the user to the appropriate
 user or room information.
 
-TODO Behavior with edits.
-
 {{% boxes/note %}}
 Similar to legacy [matrix.to URLs](/appendices/#matrixto-navigation),
 groups used to be representable by mentions. They follow a similar format
--- a/content/client-server-api/modules/mentions.md
+++ b/content/client-server-api/modules/mentions.md
@@ -1,6 +1,8 @@
 
 ### User and room mentions
 
+{{< changed-in v="1.7" >}}
+
 This module allows users to "mention" other users and rooms within a room event.
 This is primarily used as an indicator that the recipient should receive a notification
 about the event.
--- a/content/client-server-api/modules/mentions.md
+++ b/content/client-server-api/modules/mentions.md
@@ -13,26 +13,6 @@ the event to reference the entity being mentioned.
 
 {{% definition path="api/client-server/definitions/m.mentions" %}}
 
-Although it is possible to silently mention users, it is recommended to include a
-[Matrix URI](/appendices/#uris) in the HTML body of  an [m.room.message](#mroommessage)
-event. This applies only to [m.room.message](#mroommessage) events where the `msgtype` is
-`m.text`, `m.emote`, or `m.notice`. The `format` for the event must be
-`org.matrix.custom.html` and therefore requires a `formatted_body`.
-
-Similarly, the entire room can be mentioned:
-
-```json
-{
-    "body": "Hello @room!",
-    "msgtype": "m.text",
-    "format": "org.matrix.custom.html",
-    "formatted_body": "Hello @room!",
-    "m.mentions": {
-        "room": true
-    }
-}
-```
-
 Additionally, see the [`.m.rule.is_user_mention`](#_m_rule_is_user_mention) and
 [`.m.rule.is_room_mention`](#_m_rule_is_room_mention) push rules.
 Users should not add their own Matrix ID to the `m.mentions` property as outgoing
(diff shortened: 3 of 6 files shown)


content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,50 @@
 ### User and room mentions
```
clokep, line 2: I would be curious if this is the right spot to document `m.mentions`?

content/client-server-api/modules/mentions.md
```diff
@@ -45,14 +69,19 @@ Clients should display mentions differently from other elements. For
...
+If the current user is mentioned in a message, the client should show that
 mention differently from other mentions, such as by using a red
 background color to signify to the user that they were mentioned.
 
 When clicked, the mention should navigate the user to the appropriate
 user or room information.
 
+TODO Behavior with replies.
+
+TODO Behavior with edits.
```
clokep, line 81: Does it make sense to include information here about handling replies / edits or in their respective sections? I think here (and then cross-reference), but I'm not sure.

content/client-server-api/modules/mentions.md
richvdh: I'd have thought so.

content/client-server-api/modules/mentions.md
richvdh: My instinct would actually be to go the other way around (keep the mentions section simple, and put the thorny details of the interactions with replies and edits in their respective sections). There is already an [edits of replies](https://spec.matrix.org/v1.6/client-server-api/#edits-of-replies) section, and by analogy I think an "edits of mentions" section would make sense.

Not a strong opinion though.

clokep: Edits of replies which include mentions does get thorny! 👍 That sounds like a better place indeed! Thank you!

changelogs/client_server/newsfragments/1508.feature
```diff
@@ -0,0 +1,2 @@
+Document the `m.mentions` property; the `.m.rule.is_user_mention` and `.m.rule.is_room_mention` push rules; and other behaviour from [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952).
```
clokep, line 1: The MSC also contains [a section about bridging](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/intentional-mentions/proposals/3952-intentional-mentions.md#impact-on-bridging) but honestly this isn't too interesting -- I think it mostly has to do with the event format which is covered by this PR.

Do we think we need to include any specific info in the AS spec?

clokep:
I think this is going to need a bit more work, but it is ready for an initial round of review.

content/client-server-api/modules/event_replacements.md
```diff
@@ -293,6 +293,74 @@ events will not be included in the aggregation bundled with the original
...
+  for the final version of the event.
+
+The difference between these properties ensures that users will not be notified
+for each edit revision of an event, but allows for new users to be mentioned (or
+for re-notifying if the sending client feels a large enough revision was made).
+
+For example, if there is an event mentioning Alice:
+
+```json5
+```json
```
MTRNord, line 313:
```suggestion
```

I guess one of the 2 shouldn't exist :)

content/client-server-api/modules/event_replacements.md
```diff
@@ -293,6 +293,73 @@ events will not be included in the aggregation bundled with the original
...
+        "m.mentions": {
+            "user_ids": ["@alice:example.org"]
+        }
+    }
+}
+```
+
+And an edit to also mention Bob:
+
+```json5
```
anoadragon453, line 327: Can we use json6 instead?

anoadragon453 (changes requested):
Overall I think this looks great! I just have a few changes requested below.

Edit: Sorry, ignore this. It was for a GitHub demo :)

content/client-server-api/modules/event_replacements.md
anoadragon453: I've done that!

content/client-server-api/modules/push.md
```diff
@@ -521,8 +521,43 @@ Definition:
 }
 ```
 
+**`.m.rule.is_user_mention`**
```
richvdh, line 524: consider giving these anchors so that they can be referenced from `mentions.md`:

```suggestion
<a id="_m_rule_is_user_mention"/> **`.m.rule.is_user_mention`**
```

content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,57 @@
...
+messages cannot self-notify.
+
+To disable legacy behaviour of notifications occurring due to matching against
+the localpart of Matrix IDs and display names it is recommended that clients include
+a `m.mentions` property on each event. If there are no mentions to include it can
+be an empty object.
+
 #### Client behaviour
 
 In addition to using the appropriate `Matrix URI` for the mention,
```
richvdh, line 58: I'm not convinced the text in this section is still accurate. In particular, it seems to treat "mention" as equivalent to "matrix URI link", which now seems incorrect.

content/client-server-api/modules/mentions.md
```diff
@@ -45,8 +76,7 @@ Clients should display mentions differently from other elements. For
 example, this may be done by changing the background color of the
 mention to indicate that it is different from a normal link.
 
-If the current user is mentioned in a message (either by a mention as
-defined in this module or by a push rule), the client should show that
+If the current user is mentioned in a message, the client should show that
 mention differently from other mentions, such as by using a red
 background color to signify to the user that they were mentioned.
```
richvdh, line 81: there is text below here that says group mentions look the same as room mentions. That is certainly no longer true. I feel like we should just remove the stuff about groups.

content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,57 @@
 ### User and room mentions
```
richvdh, line 3:
```suggestion

{{< changed-in v="1.7" >}}

```

content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,57 @@
...
+    "m.mentions": {
+        "room": true
+    }
+}
+```
+
+Additionally, see [the `.m.rule.is_user_mention` and `.m.rule.is_room_mention` push rules](#predefined-rules).
+Users should not add their own Matrix ID to the `m.mentions` property as outgoing
+messages cannot self-notify.
+
```
richvdh, line 50: It would be nice to include some text in here about how this changed in this spec version.

Normally we do that with a `{% boxes/rationale %}` box (for some reason)

content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,57 @@
...
+To disable legacy behaviour of notifications occurring due to matching against
+the localpart of Matrix IDs and display names it is recommended that clients include
+a `m.mentions` property on each event. If there are no mentions to include it can
+be an empty object.
```
richvdh, line 54: This is a bit hard to grok for a first-time reader who is unaware of this legacy behaviour.

```suggestion
Note that, for backwards compatibility, push rules such as `.m.rule.contains_user_name` <link> continue to
match if the *content* of the event contains the user's name or ID. To avoid unintentional notifications,
**it is recommended that clients include a `m.mentions` property on each event**. (If there are no mentions to include it can be an empty object.)
```

content/client-server-api/modules/push.md
```diff
@@ -521,8 +521,43 @@ Definition:
...
+
```
richvdh, line 525:
```suggestion

{{< added-in v="1.7" >}}

```

richvdh: Likewise in the other rule. And can we add `changed-in` for the deprecated rules?

changelogs/client_server/newsfragments/1508.feature
richvdh: Yes, wouldn't bother to call it out.

content/client-server-api/modules/event_replacements.md
```diff
@@ -293,6 +293,73 @@ events will not be included in the aggregation bundled with the original
 event. Note that the subsequent edits are not actually redacted themselves:
 they simply serve no purpose within the visible timeline.
 
+#### Edits of events with mentions
+
+When editing an event with [user and room mentions](#user-and-room-mentions) the
+replacement event will hae two `m.mentions` properties:
```
richvdh, line 299:
```suggestion
replacement event will have two `m.mentions` properties:
```

content/client-server-api/modules/mentions.md
```diff
@@ -2,26 +2,57 @@
 ### User and room mentions
 
 This module allows users to mention other users and rooms within a room message.
-This is achieved by including a [Matrix URI](/appendices/#uris) in the HTML body of
-an [m.room.message](#mroommessage) event. This module does not have any server-specific
-behaviour to it.
+This is achieved by including metadata in the `m.mentions` content property of
+the event. The `m.mentions` property consists of an object with two fields:
```
richvdh, line 6: I wonder if we need to make it explicit that `m.mentions` is encrypted in the same way as the rest of `content`?

richvdh (changes requested):
thanks @clokep! A few thoughts and comments

content/client-server-api/modules/push.md
```diff
@@ -521,8 +521,43 @@ Definition:
...
+            "set_tweak": "highlight"
+        }
+    ]
+}
+```
+
 **`.m.rule.contains_display_name`**
 
+As of `v1.7`, this rule is deprecated and should only be enabled if the event
+does not have an `m.mentions` property.
```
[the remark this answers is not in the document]
clokep: > Hrm. As far as I know all the other push rule defintions contain their conditions in the json definition. Can we do that for this `m.mentions` condition?

There's no way to represent "if a property exists" in push rules, so this is done as an exception, not using a generic mechanism. [The MSC says](https://github.com/matrix-org/matrix-spec-proposals/blob/clokep/intentional-mentions/proposals/3952-intentional-mentions.md#backwards-compatibility):

> To avoid unintentional mentions these rules are modified to only apply when the `m.mentions` property is missing

content/client-server-api/modules/mentions.md
clokep: I assumed that any property not given as an exception is encrypted by default?

content/client-server-api/modules/mentions.md
clokep: I think this section is mostly about *displaying events with mentions in them*? So yes, it needs updating but I'm not quite sure what to do there. Will take an attempt.

clokep: I think most of it is accurate-ish, I did attempt to clean it up a bit though. WDYT?

clokep:
@richvdh Thank you very much for the review! I'm quite sure this will need a bit more back and forth, but I've handled your comments *or* replied to them (I've left open the threads I consider unresolved). Please take another look at your leisure!

content/client-server-api/modules/mentions.md
```diff
@@ -1,75 +1,59 @@
 
 ### User and room mentions
 
-This module allows users to mention other users and rooms within a room message.
+{{< changed-in v="1.7" >}}
```
richvdh, line 4: sorry, I think this should have been %s to use the right spacing (for reasons I have never quite taken the time to grok)

```suggestion
{{% changed-in v="1.7" %}}
```

richvdh: similarly for the push rules

content/client-server-api/modules/mentions.md
```diff
@@ -1,75 +1,59 @@
...
-a `m.mentions` property on each event. If there are no mentions to include it can
-be an empty object.
+Note that, for backwards compatibility, push rules such as [`.m.rule.contains_display_name`](#_m_rule_contains_display_name),
+[`.m.rule.contains_user_name`](#_m_rule_contains_user_name), and
+[`.m.rule.roomnotif`](#_m_rule_roomnotif) continue to  match if the `body` of
+the event contains the user's display name or ID. To avoid unintentional notifications,
+**it is recommended that clients include a `m.mentions` property on each event**.
+(If there are no mentions to include it can be an empty object.)
+
+{% boxes/rationale %}
```
richvdh, line 28:
```suggestion
{{% boxes/rationale %}}
```

content/client-server-api/modules/mentions.md
```diff
@@ -1,75 +1,59 @@
...
+In previous versions of the specification, mentioning users was done by
+including the user's display name or the localpart of their Matrix ID and room
+mentions were done by including the string "@room" in the plaintext `body` of
+the event. This was prone to confusing and buggy behaviour.
+{% /boxes/rationale %}
```
richvdh, line 33:
```suggestion
{{% /boxes/rationale %}}
```

content/client-server-api/modules/mentions.md
richvdh: I think it's much better now, thanks!

content/client-server-api/modules/push.md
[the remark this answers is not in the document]
richvdh: Hrm, that's awkward. I worry that this important piece of information is going to get overlooked. (Also, it's unsatisfactory that the JSON retrurned by `/pushrules` is no longer comprehensive, but the time to address that problem has passed).

Can we bold the text from "should only be enabled..." ?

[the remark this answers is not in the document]
richvdh: Also, link `m.mentions` back to `https://pr1508--matrix-spec-previews.netlify.app/client-server-api/#definition-mmentions` ?

content/client-server-api/modules/mentions.md
richvdh: I hesitate because (a) there was debate about it in the MSC stage, and people might still have the wrong impression; (b) afaik the only other "mixin" property we have like this is `m.relates_to`, which is *not* encrypted.

I think a couple of words making it explicit wouldn't hurt.

content/client-server-api/modules/push.md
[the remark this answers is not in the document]
clokep: > it's unsatisfactory that the JSON retrurned by `/pushrules` is no longer comprehensive, but the time to address that problem has passed

I do hope that we can remove these rules fairly soon...

content/client-server-api/modules/mentions.md
clokep: That's fair, I added a note about ti and pointed back to how clients should process push rules (for encrypted events).

clokep:
@richvdh I've handled all those comments! I'm not sure if there's anything to do about the special behavior discussed in https://github.com/matrix-org/matrix-spec/pull/1508#discussion_r1196364305, however.

richvdh:
>  I'm not sure if there's anything to do about the special behavior discussed in [#1508 (comment)](https://github.com/matrix-org/matrix-spec/pull/1508#discussion_r1196364305), however.

Yes, this is probably the best we can do for now.

richvdh (approved):
lgtm, thank you!

PR #1511, opened 2023-05-04
Add spec for MSC2746

dbkr:
Adds https://github.com/matrix-org/matrix-spec-proposals/pull/2746

Preview: https://pr1511--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/voip_events.md
+++ b/content/client-server-api/modules/voip_events.md
@@ -122,6 +122,12 @@ or not there have been any changes to the Matrix spec.
 
 #### Events
 
+##### Common Fields
+
+{{% event-fields event_type="call_event" %}}
+
+##### Events
+
 {{% event-group group_name="m.call" %}}
 
 #### Client behaviour
--- a/content/client-server-api/modules/voip_events.md
+++ b/content/client-server-api/modules/voip_events.md
@@ -73,7 +73,7 @@ up the call.
 #### ICE Candidate Batching
 Clients should aim to send a small number of candidate events, with guidelines:
  * ICE candidates which can be discovered immediately or almost immediately in the invite/answer
-   event itself (eg. host candidates). If server reflexive or relay candiates can be gathered in
+   event itself (eg. host candidates). If server reflexive or relay candidates can be gathered in
    a sufficiently short period of time, these should be sent here too. A delay of around 200ms is
    suggested as a starting point.
  * The client should then allow some time for further candidates to be gathered in order to batch them,
--- a/content/client-server-api/modules/voip_events.md
+++ b/content/client-server-api/modules/voip_events.md
@@ -6,8 +6,8 @@ This module outlines how two users in a room can set up a Voice over IP
 WebRTC 1.0 standard. Call signalling is achieved by sending [message
 events](#events) to the room. In this version of the spec, only two-party
 communication is supported (e.g. between two peers, or between a peer
-and a multi-point conferencing unit). This means that clients MUST only
-send call events to rooms with exactly two participants.
+and a multi-point conferencing unit). Calls can take place in rooms with
+multiple members, but only two devices can take part in the call.
 
 All VoIP events have a `version` field. This is used to determine whether
 devices support this new version of the protocol. For example, clients can use
--- a/content/client-server-api/modules/voip_events.md
+++ b/content/client-server-api/modules/voip_events.md
@@ -21,6 +21,26 @@ backwards-compatible.  If it does become necessary to introduce a non
 backwards-compatible VoIP spec, the intention would be for it to simply use a
 separate set of event types.
 
+#### Party Identifiers
+Whenever a client first participates in a new call, it generates a `party_id` for itself to use for the
+duration of the call. This needs to be long enough that the chance of a collision between multiple devices
+both generating an answer at the same time generating the same party ID is vanishingly small: 8 uppercase +
+lowercase alphanumeric characters is recommended. Parties in the call are identified by the tuple of
+`(user_id, party_id)`.
+
+The client  adds a `party_id` field containing this ID to the top-level of the content of all VoIP events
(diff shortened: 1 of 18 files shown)


data/event-schemas/schema/m.call.negotiate.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+  The `description` field is the same as the `offer` field in `m.call.invite`
+  and `answer` field in `m.call.answer` and is an `RTCSessionDescriptionInit`
+  object as per https://www.w3.org/TR/webrtc/#dom-rtcsessiondescriptioninit.
+
+  Once an `m.call.negotiate` event is received, the client must respond with
+  another `m.call.negotiate` event, with the SDP answer (with `"type": "answer"`)
+  in the `description` property.
+
+  This MSC also proposes clarifying the `m.call.invite` and `m.call.answer`
```
uhoreg, line 31: Shouldn't refer to "This MSC" in the spec.  I think this paragraph needs to be applied to the `m.call.invite` and `m.call.answer` schemas.

uhoreg (changes requested):

content/client-server-api/modules/voip_events.md
```diff
@@ -42,6 +154,43 @@ Or a rejected call:
...
+   is ringing, and will time the call out in the normal way if no other device responds.
+
+##### Streams
+
+Clients are expected to send one stream with one track of kind `audio` (creating a
+voice call). They can optionally send a second track in the same stream of kind
+`video` (creating a video call).
+
+Clients implementing this specification use the first stream and will ignore
+any streamless tracks. Note that in the Javascript WebRTC API, this means
```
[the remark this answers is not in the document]
dbkr: Well, we don't have to but since it is so tied to browsers, it feels nice to explicitly explain what that means to b as helpful as possible.

data/event-schemas/schema/m.call.hangup.yaml
```diff
@@ -1,35 +1,66 @@
...
-}
+---
+type: object
+description: |
+  Sent by either party to signal their termination of the call. This can
+  be sent either once the call has has been established or before to abort the call.
+
+  The meanings of the `reason` field are as follows:
+   * `ice_timeout`: The connection failed after some media was exchanged (as opposed to current
+   * `ice_failed` which means no media connection could be established). Note that, in the case of
```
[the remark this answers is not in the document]
dbkr: Mm, yes, fairly sure this didn't come out as intended.

data/event-schemas/schema/m.call.hangup.yaml
```diff
@@ -1,35 +1,66 @@
...
+allOf:
+- "$ref": core-event-schema/room_event.yaml
+properties:
+  content:
+    type: object
+    properties:
+      call_id:
+        type: string
+        description: The ID of the call this event relates to.
+      version:
```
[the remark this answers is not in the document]
dbkr: Yeah, this is neater. The only downside is that it repeats all the fields from a room event and I'm not sure how to make it not do that.

data/event-schemas/schema/m.call.reject.yaml
```diff
@@ -0,0 +1,27 @@
...
+  but if the calling device sees an `answer` before the `reject`, it disregards the
+  reject event and carries on. The reject has a `party_id` just like an answer, and
+  the caller sends a `select_answer` for it just like an answer. If another client
+  had already sent an answer and sees the caller select the reject response instead
+  of its answer, it ends the call. If the `m.call.invite` event has `version` `0`,
+  the callee sends an `m.call.hangup` event. If the calling user chooses to end the
+  call before setup is complete, the client sends `m.call.hangup` as previously.
+
+  Note that, unlike `m.call.hangup`, this event has no `reason` field: the rejection of
+  a call is always implicitly because the user chose not to answer it.
```
uhoreg, line 15:
```suggestion
  a call is always implicitly because the user chose not to answer it.
x-addedInMatrixVersion: "1.7"
```

data/event-schemas/schema/m.call.select_answer.yaml
```diff
@@ -0,0 +1,26 @@
+{
+    "type": "object",
+    "description": "This event is sent by the caller's client once it has decided which other client to talk to, by selecting one of multiple possible incoming `m.call.answer` events. Its `selected_party_id` field indicates the answer it's chosen. The `call_id` and `party_id` of the caller is also included. If the callee's client sees a `select_answer` for an answer with party ID other than the one it sent, it ends the call and informs the user the call was answered elsewhere. It does not send any events. Media can start flowing before this event is seen or even sent.  Clients that implement previous versions of this specification will ignore this event and behave as they did before.",
```
uhoreg, line 3:
```suggestion
    "description": "This event is sent by the caller's client once it has decided which other client to talk to, by selecting one of multiple possible incoming `m.call.answer` events. Its `selected_party_id` field indicates the answer it's chosen. The `call_id` and `party_id` of the caller is also included. If the callee's client sees a `select_answer` for an answer with party ID other than the one it sent, it ends the call and informs the user the call was answered elsewhere. It does not send any events. Media can start flowing before this event is seen or even sent.  Clients that implement previous versions of this specification will ignore this event and behave as they did before.",
    "x-addedInMatrixVersion": "1.7"
```

uhoreg (approved):
lgtm modulo some suggestions

data/event-schemas/schema/m.call.select_answer.yaml
```diff
@@ -0,0 +1,27 @@
...
+    "x-addedInMatrixVersion": "1.7"
```
turt2live, line 4:
```suggestion
    "x-addedInMatrixVersion": "1.7",
```

turt2live (commented):
To fix the build

turt2live:
Looks like all threads are resolved here - merging.

PR #1517, opened 2023-05-09
Add spec for MSC2449: Require users to have visibility on an event when submitting reports

Half-Shot:
https://github.com/matrix-org/matrix-spec-proposals/issues/2249

Preview: https://pr1517--matrix-spec-previews.netlify.app


content/client-server-api/modules/report_content.md
```diff
@@ -18,3 +18,6 @@ Servers are free to handle the reported content however they desire.
 This may be a dedicated room to alert server administrators to the
 reported content or some other mechanism for notifying the appropriate
 people.
+
+The server MUST verify that the user has permission to view the event
+before accepting a report.
```
anoadragon453, line 23: Again, this feels looser than what the MSC stated, which is that the reporter must currently be joined to the room that the reported event is in. Perhaps:

```suggestion
The server MUST verify that the user reporting the event is currently
joined to the room the event is in before accepting a report.

File: data/api/client-server/report_content.yaml [hunk 2]
```diff
@@ -74,5 +77,13 @@ paths:
             type: object
           examples:
             application/json: {}
-      tags:
-        - Reporting content
+        404:
+          description: |-
+            The event was not found or you do not have permission to read this event.
```
anoadragon453, line 82: MSC2249 has [a line](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/reporting-fixes/proposals/2249-report-require-joined.md#proposal) that says it is up to implementations whether they would like to return this error for reported events that have been redacted. Could you add that here?

```suggestion
            The event was not found or you do not have permission to read this event.
            
            Homeserver implementations may additionally return this error if the
            reported event has been redacted.
```

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,14 @@ securityDefinitions:
 paths:
   "/rooms/{roomId}/report/{eventId}":
     post:
-      summary: Reports an event as inappropriate.
+      summary: Reports an event as inappropriate. You must have permission to
+        retrieve this event e.g. by being a member in the room for this event.
       description: |-
         Reports an event as inappropriate to the server, which may then notify
         the appropriate people.
```
anoadragon453, line 36: Could you add a line about combating timing attacks by adding artificial delay to the response, as mentioned in [this bit of the MSC](https://github.com/matrix-org/matrix-spec-proposals/blob/hs/reporting-fixes/proposals/2249-report-require-joined.md#security-considerations)? Something like:

```suggestion
        Reports an event as inappropriate to the server, which may then notify
        the appropriate people.
        
        It may be possible for clients to deduce whether an event exists by
        timing the response, as only a report for an event that does exist
        will require the homeserver to check whether a user is joined to
        the room. To combat this, homeserver implementations may add
        a random delay when generating a response.
```

data/api/client-server/report_content.yaml
```diff
@@ -74,5 +77,13 @@ paths:
             type: object
           examples:
             application/json: {}
-      tags:
-        - Reporting content
+        404:
```
anoadragon453, line 80: The MSC didn't actually mention the status code that should be used (and our bad for not calling this out before accepting), but `404` [was agreed upon](https://github.com/matrix-org/matrix-spec-proposals/pull/2249#discussion_r1124909199) by the SCT and the author in the PR comments.

anoadragon453 (dismissed):
Looks good overall! Just a few nitpicks looking back and forth between the MSC.

data/api/client-server/report_content.yaml
turt2live: fwiw, process-wise I'm comfortable saying that the MSC saying `M_NOT_FOUND` being the error code strongly implies a 404 status code, though we should be careful to avoid this situation in future MSCs.

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,14 @@ securityDefinitions:
 paths:
   "/rooms/{roomId}/report/{eventId}":
```
turt2live, line 30: This also needs the `x-changedInMatrixVersion` thing found on other endpoints.

content/client-server-api/modules/report_content.md
turt2live: further, a `changed-in` annotation:


```suggestion
{{< changed-in v="1.7" >}} The server MUST verify that the user 
reporting the event is currently joined to the room the event is 
in before accepting a report.
```

turt2live (changes requested):
plus what anoa said - otherwise lgtm, thank you!

changelogs/client_server/newsfragments/1517.feature
```diff
@@ -0,0 +1 @@
+Event reporters must be able to view the event in order to report it. [MSC2249](https://github.com/matrix-org/matrix-spec-proposals/pull/2249).
```
turt2live, line 1:
```suggestion
Require callers to be joined to the room to report its events, as per [MSC2249](https://github.com/matrix-org/matrix-spec-proposals/pull/2249).
```

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,20 @@ securityDefinitions:
...
     post:
-      summary: Reports an event as inappropriate.
+      summary: Reports an event as inappropriate. You must be joined to the room in order
+      to report an event.
```
turt2live, line 33: The validator be unhappy, and the summary is "wrong" from a modern point of view:
```suggestion
      summary: Report an event in a joined room as inappropriate.
```

shorter is better for the `summary`.

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,20 @@ securityDefinitions:
...
       description: |-
         Reports an event as inappropriate to the server, which may then notify
         the appropriate people.
+        
+        It may be possible for clients to deduce whether an event exists by
```
turt2live, line 38: nit: we're trying to not use "may" due to English not translating it nicely.
```suggestion
        It might be possible for clients to deduce whether an event exists by
```

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,20 @@ securityDefinitions:
...
+        timing the response, as only a report for an event that does exist
+        will require the homeserver to check whether a user is joined to
+        the room. To combat this, homeserver implementations may add
```
turt2live, line 41:
```suggestion
        the room. To combat this, homeserver implementations should add
```

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,20 @@ securityDefinitions:
 paths:
   "/rooms/{roomId}/report/{eventId}":
     post:
-      summary: Reports an event as inappropriate.
+      summary: Reports an event as inappropriate. You must be joined to the room in order
+      to report an event.
       description: |-
         Reports an event as inappropriate to the server, which may then notify
         the appropriate people.
```
turt2live, line 36:
```suggestion
        Reports an event as inappropriate to the server, which may then notify
        the appropriate people. The caller must be joined to the room to report
        it.
```

data/api/client-server/report_content.yaml
```diff
@@ -29,11 +29,20 @@ securityDefinitions:
...
+        
+        It may be possible for clients to deduce whether an event exists by
+        timing the response, as only a report for an event that does exist
+        will require the homeserver to check whether a user is joined to
+        the room. To combat this, homeserver implementations may add
+        a random delay when generating a response.
       operationId: reportContent
+      tags:
+        - Reporting content
```
turt2live, line 45: out of interest, why did this get moved up? We try to keep it at the bottom for consistency (though this isn't a real requirement)

data/api/client-server/report_content.yaml
turt2live: without "may", and wording change:
```suggestion
            The event was not found or you are not joined to the room where the
            event resides.
            
            Homeserver implementations can additionally return this error if the
            reported event has been redacted.
```

data/api/client-server/report_content.yaml
```diff
@@ -67,12 +76,26 @@ paths:
...
+        404:
+          description: |-
+            The event was not found or you do not have permission to read this event.
+
+            Homeserver implementations may additionally return this error if the
+            reported event has been redacted.
+          examples:
+            application/json: {
+              "errcode": "M_NOT_FOUND",
+              "error": "The event was not found or you do not have permission to read this event."
```
turt2live, line 98:
```suggestion
              "error": "The event was not found or you are not joined to the room."
```

data/api/client-server/report_content.yaml
turt2live: without "may":


```suggestion
        Reports an event as inappropriate to the server, which may then notify
        the appropriate people.
        
        It can be possible for clients to deduce whether an event exists by
        timing the response, as only a report for an event that does exist
        will require the homeserver to check whether a user is joined to
        the room. To combat this, homeserver implementations are encouraged
        to add a random delay when generating the response.
```

turt2live (commented):
otherwise lgtm. 

The MSC was very clear that the condition was being joined to the room, not visibility of the event - my comments are to try and bring that in line.

It also looks like you have a couple unreviewed threads from @anoadragon453 to look at 😇 (I've added to them as well)

anoadragon453:
I plan to take over the maintenance of this PR in the author's place.

data/api/client-server/report_content.yaml
anoadragon453: I folded this suggested into the one above; otherwise GitHub gets a bit confused as they overlap.

data/api/client-server/report_content.yaml
anoadragon453: I've moved this back to the bottom of the file.

anoadragon453 (commented):
This PR has been updated with changes from review.

content/client-server-api/modules/report_content.md
```diff
@@ -18,3 +18,7 @@ Servers are free to handle the reported content however they desire.
 This may be a dedicated room to alert server administrators to the
 reported content or some other mechanism for notifying the appropriate
 people.
+
+{{< changed-in v="1.7" >}} The server MUST verify that the user
```
turt2live, line 22:
```suggestion
{{< changed-in v="1.8" >}} The server MUST verify that the user 
```

data/api/client-server/report_content.yaml
```diff
@@ -59,6 +66,9 @@ paths:
         required: true
       security:
         - accessToken: []
+      x-changedInMatrixVersion:
+        1.7: |
```
turt2live, line 70:
```suggestion
        1.8: |
```

turt2live (approved):
otherwise lgtm - thanks :)

Half-Shot:
Oh, apparently I forgot I wasn't working on this...sorry @anoadragon453!

PR #1518, opened 2023-05-11
Add 'deprecated:true' to /v1/send_{join,leave}

richvdh:
The words already say this is deprecated, but it was missing the flag.

Preview: https://pr1518--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/server-server/joins-v1.yaml
+++ b/data/api/server-server/joins-v1.yaml
@@ -218,6 +218,7 @@ paths:
 
   "/send_join/{roomId}/{eventId}":
     put:
+      deprecated: true
       summary: Submit a signed join event to a resident server
       description: |-
         **Note:**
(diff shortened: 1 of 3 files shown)


data/api/server-server/joins-v1.yaml
```diff
@@ -218,6 +218,7 @@ paths:
 
   "/send_join/{roomId}/{eventId}":
     put:
+      deprecated: true
```
turt2live, line 221: We should also hit the `send_leave` endpoint with this, per the MSC mentioned in https://github.com/matrix-org/matrix-spec/pull/1518#issuecomment-1544322312

turt2live (commented):
otherwise lgtm

turt2live (approved):

PR #1521, opened 2023-05-12
Document why `/state_ids` can respond with a 404

MadLittleMods:
Document why `/state_ids` can respond with a 404

*As [discussed in an internal room](https://matrix.to/#/!SGNQGPGUwtcPBUotTL:matrix.org/$SoOPVIJch_TV84Df3uqf1SOUUNFil2d6GZJwE4mcqZs?via=jki.re&via=matrix.org&via=element.io)*

Preview: https://pr1521--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/server-server/events.yaml
+++ b/data/api/server-server/events.yaml
@@ -137,6 +137,17 @@ paths:
                   type: string
                 example: ["$an_event:example.org"]
             required: ['auth_chain_ids', 'pdu_ids']
+        404:
+          description: |-
+            No event was found or the server doesn't know about the state at that event
+            to return anything useful.
+          examples:
+            application/json: {
+                "errcode": "M_NOT_FOUND",
+                "error": "Could not find event `$Rqnc-F-dvnEYJTyHq_iKxU2bZ1CI92-kuZq3a5lr5Zg"
+              }
+          schema:
+            "$ref": "../client-server/definitions/errors/error.yaml"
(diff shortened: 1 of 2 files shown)


data/api/server-server/events.yaml
```diff
@@ -137,6 +137,17 @@ paths:
...
             required: ['auth_chain_ids', 'pdu_ids']
+        404:
+          description: |-
+            No event was found or the server doesn't know about the state at that event
+            to return anything useful.
+          examples:
+            application/json: {
+                "errcode": "M_NOT_FOUND",
+                "error": "Could not find event $Rqnc-F-dvnEYJTyHq_iKxU2bZ1CI92-kuZq3a5lr5Zg"
+              }
```
MadLittleMods, line 148: I haven't actually experienced this response personally (because I'm not a server to observe but Synapse does error with these in the logs:

```
SynapseError: 404 - State not known at event $-04o8ieIyqp4uhGC1u3RbrW_991wYR3Tp3jqeVLUAAM
SynapseError: 404 - Could not find event $2k0rFR9Msehfk9bNCWo3eTvT1QQUwEgOXlVsQ_FNEhM
```

richvdh (changes requested):
Thanks! Looks good, but please could you add a newsfragment?

richvdh (approved):
thanking you

PR #1531, opened 2023-05-22
Fixes and clarifications around m.reaction

zecakeh:
See each commit message.

The last two commits might be more controversial given the intent behind the `m.reaction` event, but they respect the original MSC and reflect the current use of this event in the wild.


richvdh (approved):
fully agreed on everything here. Thank you!

PR #1538, opened 2023-05-23
Add release checklist issue template; Document some of our timelines around releases

turt2live:
Rationale for the issue template: I'm tired of hitting copy/paste every 2 months 😛 If someone uses the issue template inappropriately, we'll treat it just like an improperly filed issue. Ideally we'd have a bot that creates this issue for us, but this is fine for now, I think.

As for the docs: many of the dates are pulled from thin air, but aren't too far off from what I already consider. I don't think the docs are in the right place here, but for sharing the knowledge as it exists this is probably fine for now.

Preview: https://pr1538--matrix-spec-previews.netlify.app

### DIFF
--- a/meta/releasing.md
+++ b/meta/releasing.md
@@ -12,7 +12,7 @@ following ranges:
 * Q1: January 20-27 (critically, before FOSDEM).
 * Q2: May 20-27.
 * Q3: August 20-27.
-* Q4: November 1-15 (before recurring November conflicts).
+* Q4: November 1-15 (before recurring November conflicts, like IETF).
 
 The SCT aims to have dates picked out by:
 
@@ -47,7 +47,7 @@ slightly different theme to it:
 * Q1: Massive feature release, if possible. This generally happens thanks to FOSDEM.
 * Q2: Regular feature release, if possible.
 * Q3: Momentum-continuing feature release, if possible.
-* Q4: Maintenance release (or features, if possible).
+* Q4: Preferably a maintenance release, but will accept features per normal.
 
 ## Prerequisites / preparation
 
--- a/meta/releasing.md
+++ b/meta/releasing.md
@@ -4,6 +4,51 @@ The whole specification is now released as a single unit/artifact. This document
 the process for releasing the specification and a description of how the (public)
 machinery works.
 
+## Timeline
+
+The spec is released each calendar quarter. The target release dates are within the
+following ranges:
+
+* Q1: January 20-27 (critically, before FOSDEM).
+* Q2: May 20-27.
+* Q3: August 20-27.
+* Q4: November 1-15 (before recurring November conflicts).
+
+The SCT aims to have dates picked out by:
+
+* Q1: January 10.
+* Q2: May 1.
+* Q3: August 1.
+* Q4: October 15.
+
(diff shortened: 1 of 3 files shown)


meta/releasing.md
```diff
@@ -4,6 +4,51 @@ The whole specification is now released as a single unit/artifact. This document
...
+
+*This section is a work in progress.*
+
+Mentioned above, the SCT aims to have spec releases quarterly. Each quarter has a
+slightly different theme to it:
+
+* Q1: Massive feature release, if possible. This generally happens thanks to FOSDEM.
+* Q2: Regular feature release, if possible.
+* Q3: Momentum-continuing feature release, if possible.
+* Q4: Maintenance release (or features, if possible).
```
richvdh, line 50: somewhat unconvinced we need to spell this out, or that we should. It might end up working out this way, but I don't think it's a thing that we need to strive for.

In other words: if a Q4 release ends up with lots of new features for some reason, is that a bad thing?

meta/releasing.md
```diff
@@ -4,6 +4,51 @@ The whole specification is now released as a single unit/artifact. This document
...
 
+## Timeline
+
+The spec is released each calendar quarter. The target release dates are within the
+following ranges:
+
+* Q1: January 20-27 (critically, before FOSDEM).
+* Q2: May 20-27.
+* Q3: August 20-27.
+* Q4: November 1-15 (before recurring November conflicts).
```
richvdh, line 15: .... (what happens in November?)

meta/releasing.md
```diff
@@ -4,6 +4,51 @@ The whole specification is now released as a single unit/artifact. This document
...
+
+The SCT aims to have dates picked out by:
+
+* Q1: January 10.
+* Q2: May 1.
+* Q3: August 1.
+* Q4: October 15.
```
richvdh, line 22: this is all very formal... Honestly I'd have thought "The SCT will normally announce a release date 2-3 weeks before the release" would be fine.  Does anyone other than ourselves actually care exactly what release date is, to that degree?

richvdh (approved):
lgtm.

Have emitted a couple of thoughts which you are welcome to disregard.

meta/releasing.md
turt2live: It's not a bad thing to have more things than what's here. This is largely saying that by November we don't usually have a lot of features, so releasing without them is fine.

The section is missing a bunch of words about how we don't want to overload projects with feature work, but also don't want to arbitrarily delay releases - this balance is something we do, but don't have words for. The rough idea is that if we make a large release then there should be a smaller release to follow it (maybe not the immediate next one, but pretty close).

meta/releasing.md
turt2live: IETF meetup, people start disappearing for holidays, most projects are wrapping up, etc.

meta/releasing.md
turt2live: This doc is written for us, it just happens to be publicly accessible. We don't really have a good place to put process stuff like this, so I've stuck it in the release procedures. No one but us cares how to cut a release either :D

meta/releasing.md
turt2live: have stuck a mention in the doc so we don't forget

meta/releasing.md
turt2live: have clarified the intent of the release, I think

PR #1516, opened 2023-05-09
Add spec for MSC2659: application service ping endpoint

tulir:

PR #1541, opened 2023-05-24
Mark the appservice ping response `duration_ms` field as required

zecakeh:
As intended in MSC2659, forgotten in #1516.

Checked with the author [in the matrix spec room](https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$jKIn1Rh4T6NcCO0D38rFyOi6JOlQCxHSPr-kFHfU76o?via=matrix.org&via=libera.chat&via=element.io).

Preview: https://pr1541--matrix-spec-previews.netlify.app


turt2live (approved):
Thanks!

PR #1546, opened 2023-05-26
Fix `m.call.negotiate` schema and example

zecakeh:
Introduced in #1511. The properties and example look like a copy-paste of `m.call.invite`, but don't match the description of the event, nor MSC2746.

cc @dbkr

Preview: https://pr1546--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/examples/m.call.negotiate.yaml
+++ b/data/event-schemas/examples/m.call.negotiate.yaml
@@ -5,6 +5,7 @@
     "version" : "1",
     "party_id": "67890",
     "call_id": "12345",
+    "lifetime": 10000,
     "description": {
       "type" : "offer",
       "sdp" : "v=0\r\no=- 6584580628695956864 2 IN IP4 127.0.0.1[...]"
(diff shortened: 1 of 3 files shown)


zecakeh:
Furthermore it looks like this paragraph in the description is out of place. It probably deserves to be in the text of the module, and/or on the corresponding events description:

https://github.com/matrix-org/matrix-spec/blob/bd6644b6ee0fa909439b517a4c185a5393f42bbb/data/event-schemas/schema/m.call.negotiate.yaml#L30-L35

And given the text in it, I'm wondering if we should remove the `enum` property of the corresponding schemas, so clients don't try to validate them against the given values.

data/event-schemas/examples/m.call.negotiate.yaml
```diff
@@ -5,8 +5,7 @@
     "version" : "1",
     "party_id": "67890",
     "call_id": "12345",
-    "lifetime": 10000,
```
dbkr, line 8: negotiate still has the lifetime field though: https://github.com/matrix-org/matrix-spec-proposals/pull/2746/files#diff-4d65532ebdb47ba1fd5ea93342395abfeaab0e911360c0bd8d8839be90afd7dfR176

zecakeh: Right, sorry about that.

dbkr (approved):

PR #1561, opened 2023-06-09
Use tag name as the OpenAPI definition version

zecakeh:
I wanted to fix the spec version displayed in the [Matrix API Viewer](https://matrix.org/docs/api/#overview), which is set to "unstable" although I'm pretty sure it is only deployed for releases.

However I'm not sure how the OpenAPI definitions are deployed. My guess is that they use the artifact from the job triggered by tagging the release so this changes the CI script to use the tag name as the version, when it is available.

It might not fix what I want but at least it's not wrong :slightly_smiling_face:.

Preview: https://pr1561--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -93,18 +93,26 @@ jobs:
           pip install -r scripts/requirements.txt
       - name: "📦 Asset creation"
         run: |
+          if [[ "${GITHUB_REF}" == refs/tags/* ]]; then
+            export RELEASE="${GITHUB_REF/refs\/tags\//}"
+          else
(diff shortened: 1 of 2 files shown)


richvdh:
> However I'm not sure how the OpenAPI definitions are deployed. My guess is that they use the artifact from the job triggered by tagging the release so this changes the CI script to use the tag name as the version, when it is available.

Yes, that's right. It's actually a manual job to download the artifact from Github and upload it to the web server. Releases don't happen often enough for us to have automated it :)

richvdh (approved):
lgtm

PR #1563, opened 2023-06-10
Make sure examples types match schema in definitions

### LINKED ISSUE #742: Example for POST /identity/v2/terms is wrong
It shows posting a single string when it means to accept an array

zecakeh:
Most of those are integers or booleans being represented as strings or vice-versa, but a couple are a bit more involved and can change the example in the definition.

Fixes #742.

Preview: https://pr1563--matrix-spec-previews.netlify.app


data/api/client-server/pushrules.yaml
```diff
@@ -545,7 +545,7 @@ paths:
           required: true
           description: |
             The kind of rule
-          example: cake
+          example: content
```
KitsuneRal, line 548: ?

data/api/identity/v2_terms.yaml
```diff
@@ -131,7 +131,7 @@ paths:
                   items:
                     type: string
                   description: The URLs the user is accepting in this request.
-                  example: https://example.org/somewhere/terms-2.0-en.html
+                  example: [https://example.org/somewhere/terms-2.0-en.html]
```
KitsuneRal, line 134:
```suggestion
                  example: ['https://example.org/somewhere/terms-2.0-en.html']
```
For outmost clarity

data/api/server-server/leaving-v2.yaml
```diff
@@ -109,15 +109,15 @@ paths:
                 - depth
                 - content
             example: {
-              "room_id": "!somewhere:example.org",
```
KitsuneRal, line 112: Not sure this is accurate from the definition side - someone who knows Federarion API is needed here.

data/api/client-server/pushrules.yaml
zecakeh: It's an enum so it has to match one of the acceptable values.

data/api/server-server/leaving-v2.yaml
zecakeh: Note that this example is not displayed in the spec. It is replaced by the example generated from the schema properties examples.

And since the schema doesn't include a `room_id` property, [the spec doesn't show one](https://spec.matrix.org/v1.7/server-server-api/#put_matrixfederationv2send_leaveroomideventid).

data/api/client-server/pushrules.yaml
KitsuneRal: Oh - indeed

KitsuneRal (approved):

PR #1564, opened 2023-06-10
Allow `null` in `room_types` for `POST /publicRooms` endpoints

zecakeh:
Also fixes rendering of `type` as an array.

Note that this will probably conflict with #1559, as it edits a similar area of `partials/openapi/render-object-table.html`.

Preview: https://pr1564--matrix-spec-previews.netlify.app


erikjohnston (approved):
Looks sane to me!

PR #1578, opened 2023-06-17
Fix headers with custom IDs via the correct syntax

### LINKED ISSUE #1572: PR #1423 broke the display  of the Server-Server API "Knocking Rooms" header
**Link to problem area**:

https://spec.matrix.org/v1.7/server-server-api/#restricted-rooms

image

**Issue**
Header isn't displayed properly (it's just raw markdown)

#1423 seems to have introduced the change, given the version in #1407 seems to appear right.

midnightveil:
Fixes #1572

Preview: https://pr1578--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1578.clarification
+++ b/changelogs/client_server/newsfragments/1578.clarification
@@ -1,0 +1,1 @@
+Fix broken header formatting. Contributed by @midnightveil.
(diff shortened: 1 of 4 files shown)


richvdh (changes requested):
Thanks! As the CI is reporting, there is now a broken link. Looks like it is here: https://github.com/matrix-org/matrix-spec/blob/0a502a367f36e624e32e7a4753ff1b47360cf8dc/content/client-server-api/modules/end_to_end_encryption.md?plain=1#L138

richvdh (approved):
Thank you!

PR #1579, opened 2023-06-19
Render binary request and response bodies

zecakeh:
The two first commits are cleanups necessary before making the spec rendering work.

The idea is to use a different kind of table when the data is not JSON.

Here is what it looks like for a request (notice that the quotes around the example are removed):

image

This is the same for a response:

Capture d’écran du 2023-06-19 15-45-59

And here is the 200 response of `GET /_matrix/media/v3/thumbnail/{serverName}/{mediaId}`, with 2 content types:

Capture d’écran du 2023-06-19 15-46-44

Preview: https://pr1579--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -374,7 +374,7 @@ footer {
       padding: 1rem;
     }
 
-    &.object-table, &.response-table {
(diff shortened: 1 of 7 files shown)


richvdh (approved):
looks great, thanks!

PR #1582, opened 2023-06-19
Specify our usage of ABNF for grammar

turt2live:
Preview: https://pr1582--matrix-spec-previews.netlify.app

### DIFF
--- a/meta/documentation_style.rst
+++ b/meta/documentation_style.rst
@@ -189,6 +189,6 @@ When a new subheading is added, it must be referenced and ordered in this docume
 Describing grammar
 ~~~~~~~~~~~~~~~~~~
 
-Use  [RFC5234-style ABNF](https://datatracker.ietf.org/doc/html/rfc5234) when describing
+Use  `RFC5234-style ABNF <https://datatracker.ietf.org/doc/html/rfc5234>`_ when describing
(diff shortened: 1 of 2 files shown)


meta/documentation_style.rst
```diff
@@ -185,3 +185,10 @@ The subheadings which are always deemed as client-affecting are:
 * Redactions
 
 When a new subheading is added, it must be referenced and ordered in this document.
+
+Describing grammar
+~~~~~~~~~~~~~~~~~~
+
+Use  [RFC5234-style ABNF](https://datatracker.ietf.org/doc/html/rfc5234) when describing
```
richvdh, line 192: this is not how you write an RST link

```suggestion
Use  `RFC5234-style ABNF <https://datatracker.ietf.org/doc/html/rfc5234>`_ when describing
```

I *think*.

richvdh (changes requested):

meta/documentation_style.rst
turt2live: Argh, I remembered it was RST for the header and then immediately forgot.

turt2live: (github preview says yes, that's an RST link)

richvdh (approved):

PR #1584, opened 2023-06-22
Fix schema of custom fields in query for appservice API

zecakeh:
See discussion at https://github.com/matrix-org/matrix-spec/pull/1310#discussion_r1054701952.


KitsuneRal (commented):
I think the ellipses from the parameter name can also be dropped now, unless something else relies on them (I really hope not)

zecakeh:
I removed the ellipses, I don't think anything relied on them.

KitsuneRal (approved):
Thanks!!

PR #1585, opened 2023-06-22
Remove unnecessary `oneOf`s in JSON schemas

zecakeh:
`oneOf` is only necessary if there are more fields than the `type`.

Preview: https://pr1585--matrix-spec-previews.netlify.app


KitsuneRal (approved):
Fair and square :)

Issue #1553, opened 2023-06-01
MAC calculation description is unclear


richvdh:
https://spec.matrix.org/v1.7/client-server-api/#mac-calculation says:

> The MAC used is HMAC as defined in [RFC 5869](https://tools.ietf.org/html/rfc5869), using SHA-256 as the hash function. The shared secret is supplied as the input keying material. No salt is used, and in the info parameter is the concatenation of:

There are two problems here:

 * RFC5869 doesn't actually define HMACs - that is done in [RFC2104](https://datatracker.ietf.org/doc/html/rfc2104).
 * An HMAC doesn't *have* an info parameter.

What I think actually happens is:

 1. We derive a key using an HKDF as defined in RFC5869, using: SHA-256 as the hash, the shared secret as the input keying material, no salt is used, and the info parameter as given.
 2. We use the resulting key in an HMAC calculation, as defined in RFC2104, again using SHA-256 as the hash.

PR #1590, opened 2023-06-27
fix description of MAC calculation

uhoreg:
fixes https://github.com/matrix-org/matrix-spec/issues/1553

Preview: https://pr1590--matrix-spec-previews.netlify.app


richvdh (changes requested):
This is much better, but tbh I find the section rather hard to follow. Can we break it into a numbered list?

1. Do the HDKF
2. Do the MAC
3. Encode as b64

richvdh (approved):
looks great!

Issue #1544, opened 2023-05-25
Table of contents disappears on modern hugo versions


turt2live:
We probably need to update our docsy template. Thankfully, the GHA workflow is [pinned](https://github.com/matrix-org/matrix-spec/blob/fbb8a789f605a304e53b79e96b700b7f154aafcd/.github/workflows/main.yml#L150).

Tested with v0.111.3, v0.112.0, and v0.112.3 and all 3 don't *fully* build the TOC. Confirmed fixed with our pinned version (v0.93.3).

Example:
![image](https://github.com/matrix-org/matrix-spec/assets/1190097/d09d590c-61db-4370-947b-10e9ef1d8b6f)

It's missing all the other APIs, the changelog page, etc. 

I suspect [this check](https://github.com/matrix-org/matrix-spec/blob/fbb8a789f605a304e53b79e96b700b7f154aafcd/layouts/partials/sidebar-tree.html#L36) in sidebar-tree is responsible, and has changed sometime between 0.93.3 and now. The parent partial, `sidebar.html`, is *not* falling into its cache condition.

PR #1551, opened 2023-05-30
Fix the sidebar in recent versions of Hugo

zecakeh:
Basically what I did is:
1. Replace `layout/partials/sidebar-tree.html` with the one from the matrix-org/docsy fork.
2. Add back the TOC div.
3. Adapt the CSS and config to look like before.

Tested in Firefox and Brave with Hugo 0.111.3.

Fixes #1544.

Signed-off-by: Kévin Commaille <zecakeh@tedomum.fr>

Preview: https://pr1551--matrix-spec-previews.netlify.app

PR #1591, opened 2023-06-27
Bump hugo version used for the build

richvdh:
Since https://github.com/matrix-org/matrix-spec/issues/1544 is fixed, we can use a modern hugo.

Preview: https://pr1591--matrix-spec-previews.netlify.app


turt2live (approved):
A bit unfortunate that the room versions aren't auto-expanded on the index, but I'll survive.

richvdh:
> A bit unfortunate that the room versions aren't auto-expanded on the index, but I'll survive.

This came up in https://github.com/matrix-org/matrix-spec/pull/1551. Basically, the old behaviour means maintaining a fork of one of the template files and I don't think it's worth the hassle.

Issue #1372, opened 2022-12-06
Emoji SAS translations broken since the split into `-spec` and `-proposals`


turt2live:
This was brought up a while ago, and I hoped to have time to address it - that has not been the case. 

Context: the [SAS Emoji](https://spec.matrix.org/v1.5/client-server-api/#sas-method-emoji) are used in verification, and the spec references a "common translation" being available from this repo. Those translations are meant to be updated from time to time, but we tend to forget - it should become part of the release process.

The translations are specified on Element's weblate instance (https://translate.element.io/projects/matrix-doc/sas-emoji-v1/) and stored in-tree on our repo: https://github.com/matrix-org/matrix-spec/tree/main/data-definitions

The weblate git repo is extraordinarily confused as it doesn't understand what happened when some force pushes happened to the matrix-doc (now spec-proposals) repo. It needs manually unpicking on the weblate side with a lot of git magic.

Once fixed on the weblate side, we should ensure a point gets added to the release process to synchronize the translations - this would be the exact same process Element Web uses to ship (Lock, Commit, Push, Merge PR on github, Pull, Unlock).

turt2live:
With some effort, I think I've done a thing which fixes this, at least for now/manually.

First, I've updated the repo details from `matrix-doc` to `matrix-spec`, which made weblate correctly determine the diff, but wouldn't let it push to upstream. Now I've manually checked out weblate's repo (turns out it exposes it), added a few different upstreams to that checkout, and cherrypicked weblate's merge commit to matrix-spec directly. 

This has resulted in the following commits:
* https://github.com/matrix-org/matrix-spec/commit/f65007476b93ce3be009794e06b473cf5476b194
* https://github.com/matrix-org/matrix-spec/commit/7aaeac07608f97942eff88634a56b630a745ce56 (we're supposed to run a script when we do i18n updates)

After doing this, weblate's repo was further desynced so I reset it (since the commits were carried over somewhere safe). It's now happy enough:

![image](https://github.com/matrix-org/matrix-spec/assets/1190097/4423eb8d-d98f-4d9c-992e-817cec9b784f)


I'm not sure if it'll work in the future, but at least now we know we can manually fix things relatively easily.

PR #1593, opened 2023-06-29
Update link to SAS emoji data

richvdh:


turt2live (approved):
worth noting that we're doing a bad job of updating the referenced JSON file because of https://github.com/matrix-org/matrix-spec/issues/1372 btw

PR #1598, opened 2023-07-12
Fix rendered changelog with new version of towncrier

zecakeh:
The version of towncrier is not pinned in CI so lately the rendering of the changelog is broken in the unstable version (note the changelog categories that are now numbered titles):

image

This is due to https://github.com/twisted/towncrier/pull/483 that is included in towncrier 23.6.0.

The first commit fixes that by replacing the default template by a custom one. It also allows to put formatting in the template only rather than in the config file.

The second commit allows to simplify the `generate-changelog.sh` script by using the built-in [sections](https://towncrier.readthedocs.io/en/stable/configuration.html#sections) feature of towncrier, instead of generating the changelog for each section separately.

This is the result, which should be the same as before:
image

We could almost get rid of the `header.md` file except we cannot format the release date.

Preview: https://pr1598--matrix-spec-previews.netlify.app

### DIFF
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -2,7 +2,7 @@
 # current at the time of writing
 
 # we need at least version 4.0.0 for support of JSON Schema Draft 2020-12.
-jsonschema >= 4.0.0
+jsonschema == 4.17.3
 
 PyYAML >= 3.12
 requests >= 2.18.4
--- a/scripts/requirements.txt
+++ b/scripts/requirements.txt
@@ -6,4 +6,4 @@ jsonschema >= 4.0.0
 
 PyYAML >= 3.12
 requests >= 2.18.4
-towncrier == 21.9.0rc1
+towncrier == 23.6.0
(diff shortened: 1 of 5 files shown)


scripts/requirements.txt
```diff
@@ -6,4 +6,4 @@ jsonschema >= 4.0.0
 
 PyYAML >= 3.12
 requests >= 2.18.4
-towncrier == 21.9.0rc1
+towncrier == 23.6.0
```
zecakeh, line 9: Note that towncrier's version is pinned here but not in CI, should that be the case?

richvdh: > Note that towncrier's version is pinned here but not in CI, should that be the case?

Yes, it should probably be pinned in CI too. Or rather, to save us inevitably forgetting to update one or the other, CI should `pip install -r requirements.txt` rather than installing a random version of towncrier.

richvdh (approved):
fantastic, thank you!

Issue #1369, opened 2022-11-30
Bug fix PRs to tooling don't have a sensible/obvious changelog entry type


turt2live:
For example, https://github.com/matrix-org/matrix-spec/pull/1368 - normally this would be an internal bug fix, but the closest we have is "clarification" or "feature". Neither particularly cover the task well.

PR #1601, opened 2023-07-13
Use auto table-layout on desktop

not-my-profile:
On desktop the tables in the specification currently have quite awkward column widths, making them harder to read.

| Before | After |
|--|--|
|image | image

While the fixed table-layout does have an advantage on mobile since it avoids unnecessary horizontal scrolling in the tables, we can just use it on narrow browser widths via a media query, while still using the more readable auto layout on desktop.

Preview: https://pr1601--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -355,7 +355,7 @@ footer {
        * results in annoying horizontal scrolling on mobile, so we instead
        * switch to a fixed table layout on a narrow browser width.
        * (On a wider width the default auto table-layout provides better readability.)
-      *
+       *
        * Docsy makes all tables "responsive tables", which causes Bootstrap 4 to create
        * tables with a "display" property of "block".
        * However, for "table-layout: fixed" to be effective, an element must have a
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -355,7 +355,7 @@ footer {
        * results in annoying horizontal scrolling on mobile, so we instead
        * switch to a fixed table layout on a narrow browser width.
        * (On a wider width the default auto table-layout provides better readability.)
-
+      *
        * Docsy makes all tables "responsive tables", which causes Bootstrap 4 to create
        * tables with a "display" property of "block".
        * However, for "table-layout: fixed" to be effective, an element must have a
--- a/assets/scss/custom.scss
+++ b/assets/scss/custom.scss
@@ -350,19 +350,37 @@ footer {
   }
 
   table {
-    /* Docsy makes all tables "responsive tables", which causes Bootstrap 4 to create
-     * tables with a "display" property of "block".
-     *
-     * However, for "table-layout: fixed" to be effective, an element must have a
-     * "display" property of "table".
-     *
-     * Thus, we override the "display" property here. This may no longer be necessary once
-     * Docsy updates to Bootstrap v5+: https://github.com/google/docsy/issues/470.
-     * For more details, see
-     * https://github.com/matrix-org/matrix-spec/pull/1295/files#r1010759688 */
-    display: table;
-    table-layout: fixed;
-    width: 100%;
+    @media (max-width: 800px) {
+      /* Docsy by default applies `overflow-x: auto;` to tables, which
+       * results in annoying horizontal scrolling on mobile, so we instead
+       * switch to a fixed table layout on a narrow browser width.
+       * (On a wider width the default auto table-layout provides better readability.)
+
+       * Docsy makes all tables "responsive tables", which causes Bootstrap 4 to create
+       * tables with a "display" property of "block".
+       * However, for "table-layout: fixed" to be effective, an element must have a
+       * "display" property of "table".
+       *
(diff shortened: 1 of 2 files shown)


assets/scss/custom.scss
```diff
@@ -350,19 +350,37 @@ footer {
...
+
+       * Docsy makes all tables "responsive tables", which causes Bootstrap 4 to create
+       * tables with a "display" property of "block".
+       * However, for "table-layout: fixed" to be effective, an element must have a
+       * "display" property of "table".
+       *
+       * Thus, we override the "display" property here. This may no longer be necessary once
+       * Docsy updates to Bootstrap v5+: https://github.com/google/docsy/issues/470.
+       * For more details, see
+       * https://github.com/matrix-org/matrix-spec/pull/1295/files#r1010759688 */
```
richvdh, line 367: I wonder if we should update our fork of docsy now that https://github.com/google/docsy/issues/470 is closed...

Anyway, I guess that is a separate problem.

richvdh (changes requested):
Thanks! Please add a changelog and sign-off.

not-my-profile:
Just signed-off the commit. I'd be happy to add a changelog entry but none of the allowed changelog entry types fits this PR (new, feature, clarification, breaking, deprecation).  I'd classify this as "style" but I'm assuming `scripts/check-newsfragments` won't let that pass.

richvdh:
Thanks! You could make it a `clarification` in the [internal](https://github.com/matrix-org/matrix-spec/tree/main/changelogs/internal/newsfragments) section.

not-my-profile:
CONTRIBUTING.rst defines `clarification` as:

> clarification - Used when an area of the spec is being improved upon and does not change or introduce any functionality.

That really sounds like a change to the text of the spec ... so I don't really want to classify it as "clarification" since I'd consider that as misleading.

richvdh:
I mean you're right, and https://github.com/matrix-org/matrix-spec/issues/1369 mentions the fact that it's not a great match. But it's what we use for now.

Compare https://spec.matrix.org/v1.7/changelog/#internal-changestooling for where we've done this in the past. None of those are really changes to the spec.

not-my-profile:
Fair enough ... added the newsfragment.

assets/scss/custom.scss
```diff
@@ -350,19 +350,37 @@ footer {
...
-     * https://github.com/matrix-org/matrix-spec/pull/1295/files#r1010759688 */
-    display: table;
-    table-layout: fixed;
-    width: 100%;
+    @media (max-width: 800px) {
+      /* Docsy by default applies `overflow-x: auto;` to tables, which
+       * results in annoying horizontal scrolling on mobile, so we instead
+       * switch to a fixed table layout on a narrow browser width.
+       * (On a wider width the default auto table-layout provides better readability.)
+
```
richvdh, line 358:
```suggestion
      *
```

changelogs/internal/newsfragments/1601.clarification
```diff
@@ -0,0 +1 @@
+Improve the layout of tables on desktop.
```
richvdh, line 1:
```suggestion
Improve the layout of tables on desktop displays. Contributed by @not-my-profile.
```

richvdh (approved):
Thank you!

(Aside: it's fine in this case, because the changes are relatively small, but please don't force-push after a review: it makes it hard to see what has changed since the previous review. We can clean up the commit history when we merge the PR.)

assets/scss/custom.scss
```diff
@@ -350,19 +350,37 @@ footer {
...
-     * https://github.com/matrix-org/matrix-spec/pull/1295/files#r1010759688 */
-    display: table;
-    table-layout: fixed;
-    width: 100%;
+    @media (max-width: 800px) {
+      /* Docsy by default applies `overflow-x: auto;` to tables, which
+       * results in annoying horizontal scrolling on mobile, so we instead
+       * switch to a fixed table layout on a narrow browser width.
+       * (On a wider width the default auto table-layout provides better readability.)
+      *
```
richvdh, line 358: damnit

```suggestion
       *
```

changelogs/internal/newsfragments/1601.clarification
not-my-profile: I don't really like my work being attributed to my GitHub username :/

richvdh: Oh, sorry. I can remove this.

richvdh: would you prefer an alternative attribution?

not-my-profile: Yeah: Martin Fischer.

richvdh: Done: https://github.com/matrix-org/matrix-spec/commit/f5035b8e0271649816ab35dc12e9a5cc5e7783c3

not-my-profile: Thanks <3

PR #1602, opened 2023-07-18
Spec for MSC3077

dbkr:
Specs https://github.com/matrix-org/matrix-spec-proposals/pull/3077

Preview: https://pr1602--matrix-spec-previews.netlify.app


data/event-schemas/schema/m.call.negotiate.yaml
```diff
@@ -64,6 +64,23 @@ properties:
...
+          type: object
+          title: Stream Metadata Block
+          properties:
+            purpose:
+              type: string
+              enum:
+              - m.usermedia
+              - m.screenshare
+    required:
+    - purpose
```
richvdh, line 82: the indentation on this looks wrong

data/event-schemas/schema/m.call.negotiate.yaml
```diff
@@ -64,6 +64,23 @@ properties:
           Once the invite age exceeds this value, clients should discard it.
           They should also no longer show the call as awaiting an answer in the
           UI.
+      sdp_stream_metadata:
+        type: object
+        title: SDP Stream Metadata
+        x-addedInMatrixVersion: '1.7'
+        description: Object describing the streams that will be sent
```
richvdh, line 71: missing full stop

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
 
 ##### Streams
```
richvdh, line 172: can we add something to https://pr1602--matrix-spec-previews.netlify.app/client-server-api/#common-fields about the new field?

data/event-schemas/schema/m.call.negotiate.yaml
richvdh, line 71: should probably link to https://pr1602--matrix-spec-previews.netlify.app/client-server-api/#streams, since there is insufficient documentation inline

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
-any streamless tracks. Note that in the JavaScript WebRTC API, this means
-`addTrack()` must be passed two parameters: a track and a stream, not just a
-track, and in a video call the stream must be the same for both audio and video
-track.
-
-A client may send other streams and tracks but the behaviour of the other party
-with respect to presenting such streams and tracks is undefined.
+Clients may send more than one stream in a VoIP call. Metadata can be included in
+the `m.call.invite`, `m.call.answer` and `m.call.negotiate` events to describe the
+streams being sent. This is the `sdp_stream_metadata` field.
```
richvdh, line 176: please prefer 'property' over 'field': https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#general

richvdh, line 176:
```suggestion
streams being sent, using the `sdp_stream_metadata` property.
```

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
+
+This field is an object in which each key is one stream `id` in the session
+description. The values are objects with the following fields:
+
+ * `purpose` - a string indicating the purpose of the stream. For compatibility
+   between client the following values are defined:
```
richvdh, line 182: why "For compatibility between client [sic]" ?

richvdh, line 182: this is pretty hard to grok. It doesn't need a formal description of the structure here; that is provided by the event definition tables. Better to be informal and easier to parse.

```suggestion
`sdp_stream_metadata` maps from the `id` of a stream in the session description, 
to metadata about that stream. 

Currently only one property is defined for the metadata. This is `purpose`, which should be a stringindicating the purpose of the stream. The following `purpose`s are defined:
```

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
+
+If an incoming stream is not described in `sdp_stream_metadata` and
+`sdp_stream_metadata` is present, the stream should be ignored. If a stream has
+a `purpose` of an unknown type (i.e. not `m.usermedia` or `m.screenshare`), it
+should be ignored.
+
+Clients implementing this specification will ignore any streamless tracks. Note
+that in the JavaScript WebRTC API, this means `addTrack()` must be passed two
+parameters: a track and a stream, not just a track, and in a video call the
+stream must be the same for both audio and video track.
```
richvdh, line 194: I know this is existing text, but it now seems even more confusing than before, so I think we need to clarify.

1. What is the mechanism by which streamless tracks are ignored? Or is this an *instruction* to clients to ignore  such tracks?
2. what is that to do with `addTrack`'s parameters?
3. is it related to `sdp_stream_metadata`? if not, it should probably not be in the middle of the text about it?

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
+
+During the initial invite and answer exchange clients find out if the field
+`sdp_stream_metadata` is missing. For backwards compatibility if it is not
+present in the event sent by the opponent, the client should ignore any new
+incoming streams (i.e. it should use the first one) and it shouldn't send more
+than one stream (i.e. clients cannot send a video feed and a screenshare at the
+same time, as is the case in current clients).
```
richvdh, line 201: inappropriate outside an MSC

```suggestion
same time).
```

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
+should be ignored.
+
+Clients implementing this specification will ignore any streamless tracks. Note
+that in the JavaScript WebRTC API, this means `addTrack()` must be passed two
+parameters: a track and a stream, not just a track, and in a video call the
+stream must be the same for both audio and video track.
+
+During the initial invite and answer exchange clients find out if the field
+`sdp_stream_metadata` is missing. For backwards compatibility if it is not
+present in the event sent by the opponent, the client should ignore any new
```
richvdh, line 198:
```suggestion
For backwards compatibility, if `sdp_stream_metadata` is not present in the
initial `m.call.invite` or `m.call.answer` [please link] event sent by the other party, 
the client should ignore any new
```

(I'm not entirely following why it is necessary to ignore new streams, but that is certainly what the MSC says)

richvdh:
closing in favour of #1735

PR #1604, opened 2023-07-26
Add room version 11

clokep:
Spec for MSC3820

Preview: https://pr1604--matrix-spec-previews.netlify.app

### DIFF
--- a/content/rooms/fragments/v11-redactions.md
+++ b/content/rooms/fragments/v11-redactions.md
@@ -1,0 +1,38 @@
+---
+---
+
+{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
+are no longer protected from redaction. The `m.room.create` event no longer keeps the `creator`
+property and now keeps the entire `content` property. The `m.room.redaction` event keeps the
+`redacts` property under `content`. The `m.room.power_levels` event keeps the `invite` property
+under `content`.
+
+The full redaction algorithm follows.
+
+Upon receipt of a redaction event, the server must strip off any keys
+not in the following list:
+
+-   `event_id`
+-   `type`
+-   `room_id`
+-   `sender`
+-   `state_key`
+-   `content`
+-   `hashes`
+-   `signatures`
+-   `depth`
+-   `prev_events`
+-   `auth_events`
+-   `origin_server_ts`
+
+The content object must also be stripped of all keys, unless it is one
+of the following event types:
+
+-   `m.room.member` allows keys `membership`, `join_authorised_via_users_server`.
+    Additionally, it allows the `signed` key of the `third_party_invite` key.
+-   `m.room.create` allows all keys.
+-   `m.room.join_rules` allows keys `join_rule`, `allow`.
+-   `m.room.power_levels` allows keys `ban`, `events`, `events_default`,
+    `invite`, `kick`, `redact`, `state_default`, `users`, `users_default`.
+-   `m.room.history_visibility` allows key `history_visibility`.
+-   `m.room.redaction` allows key `redacts`.
--- a/content/rooms/fragments/v11-redactions.md
+++ b/content/rooms/fragments/v11-redactions.md
@@ -2,9 +2,11 @@
 ---
 
 {{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
-are no longer protected from redaction. The `m.room.create` event now keeps the entire `content` property. The `m.room.redaction` event keeps the
-`redacts` property under `content`. The `m.room.power_levels` event keeps the `invite` property
-under `content`.
+are no longer protected from redaction. The [`m.room.create`](/client-server-api#mroomcreate)
+event now keeps the entire `content` property. The [`m.room.redaction`](/client-server-api#mroomredaction)
+event keeps the `redacts` property under `content`. The
+[`m.room.power_levels`](/client-server-api#mroompower_levels) event keeps the
+`invite` property under `content`.
 
 The full redaction algorithm follows.
 
@@ -27,11 +29,13 @@ not in the following list:
 The content object must also be stripped of all keys, unless it is one
 of the following event types:
 
--   `m.room.member` allows keys `membership`, `join_authorised_via_users_server`.
+-   [`m.room.member`](/client-server-api#mroommember) allows keys `membership`, `join_authorised_via_users_server`.
     Additionally, it allows the `signed` key of the `third_party_invite` key.
--   `m.room.create` allows all keys.
--   `m.room.join_rules` allows keys `join_rule`, `allow`.
--   `m.room.power_levels` allows keys `ban`, `events`, `events_default`,
-    `invite`, `kick`, `redact`, `state_default`, `users`, `users_default`.
--   `m.room.history_visibility` allows key `history_visibility`.
--   `m.room.redaction` allows key `redacts`.
+-   [`m.room.create`](/client-server-api#mroomcreate) allows all keys.
+-   [`m.room.join_rules`](/client-server-api#mroomjoin_rules) allows keys `join_rule`, `allow`.
+-   [`m.room.power_levels`](/client-server-api#mroompower_levels) allows keys
+    `ban`, `events`, `events_default`, `invite`, `kick`, `redact`, `state_default`,
+    `users`, `users_default`.
+-   [`m.room.history_visibility`](/client-server-api#mroomhistory_visibility)
+    allows key `history_visibility`.
+-   [`m.room.redaction`](/client-server-api#mroomredaction) allows key `redacts`.
--- a/content/rooms/fragments/v11-redactions.md
+++ b/content/rooms/fragments/v11-redactions.md
@@ -2,8 +2,7 @@
 ---
 
 {{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
-are no longer protected from redaction. The `m.room.create` event no longer keeps the `creator`
-property and now keeps the entire `content` property. The `m.room.redaction` event keeps the
+are no longer protected from redaction. The `m.room.create` event now keeps the entire `content` property. The `m.room.redaction` event keeps the
 `redacts` property under `content`. The `m.room.power_levels` event keeps the `invite` property
 under `content`.
 
--- a/content/rooms/fragments/v11-redactions.md
+++ b/content/rooms/fragments/v11-redactions.md
@@ -1,14 +1,3 @@
----
----
-
-{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
-are no longer protected from redaction. The [`m.room.create`](/client-server-api#mroomcreate)
-event now keeps the entire `content` property. The [`m.room.redaction`](/client-server-api#mroomredaction)
-event keeps the `redacts` property under `content`. The
-[`m.room.power_levels`](/client-server-api#mroompower_levels) event keeps the
-`invite` property under `content`.
-
-The full redaction algorithm follows.
 
 Upon receipt of a redaction event, the server must strip off any keys
 not in the following list:
--- a/content/rooms/v11.md
+++ b/content/rooms/v11.md
@@ -9,7 +9,8 @@ rules.
 
 ## Client considerations
 
-Clients should no longer depend on the `creator` property of `m.room.create` events.
+Clients should no longer depend on the `creator` property in the `content` of
+`m.room.create` events.
 
 Clients should note that the format of `m.room.redaction` events has been modified
 and look for the `redacts` key under `content` instead of a top-level event property.
@@ -17,8 +18,9 @@ and look for the `redacts` key under `content` instead of a top-level event prop
 Clients should note that the `third_party_invite` key of `m.room.member` events
 is no longer redacted, *but* will only contain the `signed` key after redaction.
 
-Clients which implement the redaction algorithm locally should refer to the
-[redactions](#redactions) section below for a full overview.
+### Redactions
+
+{{% rver-fragment name="v11-redactions" withVersioning="true" %}}
 
 ## Server implementation components
 
@@ -241,7 +243,7 @@ Some consequences of these rules:
 
 ### Redactions
 
-{{% rver-fragment name="v11-redactions" withVersioning="true" %}}
+[See above](#redactions).
 
 ## Unchanged from v10
 
--- a/content/rooms/v11.md
+++ b/content/rooms/v11.md
@@ -239,7 +239,7 @@ Some consequences of these rules:
 
 ### Redactions
 
-{{% rver-fragment name="v11-redactions" %}}
+{{% rver-fragment name="v11-redactions" withVersioning="true" %}}
 
 ## Unchanged from v10
 
--- a/content/rooms/v11.md
+++ b/content/rooms/v11.md
@@ -39,6 +39,10 @@ create `m.room.create` and `m.room.redaction` events.
 
 Room version 11 is based upon room version 10 with the following considerations.
 
+### Redactions
+
+[See above](#redactions).
+
 ### Event format
 
 The core event format is the same as [room version 10](/rooms/v10#event-format).
@@ -242,10 +246,6 @@ Some consequences of these rules:
     to both the kick *and* ban levels, *and* greater than the target
     user's power level.
 {{% /boxes/note %}}
-
-### Redactions
-
-[See above](#redactions).
 
 ## Unchanged from v10
 
--- a/content/rooms/v11.md
+++ b/content/rooms/v11.md
@@ -10,14 +10,15 @@ rules.
 ## Client considerations
 
 Clients should no longer depend on the `creator` property in the `content` of
-`m.room.create` events.  In all room versions, clients can rely on `sender`
-instead to determine a room creator.
+[`m.room.create`](/client-server-api#mroomcreate) events.  In all room versions,
+clients can rely on `sender` instead to determine a room creator.
 
-Clients should note that the format of `m.room.redaction` events has been modified
-and look for the `redacts` key under `content` instead of a top-level event property.
+Clients should note that the format of [`m.room.redaction`](/client-server-api#mroomredaction)
+events has been modified and look for the `redacts` key under `content` instead
+of a top-level event property.
 
-Clients should note that the `third_party_invite` key of `m.room.member` events
-is no longer redacted, *but* will only contain the `signed` key after redaction.
+Clients should note that the `third_party_invite` key of [`m.room.member`](/client-server-api#mroommember)
+events is no longer redacted, *but* will only contain the `signed` key after redaction.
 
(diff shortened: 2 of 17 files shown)


content/rooms/v11.md
```diff
@@ -0,0 +1,271 @@
...
+
+For backwards-compatibility with older clients, servers should add a `redacts` property
+to the top level of `m.room.redaction` events in when serving such events over the
+Client-Server API.
+
+For improved compatibility with newer clients, servers should add a `redacts` property
+to the `content` of `m.room.redaction` events in *older* room versions when serving
+such events over the Client-Server API.
+
+### Authorization rules
```
clokep, line 61: This is a copy of the room v10 ones with the references to `creator` of the `m.room.create` removed.

I debated about making this a separate file (and the room v10) one so that you can diff them more easily?

data/event-schemas/schema/m.room.create.yaml
```diff
@@ -6,7 +6,9 @@ properties:
   content:
     properties:
       creator:
-        description: The `user_id` of the room creator. This is set by the homeserver.
+        description: |-
```
clokep, line 9: There are probably examples also -- should those use the format of the "latest" room version? Showing both sounds complicated.

(This question goes for all modified events, which I think is just `m.room.create` and `m.room.redaction`)

content/rooms/fragments/v11-redactions.md
```diff
@@ -0,0 +1,38 @@
+---
+---
+
+{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
```
clokep, line 4: The `added-in` template doesn't seem to be rendering properly here?

content/rooms/fragments/v11-redactions.md
```diff
@@ -0,0 +1,38 @@
...
+are no longer protected from redaction. The `m.room.create` event no longer keeps the `creator`
+property and now keeps the entire `content` property. The `m.room.redaction` event keeps the
+`redacts` property under `content`. The `m.room.power_levels` event keeps the `invite` property
+under `content`.
```
clokep, line 8: The reviewer should go through this with a fine-tooth comb compared to the MSC3820's sub-MSCs. I think I got them all (and reflected the changes properly below), but it is easy to make a mistake.

data/event-schemas/schema/m.room.create.yaml
richvdh: best to remove it from the examples I think

data/event-schemas/schema/m.room.create.yaml
```diff
@@ -6,7 +6,9 @@ properties:
...
+          The `user_id` of the room creator. This is set by the homeserver for room versions 1 - 10. Room versions
+          11 no longer includes this property, the event `sender` should be used instead.
```
richvdh, line 11: comma splice. And I would say something like "only relevant in room versions 1-10."

data/event-schemas/schema/m.room.redaction.yaml
```diff
@@ -5,12 +5,15 @@ description: 'This event is created by the server to describe which event has be
...
+      redacts:
+        description: The event ID that was redacted. Starting in room version 11.
+        type: string
       reason:
         description: 'The reason for the redaction, if any.'
         type: string
     type: object
   redacts:
-    description: The event ID that was redacted.
+    description: The event ID that was redacted. For room versions 1 - 10.
```
richvdh, line 16:
```suggestion
    description: Only present in room versions 1 - 10. The event ID that was redacted.
```

data/event-schemas/schema/m.room.redaction.yaml
```diff
@@ -5,12 +5,15 @@ description: 'This event is created by the server to describe which event has be
 properties:
   content:
     properties:
+      redacts:
+        description: The event ID that was redacted. Starting in room version 11.
```
richvdh, line 9:
```suggestion
        description: The event ID that was redacted. Present starting in room version 11.
```

content/rooms/v11.md
```diff
@@ -0,0 +1,271 @@
...
+Room version 11 is based upon room version 10 with the following considerations.
+
+### Event format
+
+The event format is unchanged by this room version. See [below](#event-format-1)
+for details on the current event format.
+
+#### Remove the `creator` property of `m.room.create` events
+
+The `m.room.create` event no longer has a `creator` property, which previously
```
richvdh, line 45: a link to the definition of `m.room.create` in the C-S spec would be nice.

content/rooms/v11.md
```diff
@@ -0,0 +1,271 @@
...
+was always equivalent to the `sender` of the event.
+
+#### Moving the `redacts` property of `m.room.redaction` events to a `content` property
+
+The `redacts` property of `m.room.redaction` events is moved from a top-level
```
richvdh, line 50: ditto link to `m.room.redaction`

content/rooms/v11.md
```diff
@@ -0,0 +1,271 @@
...
+
+## Server implementation components
+
+{{% boxes/warning %}}
+The information contained in this section is strictly for server
+implementors. Applications which use the Client-Server API are generally
+unaffected by the intricacies contained here. The section above
+regarding client considerations is the resource that Client-Server API
+use cases should reference.
+{{% /boxes/warning %}}
```
richvdh, line 31: I... am not convinced this is correct. Much of the below is also relevant to clients, as evidenced by the fact the changes are mentioned in the "Client considerations" section.

Can we move the event format, and redaction rule, changes to a "common" section?

content/rooms/v11.md
```diff
@@ -0,0 +1,271 @@
...
+
+This room version updates the redaction algorithm and modifies how servers should
+create `m.room.create` and `m.room.redaction` events.
+
+Room version 11 is based upon room version 10 with the following considerations.
+
+### Event format
+
+The event format is unchanged by this room version. See [below](#event-format-1)
```
richvdh, line 40: It is incongruous that we start the section saying that events have not changed, and then go on to describe how they change.

Maybe: "The *core* event format is unchanged ... However, this room version changes the *content* of some event types".

richvdh: Also: https://pr1604--matrix-spec-previews.netlify.app/rooms/v11/#event-format-1 includes "redacts" at the top level, which *is* changing.

content/rooms/fragments/v11-redactions.md
```diff
@@ -0,0 +1,38 @@
+---
+---
+
+{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
+are no longer protected from redaction. The `m.room.create` event no longer keeps the `creator`
+property and now keeps the entire `content` property. The `m.room.redaction` event keeps the
```
richvdh, line 6: well, it does preserve `creator`, if there is one, because it preserves all the content.
```suggestion
are no longer protected from redaction. The `m.room.create` event now keeps the entire `content` property. The `m.room.redaction` event keeps the
```

content/rooms/fragments/v11-redactions.md
```diff
@@ -0,0 +1,38 @@
...
+of the following event types:
+
+-   `m.room.member` allows keys `membership`, `join_authorised_via_users_server`.
+    Additionally, it allows the `signed` key of the `third_party_invite` key.
+-   `m.room.create` allows all keys.
+-   `m.room.join_rules` allows keys `join_rule`, `allow`.
+-   `m.room.power_levels` allows keys `ban`, `events`, `events_default`,
+    `invite`, `kick`, `redact`, `state_default`, `users`, `users_default`.
+-   `m.room.history_visibility` allows key `history_visibility`.
+-   `m.room.redaction` allows key `redacts`.
```
richvdh, line 38: links to the event definitions would be :star2:

content/rooms/fragments/v11-redactions.md
richvdh: lgtm

richvdh (changes requested):

content/rooms/fragments/v11-redactions.md
richvdh: https://github.com/matrix-org/matrix-spec/pull/1604#discussion_r1287317592

data/event-schemas/schema/m.room.create.yaml
clokep: Updating the examples makes sense, should we also update the schemas or show both?

content/rooms/v11.md
clokep: I believe this is shown on every room version and I just copied and pasted. Will check.

richvdh: I mean sure, and that was probably the right thing to do for some of the early room versions, but that doesn't mean that cutting-and-pasting it every time is still the right thing to do.

clokep: Weirdly this section seems to mostly duplicate what is written for the clients above though? So I think the idea is that client authors shouldn't need to care about this section since the *above* section provides all the needed info.

richvdh: yeah but the above section *doesn't*. It's currently either a very high-level summary, or (in the case of redactions), a link into this section which it says you shouldn't read if you're a client author.

content/rooms/v11.md
clokep: Right. I'm trying to get rid of that now, but it is proving difficult!

clokep: I think I cleaned this up. I had to copy & paste a bit with the event formats since you can't seem to remove fields from a JSONschema?

content/rooms/fragments/v11-redactions.md
clokep: I guess we should do this for all room versions? I might do it as a separate PR.

data/api/server-server/definitions/pdu_v11.yaml
```diff
@@ -0,0 +1,81 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+type: object
+title: Persistent Data Unit
+description: A persistent data unit (event) for room version 4 and beyond.
+example:
+  $ref: "../examples/pdu_v11.json"
+allOf:
+  # v11 is the v4 event, but without redacts. Copy the auth_events/prev_events
+  # from pdu_v4.yaml and hashes and signatures from pdu_v3.yaml.
```
clokep, line 21: I debated making a `pdu_v3_hash_sigs.yaml` and `pdu_v4_auth_prev_events.yaml` and using each twice, but that seemed like overkill? WDYT?

content/rooms/fragments/v11-redactions.md
turt2live: yea, this would be auxillary PR material imo

content/rooms/v11.md
```diff
@@ -0,0 +1,269 @@
...
+type: docs
+weight: 100
+---
+
+This room version builds on [version 10](/rooms/v10) while clarifying redaction
+rules.
+
+## Client considerations
+
+Clients should no longer depend on the `creator` property of `m.room.create` events.
```
turt2live, line 12:
```suggestion
Clients should no longer depend on the `creator` property of `m.room.create` events.
In all room versions, clients can rely on `sender` instead to determine a room creator.
```

content/rooms/v11.md
turt2live: This warning box is intentionally part of the copy/paste template, mandated by documentation style: https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#room-versions . The intention is to prevent client developers from reading "below the fold".

The redactions section actually needs to be duplicated up above, per v9's template (and doc style)

content/rooms/v11.md
```diff
@@ -0,0 +1,269 @@
...
+
+-   Unless you are a member of the room, the only permitted operations
+    (apart from the initial create/join) are: joining a public room;
+    accepting or rejecting an invitation to a room.
+-   To unban somebody, you must have power level greater than or equal
+    to both the kick *and* ban levels, *and* greater than the target
+    user's power level.
+{{% /boxes/note %}}
+
+### Redactions
```
turt2live, line 242: from my other comment: duplicate this up above in Client Considerations

(doing so would also comply with the documentation style: https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#room-versions )

turt2live (commented):
generally lgtm - I haven't checked against the MSCs, but assume that's happened in another review (or will happen). This review is primarily for documentation style consistency.

content/rooms/v11.md
turt2live: (also it needs to be re-ordered here to match the documentation style)

content/rooms/v11.md
```diff
@@ -241,7 +243,7 @@ Some consequences of these rules:
 
 ### Redactions
 
-{{% rver-fragment name="v11-redactions" withVersioning="true" %}}
+[See above](#redactions).
```
turt2live, line 247: (it's specifically supposed to be duplicated, not copied - this is where use of fragments comes into play)

content/rooms/fragments/v11-redactions.md
clokep: Actually this was easy with some regex find and replace.

content/rooms/v11.md
clokep: I copied what room version 9 does: https://spec.matrix.org/v1.7/rooms/v9/#redactions-1 I guess I'll fix that too?

clokep: Actually, all the room versions do this:

https://spec.matrix.org/v1.7/rooms/v1/#redactions-1

https://spec.matrix.org/v1.7/rooms/v6/#redactions-1

clokep: Let me know if you think I should update them all or leave them all. 👍

clokep:
I think the remaining bits are if we want to edit the examples here (see https://github.com/matrix-org/matrix-spec/pull/1604#discussion_r1275103155); if we need to insert the redactions twice (see https://github.com/matrix-org/matrix-spec/pull/1604#discussion_r1287634080).

From my POV the warning box question (https://github.com/matrix-org/matrix-spec/pull/1604#discussion_r1287350165) is resolved with keeping it, but shout if there are changes to make.

content/rooms/v11.md
turt2live: oh, let's just do what v9 and the others do.

turt2live (approved):
It fits the documentation style requirements

Thanks!

content/rooms/v11.md
```diff
@@ -0,0 +1,273 @@
...
+Clients should no longer depend on the `creator` property in the `content` of
+[`m.room.create`](/client-server-api#mroomcreate) events.  In all room versions,
+clients can rely on `sender` instead to determine a room creator.
+
+Clients should note that the format of [`m.room.redaction`](/client-server-api#mroomredaction)
+events has been modified and look for the `redacts` key under `content` instead
+of a top-level event property.
+
+Clients should note that the `third_party_invite` key of [`m.room.member`](/client-server-api#mroommember)
+events is no longer redacted, *but* will only contain the `signed` key after redaction.
```
richvdh, line 21: The way this is structured still makes me feel like this is an introduction, and that the real changes are going to be spelt out elsewhere.

Two things would help:

1. Add an "Event format" subheading.
2. Delete "Clients should note that".

content/rooms/fragments/v11-redactions.md
```diff
@@ -0,0 +1,41 @@
...
+---
+
+{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
+are no longer protected from redaction. The [`m.room.create`](/client-server-api#mroomcreate)
+event now keeps the entire `content` property. The [`m.room.redaction`](/client-server-api#mroomredaction)
+event keeps the `redacts` property under `content`. The
+[`m.room.power_levels`](/client-server-api#mroompower_levels) event keeps the
+`invite` property under `content`.
+
+The full redaction algorithm follows.
```
richvdh, line 11: These two paragraphs won't be appropriate for inclusion in room v12 (since they document the changes since room v10). Can you move them into the main `rooms/v11.md` file?

(Looks like room v9/v10 messed this up. Room v6/v7 did better: there is a paragraph in https://spec.matrix.org/v1.7/rooms/v6/#redactions which is not present in https://spec.matrix.org/v1.7/rooms/v7/#redactions)

data/event-schemas/schema/m.room.create.yaml
```diff
@@ -6,7 +6,9 @@ properties:
   content:
     properties:
       creator:
```
richvdh, line 8: needs removing from the `required` list.

data/event-schemas/schema/m.room.create.yaml
richvdh: I don't think I understand the question?

clokep: There's both an event schema and an example event for each event type; do we want to update both?

clokep: Never mind, I think we're *only* talking about the examples. Sorry.

content/rooms/fragments/v11-redactions.md
clokep: I also fixed room v9/v10.

data/event-schemas/schema/m.room.create.yaml
clokep: Also applied the same for `redacts`.

data/event-schemas/schema/m.room.create.yaml
```diff
@@ -6,7 +6,9 @@ properties:
...
-        description: The `user_id` of the room creator. This is set by the homeserver.
+        description: |-
+          The `user_id` of the room creator. Only present in room versions 1 - 10. Starting with
```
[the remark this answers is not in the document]
clokep: I applied similar languages for the redaction event.

content/rooms/v11.md
```diff
@@ -0,0 +1,284 @@
...
+---
+
+This room version builds on [version 10](/rooms/v10) while clarifying redaction
+rules.
+
+## Client considerations
+
+### Redactions
+
+{{% added-in this=true %}} The top-level `origin`, `membership`, and `prev_state` properties
```
richvdh, line 14: This isn't rendering right: the changes are in a separate paragraph to the "[New in this version]" label.

I think you want:

```suggestion
{{< added-in this=true >}} The top-level `origin`, `membership`, and `prev_state` properties
```

(cf https://gohugo.io/content-management/shortcodes/#shortcodes-without-markdown)

richvdh: ditto in v9, if you fancy fixing that while you're here

richvdh (approved):
lgtm otherwise. Thanks!

content/rooms/v11.md
clokep: I fixed it for v6, v9, and v11.

PR #1618, opened 2023-08-18
Clarify that `via` is required for `m.space.parent` and `m.space.child` as per MSC1772.

PaarthShah:
Per [MSC1772](https://github.com/matrix-org/matrix-spec-proposals/blob/6a3ab1d64cdf9f429a86b0e9d4df927dd36f0280/proposals/1772-groups-as-rooms.md) (emphasis mine):

- `m.space.child`
> The admins of a space can advertise rooms and subspaces for their space by setting m.space.child state events. The state_key is the ID of a child room or space, and the content **_must_** contain a via key which gives a list of candidate servers that can be used to join the room.

- `m.space.parent`
> Similar to m.space.child, the state_key is the ID of the parent space, and the content **_must_** contain a via key which gives a list of candidate servers that can be used to join the parent.

I was made aware of this by [this issue](https://github.com/poljar/matrix-nio/issues/433).

Preview: https://pr1618--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1618.clarification
+++ b/changelogs/client_server/newsfragments/1618.clarification
@@ -1,0 +1,1 @@
+Clarify that `via` property is required for `m.space.parent` and `m.space.child` as per MSC1772. Contributed by @PaarthShah
--- a/changelogs/client_server/newsfragments/1618.clarification
+++ b/changelogs/client_server/newsfragments/1618.clarification
@@ -1,1 +1,1 @@
-Clarify that `via` property is required for `m.space.parent` and `m.space.child` as per MSC1772. Contributed by @PaarthShah
+Clarify that the `via` property is required for `m.space.parent` and `m.space.child` as per MSC1772. Contributed by @PaarthShah.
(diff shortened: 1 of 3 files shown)


data/event-schemas/schema/m.space.child.yaml
```diff
@@ -9,8 +9,6 @@ properties:
         type: array
         description: |-
           A list of servers to try and join through. See also: [Routing](/appendices/#routing).
-
-          When not present or invalid, the child room is not considered to be part of the space.
```
turt2live, line 13: While the fields are required, the text being removed here is critical for operation of the feature. This is why when the spec PR was originally written it did not flag the fields as required.

turt2live (dismissed):

changelogs/client_server/newsfragments/1618.clarification
```diff
@@ -0,0 +1 @@
+Clarify that `via` property is required for `m.space.parent` and `m.space.child` as per MSC1772. Contributed by @PaarthShah
```
anoadragon453, line 1:
```suggestion
Clarify that the `via` property is required for `m.space.parent` and `m.space.child` as per MSC1772. Contributed by @PaarthShah.
```

anoadragon453 (approved):
LGTM, thank you for catching and acting on this!

PR #1629, opened 2023-08-30
strike element is deprecated, s replaces it

cloudrac3r:
### Pull Request Checklist

### Description

The [strike](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike) element is deprecated in the HTML spec. Clients should prefer [s](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s) instead if they wish to mark text with strikethrough without semantic meaning.

As far as I am aware, no client actually supports sending messages with strikethrough formatting, so this shouldn't affect current compatibility.

I hope I filled out this PR template correctly :)

Signed-off-by: Cadence Ember <cadence@disroot.org>

Preview: https://pr1629--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1629.clarification
+++ b/changelogs/client_server/newsfragments/1629.clarification
@@ -1,0 +1,1 @@
+The [strike](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strike) element is deprecated in the HTML spec. Clients should prefer [s](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/s) instead.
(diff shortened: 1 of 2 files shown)


KitsuneRal:
Well, in fact, a few clients support sending raw HTML (I just checked Quaternion and Element Web for a very quick check), and although Quaternion still converts `<strike>` to `<del>` even when sent as raw HTML, Element Web doesn't do that and faithfully sends `<strike>` to the homeserver. I'm afraid this would require a (very simple) MSC therefore.

MTRNord:
I think the current state is more confusing considering the syntax ele Web input for example uses is `<del>` and not any of the others.

KitsuneRal:
Well it's the same old HTML story - `<strong>` vs. `<b>` etc. `<del>` is in the recommended list of tags just as well; even though the tags are sort-of similar semantically, replacing `<strike>` with `<s>` is orthogonal to having `<del>` around.

richvdh:
Closing this for now. As others have noted, we'll need an MSC.

anoadragon453:
Re-opening as per @cloudrac3r's [request](https://github.com/matrix-org/matrix-spec/pull/1629#issuecomment-1963109725).

turt2live (approved):
Thanks!

PR #1633, opened 2023-09-12
Replace all mentions of Swagger by OpenAPI

zecakeh:
Since OpenAPI 3.1 is not compatible with Swagger anymore, it might be confusing to keep the old name or the double naming.

The last change that renames the scripts might be more controversial, as it might be confusing for people used to the old script names, so it can be reverted if wanted.

Preview: https://pr1633--matrix-spec-previews.netlify.app


anoadragon453 (approved):
I'm happy with this as well, and it makes sense if the Swagger naming is no longer technically true.

I don't have any muscle memory for these scripts, so renaming them does not affect me. As long as any of our tooling that uses them is still happy, I'm happy.

README.md
```diff
@@ -88,12 +88,12 @@ spec to `/spec`. If you'd like to serve the spec off a path instead of a domain
 to the `hugo -d "spec"` command.
 
 For building the OpenAPI definitions, create a python3 virtualenv and activate it. Then run `pip install -r ./scripts/requirements.txt`
-and finally `python ./scripts/dump-swagger.py` to generate it to `./scripts/swagger/api-docs.json`. To make use of the generated file,
+and finally `python ./scripts/dump-openapi.py` to generate it to `./scripts/openapi/api-docs.json`. To make use of the generated file,
```
richvdh, line 91: looks like dump-openapi still writes to `./scripts/swagger`: https://github.com/matrix-org/matrix-spec/blob/217fbbeab11645bfecfc1c41a48dc00a7a7da054/scripts/dump-openapi.py#L116

scripts/openapi-http-server.py
```diff
@@ -41,13 +41,13 @@ def send_my_headers(self):
...
-        'swagger_dir', nargs='?',
-        default=os.path.join(scripts_dir, 'swagger'),
+        'openapi_dir', nargs='?',
+        default=os.path.join(scripts_dir, 'openapi'),
         help='directory to serve (default: %(default)s)',
     )
     args = parser.parse_args()
 
-    os.chdir(args.swagger_dir)
+    os.chdir(args.openapi)
```
richvdh, line 50: `openapi_dir`

richvdh (approved):
thanks!

PR #1634, opened 2023-09-12
Fix enum types in JSON schemas

zecakeh:
`type: enum` does not exist, as an enum can be of any type.

Preview: https://pr1634--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1634.clarification
+++ b/changelogs/internal/newsfragments/1634.clarification
@@ -1,0 +1,1 @@
+Fix enum types in JSON schemas.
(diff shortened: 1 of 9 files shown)


KitsuneRal (approved):

KitsuneRal:
That looks good, but should we also show `string` instead of `enum` in the spec text then? For consistency.

zecakeh:
Sure we could do that, it should be simple enough. Do you want me to add a commit to this PR or should it be in another PR?

KitsuneRal:
Just add another commit here.

richvdh (approved):
lgtm but maybe @KitsuneRal wants another look?

richvdh:
Assuming kitsune is happy here.

PR #1635, opened 2023-09-12
Fix schema of `m.mentions` object

zecakeh:
That was not how a string of arrays is declared. Also the example was for a full event, instead of just the `m.mentions` object.

Preview: https://pr1635--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/m.mentions.yaml
+++ b/data/api/client-server/definitions/m.mentions.yaml
@@ -28,7 +28,9 @@ example: {
 }
 properties:
   user_ids:
-    type: string[]
+    type: array
+    items:
+      type: string
     description: A list of Matrix IDs of mentioned users.
   room:
     type: boolean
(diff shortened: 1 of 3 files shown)


data/api/client-server/definitions/m.mentions.yaml
```diff
@@ -18,17 +18,13 @@ description: |-
   Describes whether the event mentions other users or the room. This is contained
   within the event's `content` alongside other fields for the relevant event type.
 example: {
```
clokep, line 20: My thinking when doing this was to keep the example the same across spec v1.6 and 1.7, see:

* https://spec.matrix.org/v1.6/client-server-api/#user-and-room-mentions
* https://spec.matrix.org/v1.7/client-server-api/#definition-mmentions
* https://pr1635--matrix-spec-previews.netlify.app/client-server-api/#definition-mmentions

zecakeh: I see. I guess that example should be included directly in the spec rather than in the JSON schema, then?

The reason I stumbled upon this is that I'd like to add a tool that validates all JSON schemas in CI. This one would not pass as the example doesn't match the schema.

clokep: That might be better, yeah.

richvdh (approved):
lgtm, thanks!

PR #1636, opened 2023-09-12
Fix schema of `m.receipt` EDU

zecakeh:
It was missing a nesting level for the user ID -> user read receipt map.

Preview: https://pr1636--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/server-server/definitions/event-schemas/m.receipt.yaml
+++ b/data/api/server-server/definitions/event-schemas/m.receipt.yaml
@@ -41,7 +41,9 @@ allOf:
             # on. At that point, m.read can become optional (maybe).
             "m.read":
               type: object
-              description: Read receipts for users in the room.
+              description: |-
+                Read receipts for users in the room. The string key is the user
+                ID the receipt belongs to.
               additionalProperties:
                 type: object
                 title: User Read Receipt


richvdh:
A few links:
 * The corrected rendered version is at https://pr1636--matrix-spec-previews.netlify.app/server-server-api/#definition-mreceipt
 * The incorrect version is at https://spec.matrix.org/v1.8/server-server-api/#definition-mreceipt. (Though the example is correct.)
 * It seems to have been wrong since forever (it was introduced in https://github.com/matrix-org/matrix-spec/commit/e03bfbc47b5079a7bd54bbce1a4996524b6fc617)

changelogs/internal/newsfragments/1636.clarification
```diff
@@ -0,0 +1 @@
+Fix schema of `m.receipt` EDU.
```
richvdh, line 1: this isn't internal - it's part of the server-server api.

data/api/server-server/definitions/event-schemas/m.receipt.yaml
```diff
@@ -42,40 +42,42 @@ allOf:
...
-                      format: int64
-                      description: |-
-                        A POSIX timestamp in milliseconds for when the user read
-                        the event specified in the read receipt.
-                      example: 1533358089009
-                    thread_id:
-                        type: string
-                        x-addedInMatrixVersion: "1.4"
+              additionalProperties:
+                type: object
```
richvdh, line 46: I think we should document what the key of the object is (the user id), much as we do for the `content` object above it: https://github.com/matrix-org/matrix-spec/blob/9c995dfcb6520d40372e7f5ee7004de001aefc7d/data/api/server-server/definitions/event-schemas/m.receipt.yaml#L34-L35

richvdh (changes requested):

changelogs/internal/newsfragments/1636.clarification
zecakeh: The changelog was moved

richvdh (approved):
thanks!

PR #1638, opened 2023-09-14
add "case sensitive" to `/publicRooms`

### LINKED ISSUE #1619: Server names are case sensitive in /publicRooms
The spec should mention that the server field of GET/PUT /publicRooms is case sensitive. https://spec.matrix.org/v1.7/client-server-api/#get_matrixclientv3publicrooms
The client should lowercase it before sending. This probably effects other endpoints too.

Alternatively, servers can lowercase it or the entire spec can be overhauled to make server names case insensitive everywhere.

mohit1523:
Fixes #1619 

This PR add the "case sensitive" to the server specs of /publicRooms so that client should lowercase it before sending.

Screenshot 2023-09-14 233219

Preview: https://pr1638--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1638.feature
+++ b/changelogs/client_server/newsfragments/1638.feature
@@ -1,1 +1,1 @@
-Adds "case sensitive" to the /publicRooms 
+Add a note to the `/publicRooms` API that the server name is case sensitive. 
--- a/changelogs/client_server/newsfragments/1638.feature
+++ b/changelogs/client_server/newsfragments/1638.feature
@@ -1,0 +1,1 @@
+Adds "case sensitive" to the /publicRooms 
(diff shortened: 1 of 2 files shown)


changelogs/client_server/newsfragments/1638.feature
```diff
@@ -0,0 +1 @@
+Adds "case sensitive" to the /publicRooms
```
richvdh, line 1:
```suggestion
Add a note to the `/publicRooms` API that the server name is case sensitive. 
```

richvdh (approved):
thanks

richvdh:
I'm supposed to ask you for [sign off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off) but since it's only 4 words we'll take the risk that you won't sue us

mohit1523:
> I'm supposed to ask you for [sign off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off) but since it's only 4 words we'll take the risk that you won't sue us

Thanks to you sir 😄

PR #1639, opened 2023-09-18
Clarify that an `m.room.name` event with an absent `name` field is not expected behavior

### LINKED ISSUE #1632: `m.room.name`'s `name` field is marked as required but is described as possibly absent
**Link to problem area**: https://spec.matrix.org/v1.8/client-server-api/#mroomname

**Issue**
The schema of the event says that the `name` field is required. The description of the event says that the same field can be "absent, null, or empty".

**Expected behaviour**
It should either be required or not.

For context, Synapse doesn't accept that a client sends this event if the field is not a string.

zecakeh:
Fixes #1632

For me the "if" makes a lot of difference in understanding that the paragraph does not mean that it should be the expected behavior.

Preview: https://pr1639--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1639.clarification
+++ b/changelogs/client_server/newsfragments/1639.clarification
@@ -1,0 +1,1 @@
+Clarify that an `m.room.name` event with an absent `name` field is not expected behavior.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
much clearer, thanks!

PR #1644, opened 2023-09-22
Note that /logout doesn't take a body

reivilibre:
`/logout` says it does not take a request body[^1], so add it to the list of exceptions at the top.

I note that Complement does not send a request body for this endpoint, so that must mean Synapse and Dendrite allow that.

[^1]: https://spec.matrix.org/v1.8/client-server-api/#post_matrixclientv3logout

Preview: https://pr1644--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1644.clarification
+++ b/changelogs/client_server/newsfragments/1644.clarification
@@ -1,0 +1,1 @@
+Clarify specification by adding `/logout` to the overview list of endpoints that don't take a request body.
--- a/changelogs/client_server/newsfragments/1644.clarification
+++ b/changelogs/client_server/newsfragments/1644.clarification
@@ -1,1 +1,1 @@
-Clarify specification by adding `/logout` to the overview list of endpoints that don't take a request body.
+Clarify specification by adding `/logout` to the overview list of endpoints that don't take a JSON request body.
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -29,12 +29,12 @@ JSON object.  Clients should supply a `Content-Type` header of
 
 The exceptions are:
 
-- [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload),
-  which takes any type of request body depending on what type of media is to be uploaded.
-- [`PUT /_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
-  which takes any type of request body depending on what type of media is to be uploaded.
-- [`POST /_matrix/client/v3/logout`](#post_matrixclientv3logout),
-  which does not take any request body.
+- [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload) and 
+  [`PUT /_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
+  both of which take the uploaded media as the request body.
+- [`POST /_matrix/client/v3/logout`](#post_matrixclientv3logout) and
+  [`POST /_matrix/client/v3/logout/all`](#post_matrixclientv3logoutall),
+  which take an empty request body.
 
 Similarly, all endpoints require the server to return a JSON object,
 with the exception of 200 responses to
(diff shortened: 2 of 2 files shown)


changelogs/client_server/newsfragments/1644.clarification
```diff
@@ -0,0 +1 @@
+Clarify specification by adding `/logout` to the overview list of endpoints that don't take a request body.
```
richvdh, line 1: Note that the other two in the list *do* take a request body; it's just not a JSON object. Could you reword slightly?

content/client-server-api/_index.md
```diff
@@ -23,8 +23,9 @@ Clients are authenticated using opaque `access_token` strings (see [Client
 Authentication](#client-authentication) for details).
 
 All `POST` and `PUT` endpoints, with the exception of [`POST
-/_matrix/media/v3/upload`](#post_matrixmediav3upload) and [`PUT
-/_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
+/_matrix/media/v3/upload`](#post_matrixmediav3upload), [`PUT
+/_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid)
+and [`POST /_matrix/client/v3/logout`](#post_matrixclientv3logout),
```
richvdh, line 28: it might be worth emphasising what the exceptions do require.

richvdh (changes requested):
Hrm, this makes me sad.

I *want* to claim that this is misleading wording on behalf of the `/logout` endpoint definition, and a misfeature in complement/synapse/dendrite, rather than an indication that the body should actually allowed to be empty.  Granting `/logout` an exception in this way very much puts it in a class of one.

On the other hand, it's clearly been this way since forever, and Element-Web relies upon it. So I guess I have to suck up the fact that there's an exception and make it something to fix for OIDC.

Anyway, `/logout/all` has the same problem I think.

richvdh:
(we should file a 'Wart' issue against the spec for this)

richvdh:
@reivilibre would you be able to update this?

reivilibre:
Issue created at https://github.com/matrix-org/matrix-spec/issues/1845 but I don't have the ability to add a Wart label

content/client-server-api/_index.md
```diff
@@ -22,13 +22,20 @@ recommended outside test environments.
...
 `application/json` for all requests with JSON bodies, but this is not required.
 
+The exceptions are:
+
+- [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload),
+  which takes any type of request body depending on what type of media is to be uploaded.
+- [`PUT /_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
+  which takes any type of request body depending on what type of media is to be uploaded.
+- [`POST /_matrix/client/v3/logout`](#post_matrixclientv3logout),
+  which does not take any request body.
```
richvdh, line 37: logout/all has the same problem:

```suggestion
- [`POST /_matrix/client/v3/logout`](#post_matrixclientv3logout) and
  [`POST /_matrix/client/v3/logout/all`](#post_matrixclientv3logoutall),
  which take an empty request body.
```

content/client-server-api/_index.md
```diff
@@ -22,13 +22,20 @@ recommended outside test environments.
...
 require the client to supply a request body containing a (potentially empty)
 JSON object.  Clients should supply a `Content-Type` header of
 `application/json` for all requests with JSON bodies, but this is not required.
 
+The exceptions are:
+
+- [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload),
+  which takes any type of request body depending on what type of media is to be uploaded.
+- [`PUT /_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
+  which takes any type of request body depending on what type of media is to be uploaded.
```
richvdh, line 35: I feel like this is getting a bit verbose. (And sorry, I know I previously encouraged you to spell this out more.)

Can we combine them into a single entry:

```suggestion
- [`POST /_matrix/media/v3/upload`](#post_matrixmediav3upload) and 
  [`PUT /_matrix/media/v3/upload/{serverName}/{mediaId}`](#put_matrixmediav3uploadservernamemediaid),
  both of which take the uploaded media as the request body.
```

richvdh (approved):
Thanks!

PR #1646, opened 2023-09-26
Remove required 'fieldname' in Protocol definition

zecakeh:
This field does not exist and makes example validation fail. It does not change anything in the rendered spec.

My guess is that it was trying to mean that for every field in `user_fields` and `location_fields`, there should be an entry in `field_types`. It might be interesting to see if it is possible to define that using JSON schema.

Preview: https://pr1646--matrix-spec-previews.netlify.app


richvdh (approved):
thanks!

PR #1647, opened 2023-09-26
Fix schemas used for account data and presence events in `GET /initialSync`

zecakeh:
Otherwise the example will fail to validate against the schema.

These schemas are chosen because these events don't have an event ID.

Preview: https://pr1647--matrix-spec-previews.netlify.app


richvdh (approved):
Thanks!

PR #1665, opened 2023-10-20
server-server-api: Fix grammar

progval:
Preview: https://pr1665--matrix-spec-previews.netlify.app

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -187,7 +187,7 @@ to send. The process overall is as follows:
     for `<hostname>`.
 
 6.  If the `/.well-known` request returned an error response, and the
-    SRV records were not found, an IP address is resolved using CNAME, AAAA and A
+    no SRV records were found, an IP address is resolved using CNAME, AAAA and A
     records. Requests are made to the resolved IP address using port
     8448 and a `Host` header containing the `<hostname>`. The target
     server must present a valid certificate for `<hostname>`.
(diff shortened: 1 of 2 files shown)


content/server-server-api.md
```diff
@@ -187,7 +187,7 @@ to send. The process overall is as follows:
     for `<hostname>`.
 
 6.  If the `/.well-known` request returned an error response, and the
-    SRV record was not found, an IP address is resolved using CNAME, AAAA and A
+    SRV records were not found, an IP address is resolved using CNAME, AAAA and A
```
MTRNord, line 190: Isn't matrix set up in a way where there only is one SRV record at any given time? (Though on a quick search I didnt see anything not allowing multiple)

uhoreg: Standard SRV resolution allows for multiple records, and Matrix never forbade it.  But also, steps 4 and 5 refer to different SRV records (`_matrix-fed._tcp.<hostname>` and `_matrix._tcp.<hostname>`), so it really should be plural.

On the other hand "the SRV records were not found" sounds like SRV records exist but could not be found for whatever reason.  So maybe something like "... and no SRV records are present ...", or "... and no SRV records are found ..." might be better?

content/server-server-api.md
```diff
@@ -187,7 +187,7 @@ to send. The process overall is as follows:
     for `<hostname>`.
 
 6.  If the `/.well-known` request returned an error response, and the
```
uhoreg, line 189:
```suggestion
6.  If the `/.well-known` request returned an error response, and
```

uhoreg (approved):
Thanks!

PR #1676, opened 2023-11-21
fallback keys should have a "fallback: true" property

uhoreg:
This seems to have been forgotten in the original spec PR for fallback keys.

ref: https://github.com/matrix-org/matrix-spec-proposals/pull/2732/files#diff-e8bbdeed17d423994b6033c3bc7db805cbe334d0946e5e86c755f051f6871459R21-R23


turt2live (approved):
Thanks!

PR #1677, opened 2023-11-22
Make clear the thread root is not in the thread

andybalaam:
This was originally written up as [MSC4037](https://github.com/matrix-org/matrix-spec-proposals/pull/4037), but discussion there concluded that a spec PR was more appropriate.

Quoting @turt2live in that discussion:

> [MSC3771](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3771-read-receipts-for-threads.md) (merged) originally introduced the threaded receipts behaviour into the spec in v1.4, but did not describe what actually constitutes what is "in a thread". Though, the implied wording of MSC3771 does *not* include the thread root in the thread for purposes of notifications. This is further backed up by Synapse being the primary cited implementation proof for MSC3771, which is backed up by this proposal's text in the above paragraph.

This PR modifies the spec to make clear that thread roots and their non-thread children are not "in the thread" for the purposes of receipt handling and unreadness.

* [Synapse](https://github.com/matrix-org/synapse/blob/v1.87.0/synapse/rest/client/receipts.py#L116-L154) has the behaviour specified in this PR.

* [matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/blob/develop/src/client.ts#L5004) was recently fixed to match the behaviour required by this PR, as a bug fix for stuck notifications bugs resulting from inconsistency between Synapse and Element Web.

Signed-off-by: Andy Balaam <andy.balaam@matrix.org>

Preview: https://pr1677--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/receipts.md
+++ b/content/client-server-api/modules/receipts.md
@@ -137,16 +137,22 @@ either a thread root's event ID or `main` for the main timeline.
 
 Threading introduces a concept of multiple conversations being held in the same
 room and thus deserve their own read receipts and notification counts. An event is
-considered to be "in a thread" if it meets any of the following criteria:
-* It has a `rel_type` of `m.thread`.
-* It has child events with a `rel_type` of `m.thread` (in which case it'd be the
-  thread root).
-* Following the event relationships, it has a parent event which qualifies for
-  one of the above. Implementations should not recurse infinitely, though: a
-  maximum of 3 hops is recommended to cover indirect relationships.
+considered to be "in a thread" if:
 
-Events not in a thread but still in the room are considered to be part of the
-"main timeline", or a special thread with an ID of `main`.
+* It has a rel_type of m.thread, or
+* It has a parent event with this rel_type, or a parent of a parent, or further
+  up the chain of relations. (Implementations should not recurse to arbitrary
+  depth: a maximum of 3 hops is recommended to cover indirect relationships.)
+
+Events not in a thread but still in the room are considered to be in the "main
+timeline". Threaded receipts for the main timeine use a special thread ID of
+`main`.
+
+Thread roots are considered to be in the main timeline, as are events that are
+related to a thread root via non-thread relations. Note that clients providing a
+single-thread view will probably want to include thread roots and some of their
+child events (e.g. reactions) in that view, even though from a receipt point of
+view they are not part of that thread.
 
 The following is an example DAG for a room, with dotted lines showing event
 relationships and solid lines showing topological ordering.
--- a/content/client-server-api/modules/receipts.md
+++ b/content/client-server-api/modules/receipts.md
@@ -150,10 +150,7 @@ timeline". When referring to the main timeline as a thread (e.g. in receipts
 and notifications counts) a special thread ID of `main` is used.
 
 Thread roots are considered to be in the main timeline, as are events that are
-related to a thread root via non-thread relations. Note that clients providing a
-single-thread view will probably want to include thread roots and some of their
-child events (e.g. reactions) in that view, even though from a receipt point of
-view they are not part of that thread.
+related to a thread root via non-thread relations.
 
 The following is an example DAG for a room, with dotted lines showing event
 relationships and solid lines showing topological ordering.
--- a/content/client-server-api/modules/receipts.md
+++ b/content/client-server-api/modules/receipts.md
@@ -145,8 +145,8 @@ considered to be "in a thread" if:
   depth: a maximum of 3 hops is recommended to cover indirect relationships.)
 
 Events not in a thread but still in the room are considered to be in the "main
-timeline". Threaded receipts for the main timeline use a special thread ID of
-`main`.
+timeline". When referring to the main timeline as a thread (e.g. in receipts
+and notifications counts) a special thread ID of `main` is used.
 
 Thread roots are considered to be in the main timeline, as are events that are
 related to a thread root via non-thread relations. Note that clients providing a
--- a/content/client-server-api/modules/receipts.md
+++ b/content/client-server-api/modules/receipts.md
@@ -145,7 +145,7 @@ considered to be "in a thread" if:
   depth: a maximum of 3 hops is recommended to cover indirect relationships.)
 
 Events not in a thread but still in the room are considered to be in the "main
-timeline". Threaded receipts for the main timeine use a special thread ID of
+timeline". Threaded receipts for the main timeline use a special thread ID of
 `main`.
 
 Thread roots are considered to be in the main timeline, as are events that are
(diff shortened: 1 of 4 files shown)


andybalaam:
I believe this should have the spec-bug label but I don't have permission to add it.

andybalaam:
I think the changelog entry should be `.bugfix` but the validator appears to reject items with that name?

richvdh:
> I think the changelog entry should be `.bugfix` but the validator appears to reject items with that name?

We use different suffixes in spec to other projects: https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog. `clarification` is right here, if the argument is just that the spec was "misleading".

As for labels: we don't normally bother with them on PRs for the spec.

andybalaam:
@richvdh I was going by:

> Likewise, corrections to the specification, to fix situations where, in practice, servers and clients behave differently to the specification, including anything with the [spec-bug](https://github.com/matrix-org/matrix-spec/issues?q=is%3Aopen+label%3Aspec-bug+sort%3Aupdated-desc) label.

From https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#other-changes

I guess this does fit under "clarification" from https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst?rgh-link-date=2023-11-22T11%3A20%3A43Z#adding-to-the-changelog but if a `.bugfix` existed I think it would fit better, since we are correcting the spec to conform with implementations, because those implementations do what we meant to say first time.

andybalaam:
> @richvdh I was going by:
> 
> > Likewise, corrections to the specification, to fix situations where, in practice, servers and clients behave differently to the specification, including anything with the [spec-bug](https://github.com/matrix-org/matrix-spec/issues?q=is%3Aopen+label%3Aspec-bug+sort%3Aupdated-desc) label.

Ah, but now I see that is referring to labels on issues, not PRs.

content/client-server-api/modules/receipts.md
```diff
@@ -137,16 +137,22 @@ either a thread root's event ID or `main` for the main timeline.
...
-  maximum of 3 hops is recommended to cover indirect relationships.
-
-Events not in a thread but still in the room are considered to be part of the
-"main timeline", or a special thread with an ID of `main`.
+considered to be "in a thread" if:
+
+* It has a rel_type of m.thread, or
+* It has a parent event with this rel_type, or a parent of a parent, or further
+  up the chain of relations. (Implementations should not recurse to arbitrary
+  depth: a maximum of 3 hops is recommended to cover indirect relationships.)
```
clokep, line 145: This seems to have lost all the formatting:

```suggestion
* It has a `rel_type` of `m.thread`, or
* It has a parent event with this `rel_type`, or a parent of a parent, or further
  up the chain of relations. (Implementations should not recurse to arbitrary
  depth: a maximum of 3 hops is recommended to cover indirect relationships.)
```

Personally I find the wording of the last bullet more confusing than the previous wording. 🤷

content/client-server-api/modules/receipts.md
```diff
@@ -137,16 +137,22 @@ either a thread root's event ID or `main` for the main timeline.
...
+
+Events not in a thread but still in the room are considered to be in the "main
+timeline". Threaded receipts for the main timeine use a special thread ID of
+`main`.
```
clokep, line 149: It isn't just about receipts, it is also about notification counts, etc. which is why it is noted to have a special ID.

content/client-server-api/modules/threading.md
```diff
@@ -12,11 +12,11 @@ as by providing some context to what is going on in the thread but keeping the f
...
-*thread root* (the first event in a thread). It is not possible to create a
-thread from an event which itself is the child of an event relationship (i.e.,
-one with an `m.relates_to` property). It is therefore also not possible to nest
-threads. All events in a thread reference the thread root instead of the
-most recent message, unlike rich reply chains.
+*thread root*. It is not possible to create a thread from an event which itself
+is the child of an event relationship (i.e., one with an `m.relates_to`
+property). It is therefore also not possible to nest threads. All events in a
+thread reference the thread root instead of the most recent message, unlike rich
+reply chains.
```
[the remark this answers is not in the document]
andybalaam: The thread root is not in the thread, so calling it the first event in the thread seemed confusing to me.

content/client-server-api/modules/receipts.md
andybalaam: The crucial part is to remove "It has child events with a `rel_type` of `m.thread` (in which case it'd be the thread root)", which was incorrect.

Travis told me not to use the word "ancestor", which was how I previously tried to describe it.

I hate how it reads now, so suggestions welcome :-)

content/client-server-api/modules/receipts.md
clokep: That would be fine. If a future use of it comes along then the phrasing can be updated. 👍

andybalaam:
```suggestion
Events not in a thread but still in the room are considered to be in the "main
timeline". When referring to the main timeline as a thread (e.g. in receipts
and notifications counts) a special thread ID of `main` is used.
```

content/client-server-api/modules/receipts.md
andybalaam: Done in c8e39c10

content/client-server-api/modules/receipts.md
```diff
@@ -137,16 +137,22 @@ either a thread root's event ID or `main` for the main timeline.
...
+
+Events not in a thread but still in the room are considered to be in the "main
+timeline". When referring to the main timeline as a thread (e.g. in receipts
+and notifications counts) a special thread ID of `main` is used.
+
+Thread roots are considered to be in the main timeline, as are events that are
+related to a thread root via non-thread relations. Note that clients providing a
+single-thread view will probably want to include thread roots and some of their
+child events (e.g. reactions) in that view, even though from a receipt point of
+view they are not part of that thread.
```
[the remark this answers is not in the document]
andybalaam: Removed in 3c439d09

content/client-server-api/modules/threading.md
```diff
@@ -12,11 +12,11 @@ as by providing some context to what is going on in the thread but keeping the f
 history behind a disclosure.
 
 Threads are established using a `rel_type` of `m.thread` and reference the
-*thread root* (the first event in a thread). It is not possible to create a
-thread from an event which itself is the child of an event relationship (i.e.,
-one with an `m.relates_to` property). It is therefore also not possible to nest
-threads. All events in a thread reference the thread root instead of the
-most recent message, unlike rich reply chains.
+*thread root*. It is not possible to create a thread from an event which itself
```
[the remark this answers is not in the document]
andybalaam: Attempted a definition in 058e97e5 . I removed the original definition because it was not accurate (because it said the thread root was in the thread, which it is not).

turt2live (approved):
Thanks!

PR #1685, opened 2023-12-01
Batch receipts by thread as well as room

### LINKED ISSUE #1684: Batching receipts should be per room/thread, not just per room
**Link to problem area**: https://github.com/matrix-org/matrix-spec/blob/main/content/client-server-api/modules/receipts.md?plain=1#L211

**Issue**
What is wrong?

In "Server behaviour" under "Receipts" we say:

> For efficiency, receipts SHOULD be batched into one event per room
> before delivering them to clients.

However, if a user is sending receipts for multiple threads in the same room simultaneously, this batching would prevent sending some of those receipts.

Ideally, the format would not be:

```
room_id: receipt_type: user_id: info_including_thread_id
```

but something like:

```
room_id: thread_id: receipt_type: user_id: info
```

But it's probably too late for that.

**Expected behaviour**

So instead I suggest rewording the above sentence to something like:

> For efficiency, receipts SHOULD be batched into one event per room **/thread**
> before delivering them to clients.

I haven't been able to verify server behaviour in this situation, but maybe a server dev could comment - do we actually lose receipts in this circumstance at the moment?

andybalaam:
Fixes #1684 

Avoid implying receipts for the same user+room but different threads should be batched together, since the JSON format does not allow it.

Preview: https://pr1685--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1685.clarification
+++ b/changelogs/client_server/newsfragments/1685.clarification
@@ -1,1 +1,1 @@
-Separate receipt batches by thread as well as room
+Clarify that read-receipts should be batched by thread as well as room.
--- a/changelogs/client_server/newsfragments/1685.clarification
+++ b/changelogs/client_server/newsfragments/1685.clarification
@@ -1,0 +1,1 @@
+Separate receipt batches by thread as well as room
--- a/content/client-server-api/modules/receipts.md
+++ b/content/client-server-api/modules/receipts.md
@@ -208,7 +208,7 @@ event when the user expands that thread.
 
 #### Server behaviour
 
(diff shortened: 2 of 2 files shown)


changelogs/client_server/newsfragments/1685.clarification
```diff
@@ -0,0 +1 @@
+Separate receipt batches by thread as well as room
```
richvdh, line 1:
```suggestion
Clarify that read-receipts should be batched by thread as well as room.
```

richvdh (approved):
tyvm

PR #1687, opened 2023-12-01
Clarify that threads can be created based on replies

### LINKED ISSUE #1683: Spec says you can't create a thread from a reply, but popular clients and servers say you can
**Link to problem area**: https://github.com/matrix-org/matrix-spec/blob/main/content/client-server-api/modules/threading.md?plain=1#L15

**Issue**
What is wrong?

The spec says you can't make a thread based on a reply:

> It is not possible to create a thread from an event which itself
> is the child of an event relationship (i.e., one with an `m.relates_to`
> property). It is therefore also not possible to nest threads.

But Element Web allows threads to be created with replies as thread root:

image

The reply and thread events shown above look like:

```
Event ID: $8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU
{
  "type": "m.room.message",
  "content": {
    "msgtype": "m.text",
    "body": "> <@x:y.co> Message\n\nReply to message",
...
    "m.relates_to": {
      "m.in_reply_to": {
        "event_id": "$E7IFEzhoNhXMr14TkrTbXtWCQoMEeVgdCyJJWG9a4Rs"
      }
    }
  }
}
```

```
Event ID: $q0UI_f7kEAW3E6X19wCFzplwksWTaYG15frgdjpgeko
Thread Root ID: $8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU
{
  "type": "m.room.message",
  "content": {
    "msgtype": "m.text",
    "body": "Thread from reply",
    "m.mentions": {},
    "m.relates_to": {
      "rel_type": "m.thread",
      "event_id": "$8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU",
      "is_falling_back": true,
      "m.in_reply_to": {
        "event_id": "$8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU"
      }
    }
  }
}
```

So the thread root is the reply event.

Here is the exchange with Synapse:

```
GET https://matrix-client.matrix.org/_matrix/client/v1/rooms/!OQYixnWsCXQExbbeWl%3Abeta.matrix.org/relations/%248UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU/m.thread?dir=b&limit=2

200

{
  "chunk": [
    {
      "content": {
        "ciphertext": "...",
        "m.relates_to": {
          "event_id": "$8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU",
          "is_falling_back": true,
          "m.in_reply_to": {
            "event_id": "$q0UI_f7kEAW3E6X19wCFzplwksWTaYG15frgdjpgeko"
          },
          "rel_type": "m.thread"
        },
        ...
      },
      "event_id": "$TWdaar0SQQTHbveDWg3ZyAIn0nApyMHUjnLAAcimSTQ",
      ...
    },
    {
      "content": {
        "ciphertext": "...",
        "m.relates_to": {
          "event_id": "$8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU",
          "is_falling_back": true,
          "m.in_reply_to": {
            "event_id": "$8UHA1RWgo_0r5V1gohzjPfupGCJI-SZQc8OqjATmelU"
          },
          "rel_type": "m.thread"
        },
        ...
      },
      ...
      "event_id": "$q0UI_f7kEAW3E6X19wCFzplwksWTaYG15frgdjpgeko",
    }
  ]
}
```

So Synapse is OK with it.

Element X displays it as a thread fallback, similar to other threads.

Nheko shows it as a thread (notice the little purple icon near the ticks):

image

**Expected behaviour**

I think the spec should probably allow replies to be thread roots - I don't see any harm in it.

So the wording might become:

> It is not possible to nest threads: you are not allowed to create a thread
> from an event which itself is in a thread (see ... for the exact definition of
> when an event is in a thread).

This would allow any event except one that is in a thread to become a thread root, including reactions, edits etc. We might want to restrict it to just replies, but that seems a little messy.

This sentence, later in the same section would also need rewording:

> As shown, any event without a `rel_type` can become a thread root by simply referencing it
> using an `m.thread` relationship.

Maybe to:

> As shown, any event **not in a thread** can become a thread root by simply referencing it
> using an `m.thread` relationship.

andybalaam:
Fixes #1683 

Preview: https://pr1687--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1687.clarification
+++ b/changelogs/client_server/newsfragments/1687.clarification
@@ -1,0 +1,1 @@
+Clarify that threads can be created based on replies
--- a/changelogs/client_server/newsfragments/1687.clarification
+++ b/changelogs/client_server/newsfragments/1687.clarification
@@ -1,1 +1,1 @@
-Clarify that threads can be created based on replies
+Clarify that threads can be created based on replies.
--- a/content/client-server-api/modules/threading.md
+++ b/content/client-server-api/modules/threading.md
@@ -14,7 +14,8 @@ history behind a disclosure.
 Threads are established using a `rel_type` of `m.thread` and reference the
 *thread root* (the main timeline event to which the thread events refer). It is not possible to create a thread from an event which itself
 is the child of an event relationship (i.e., one with an `m.relates_to`
-property). It is therefore also not possible to nest threads. All events in a
+property with a `rel_type` property - see [Relationship types](#relationship-types)).
+It is therefore also not possible to nest threads. All events in a
 thread reference the thread root instead of the most recent message, unlike rich
 reply chains.
 
--- a/content/client-server-api/modules/threading.md
+++ b/content/client-server-api/modules/threading.md
@@ -15,9 +15,10 @@ Threads are established using a `rel_type` of `m.thread` and reference the
 *thread root* (the main timeline event to which the thread events refer). It is not possible to create a thread from an event which itself
 is the child of an event relationship (i.e., one with an `m.relates_to`
 property with a `rel_type` property - see [Relationship types](#relationship-types)).
-It is therefore also not possible to nest threads. All events in a
-thread reference the thread root instead of the most recent message, unlike rich
-reply chains.
+It is therefore also not possible to nest threads. 
+
+Unlike rich reply chains, all events in a thread reference the thread root
+instead of the most recent message.
 
 As a worked example, the following represents a thread and how it would be formed:
 


changelogs/client_server/newsfragments/1687.clarification
```diff
@@ -0,0 +1 @@
+Clarify that threads can be created based on replies
```
richvdh, line 1:
```suggestion
Clarify that threads can be created based on replies.
```

content/client-server-api/modules/threading.md
```diff
@@ -14,7 +14,8 @@ history behind a disclosure.
 Threads are established using a `rel_type` of `m.thread` and reference the
 *thread root* (the main timeline event to which the thread events refer). It is not possible to create a thread from an event which itself
 is the child of an event relationship (i.e., one with an `m.relates_to`
-property). It is therefore also not possible to nest threads. All events in a
+property with a `rel_type` property - see [Relationship types](#relationship-types)).
+It is therefore also not possible to nest threads. All events in a
 thread reference the thread root instead of the most recent message, unlike rich
 reply chains.
```
richvdh, line 20: I think a new paragraph would help readability here:

```suggestion
It is therefore also not possible to nest threads. 

Unlike rich reply chains, all events in a thread reference the thread root
instead of the most recent message.
```

richvdh (approved):
tyvm!

Issue #1686, opened 2023-12-01
Spec incorrectly claims that state events are *overwritten*


OSto-git-ut:
**Link to problem area**:
[Link](https://spec.matrix.org/v1.9/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey)

**Issue**
Specification says: 
"State events can be sent using this endpoint. These events will be overwritten if room id, event type and state key all match."

Under [Link]( https://spec.matrix.org/v1.9/client-server-api/#types-of-room-events) its stated, that 
"State events: These are events which update the metadata state of the room (e.g. room topic, room membership etc). State is keyed by a tuple of event type and a state_key. **State in the room** with the same key-tuple will be **overwritten**."

**Expected behaviour**
I assume that for the /_matrix/client/v3/rooms/{roomId}/state/{eventType}/{stateKey}  endpoint, also the change of the state and not overwriting of the event is supposed to be done. Otherwise the DAG (graph) of state events is destroyed.

AdamBien:
I found this one:

> "State events can be sent using this endpoint. 
> These events will be overwritten if <room id>, <event type> and <state key> all match."

https://spec.matrix.org/v1.3/client-server-api/#put_matrixclientv3roomsroomidstateeventtypestatekey

AdamBien:
> That's not what it says; your copy is missing a bunch of critical words:

...sorry - the text in the brackets was removed by the markdown processor, but you are pointing to the right fragment. Thanks for offering the pull requests :-)

AllMightLegend:
Found this : 
"State events can be sent using this endpoint. These events will be overwritten if <room id>, <event type> and <state key> all match."
And
"If the event type being sent is m.room.canonical_alias servers SHOULD ensure that any new aliases being listed in the event are valid per their grammar/syntax and that they point to the room ID where the state event is to be sent. Servers do not validate aliases which are being removed or are already present in the state event."

I think it is meant that the "State" and "events" both are being referred to the state events that will be change if all of them match

PR #1689, opened 2023-12-04
Clarification: the state and not the events are overwritten

AdamBien:
The history of events remains / the event is appended and only the state is overwritten

Preview: https://pr1689--matrix-spec-previews.netlify.app

Fixes: #1686

### DIFF
--- a/data/api/client-server/room_state.yaml
+++ b/data/api/client-server/room_state.yaml
@@ -20,7 +20,7 @@ paths:
     put:
       summary: Send a state event to the given room.
       description: |
-        State events can be sent using this endpoint.  These events will be appended
+        State events can be sent using this endpoint.  These events will be processed
         and the state overwritten if `<room id>`, `<event type>` and `<state key>` all
         match.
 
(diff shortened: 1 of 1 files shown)


data/api/client-server/room_state.yaml
```diff
@@ -20,8 +20,8 @@ paths:
     put:
       summary: Send a state event to the given room.
       description: |
-        State events can be sent using this endpoint.  These events will be
-        overwritten if `<room id>`, `<event type>` and `<state key>` all
+        State events can be sent using this endpoint.  These events will be appended
```
richvdh, line 23: appended to what?

richvdh (changes requested):
Thank you for your contribution! Please take a look at the [contributing guide](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst), particularly with respect to [changelog entries](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog) and [sign-off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off).

data/api/client-server/room_state.yaml
AdamBien: +1 we should not prescribe how the events are processed internally. I changed the wording from appended to "processed"

richvdh:
Unfortunately we can't accept this without sign-off.

PR #1690, opened 2023-12-04
Make clearer in the example that the fallback prefix sequence should be repeated for each line

Konubinix:
I needed to read the spec twice to make sure it was the case.

Also, I encountered an implementation of matrix that added the sequence only for the first line.

I believe that making the first example of the chapter show several lines will make less likely such mistakes in the future.

Preview: https://pr1690--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1690.clarification
+++ b/changelogs/client_server/newsfragments/1690.clarification
@@ -1,1 +1,1 @@
-Make clearer in the example that the fallback prefix sequence should be repeated for each line.
+Make clearer in the example for reply fallbacks that the prefix sequence should be repeated for each line.
--- a/changelogs/client_server/newsfragments/1690.clarification
+++ b/changelogs/client_server/newsfragments/1690.clarification
@@ -1,0 +1,1 @@
+Make clearer in the example that the fallback prefix sequence should be repeated for each line.
--- a/content/client-server-api/modules/rich_replies.md
+++ b/content/client-server-api/modules/rich_replies.md
@@ -60,7 +60,8 @@ specific fallback text is different for each `msgtype`, however the
 general format for the `body` is:
 
 ```text
-> <@alice:example.org> This is the original body
+> <@alice:example.org> This is the first line of the original body
+> This is the second line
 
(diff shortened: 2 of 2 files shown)


richvdh:
I'm sorry you found it a chore

richvdh (changes requested):
Thank you for your contribution. Please take a look at the [contributing guide](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst), particularly with respect to [changelog entries](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog) and [sign-off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off).

Konubinix:
Thank you for the answer,

I will read the document and adjust the PR accordingly.

-- 
Konubinix
GPG Key    : 7439106A
Fingerprint: 5993 BE7A DA65 E2D9 06CE  5C36 75D2 3CED 7439 106A

Konubinix:
IIUC, the only thing I needed to do in that particular case was to sign off my
commit, right?

So, I did it :-)

Please let me know if there is something I missed.

Konubinix:
I understand that updating the changelog is done only when the API changes.

richvdh:
Yeah the wording is unclear; apologies. (https://github.com/matrix-org/matrix-spec/pull/1697 should improve it.)

Please could you create `changelogs/client_server/newsfragments/1690.clarification` with a brief summary (`Clarify example for reply fallbacks.` or somesuch).

changelogs/client_server/newsfragments/1690.clarification
```diff
@@ -0,0 +1 @@
+Make clearer in the example that the fallback prefix sequence should be repeated for each line.
```
richvdh, line 1:
```suggestion
Make clearer in the example for reply fallbacks that the prefix sequence should be repeated for each line.
```

richvdh (approved):
Thank you for the contribution! Sorry the review took a while.

PR #1695, opened 2023-12-11
Clarify secret storage format

richvdh:
I found the description of how the `m.secret_storage.key.[key ID]` account data object is calculated extremely confusing.

### DIFF
--- a/content/client-server-api/modules/secrets.md
+++ b/content/client-server-api/modules/secrets.md
@@ -66,6 +66,62 @@ default key.
 |------------|-----------|------------------------------------------|
 | key        | string    | **Required.** The ID of the default key. |
 
+
+###### `m.secret_storage.v1.aes-hmac-sha2`
+
+For the purposes of allowing clients to check whether a user has correctly
+entered the key, keys for use with the `m.secret_storage.v1.aes-hmac-sha2`
+algorithm are stored with some additional data.
+
+When storing a key, clients should:
+
+1.  Given the secret storage key, generate 64 bytes by performing an
+    HKDF with SHA-256 as the hash, a salt of 32 bytes of 0, and the empty
+    string as the info. The first 32 bytes are used as the AES key,
+    and the next 32 bytes are used as the MAC key.
+
+2.  Generate 16 random bytes, set bit 63 to 0 (in order to work around
+    differences in AES-CTR implementations), and use this as the AES
+    initialization vector (IV).
+
+3.  Encrypt a message consisting of 32 byutes of 0, using AES-CTR-256 using the
+    AES key and IV generated above.
+
+4.  Pass the raw encrypted data through HMAC-SHA-256 using the MAC key
+    generated above.
+
+5.  Encode the IV from step 2, and the MAC from step 4, using base64, and store
+    the results in the `iv` and `mac` properties respectively in the
+    `m.secret_storage.key.[key ID]` account-data. (The ciphertext from step 3
+    is discarded after passing through the MAC calculation.)
+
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/secrets.md
```diff
@@ -147,58 +203,35 @@ HMAC-SHA-256. The secret is encrypted as follows:
...
-| iv         | string  |   **Required.** The 16-byte initialization vector, encoded as base64.  |
-| ciphertext | string  |   **Required.** The AES-CTR-encrypted data, encoded as base64.         |
-| mac        | string  |   **Required.** The MAC, encoded as base64.                            |
+4.  Pass the raw encrypted data through HMAC-SHA-256 using the MAC key
+    generated above.
 
-For the purposes of allowing clients to check whether a user has
-correctly entered the key, clients should:
+5.  Encode the IV from step 2, the ciphertext from step 3, and MAC from step 4,
+    using base64, and store them as the `iv`, `ciphertext`, and `mac`
```
richvdh, line 219: Question? Padded or unpadded base64? If padded, which variant?

Existing text just says "base64", which implies padded (of some variant). Web uses (I think) padded [RFC4648](https://datatracker.ietf.org/doc/html/rfc4648) base64. But the spec also [says](https://spec.matrix.org/v1.8/client-server-api/#secret-storage)

> it’s expected that most encryption schemes would have `ciphertext` and `ma`c properties, where the `ciphertext` property is the **unpadded** base64-encoded ciphertext

content/client-server-api/modules/secrets.md
```diff
@@ -66,6 +66,62 @@ default key.
...
+
+`AesHmacSha2KeyDescription`
+
+| Parameter   | Type   | Description                                                                                          |
+|-------------|--------|------------------------------------------------------------------------------------------------------|
+| name        | string | Optional. The name of the key.                                                                       |
+| algorithm   | string | **Required.** The encryption algorithm to be used for this key: `m.secret_storage.v1.aes-hmac-sha2`. |
+| passphrase  | object | See [deriving keys from passphrases](#deriving-keys-from-passphrases) section for a description of this property. |
+| iv          | string | The 16-byte initialization vector, encoded with base64.                                              |
+| mac         | string | The MAC of the result of encrypting 32 bytes of 0, encoded with base64.                              |
```
richvdh, line 112: Question: are these required or not? If not, what are clients expected to do if they are absent? Reject the key, or carry on regardless?

cf https://github.com/vector-im/element-web/issues/26721

uhoreg: The [MSC said](https://github.com/matrix-org/matrix-spec-proposals/pull/2472/files#diff-c1213725eb7eea9b96e7fa6af61456682fb88a93640423c3c43727abd7faa96aR46-R52) that clients "should" calculate and store the IV and MAC, not "must", so I guess they're optional.  I don't remember why we made it optional.

If they're not present, we have to assume the key is correct, and carry on.

content/client-server-api/modules/secrets.md
uhoreg: Yeah, I made a mistake not specifying this originally.  We want it to be unpadded, to be consistent with everything else in Matrix, but since it was left unspecified, clients should be able to handle either one, for example by base64-decoding the MAC and comparing the binary value instead of comparing the base64 strings.

uhoreg (approved):

richvdh:
@uhoreg: thanks for the clarifications. Have attempted to clarify the spec text.

PR #1699, opened 2023-12-26
Update docsy to v0.8.0

zecakeh:
I tried to apply the minimum set of changes I could to have no changes in the generated spec. Each commit can be reviewed separately and the description usually explains why the change is needed.

After the style fixes I applied, the only visible change that I could spot is that now the code blocks have a button at the top right to copy the content, and I am not sure whether we should remove it or not:

Capture d’écran du 2023-12-26 11-34-13

~~This as a draft for now because it is based on my personal fork so the website can be built while waiting for https://github.com/matrix-org/docsy/pull/2 to be merged.~~

Preview: https://pr1699--matrix-spec-previews.netlify.app


richvdh (approved):
LGTM. Thank you!

PR #1709, opened 2024-01-15
Document case of knock->join membership transition

turt2live:
This adds a line from the `knock` block to the `join` block, as the transition is legal in `knock_restricted` rooms when using restricted joins.

See [auth rule 4.3.5](https://spec.matrix.org/v1.9/rooms/v11/#authorization-rules).

Partial fix to https://github.com/matrix-org/matrix-spec/issues/2053

Preview: https://pr1709--matrix-spec-previews.netlify.app


Xiretza:
The table in https://github.com/matrix-org/matrix-spec/blob/main/data/event-schemas/schema/m.room.member.yaml will need updating as well.

Tangentially related: what happens to knocking users when a room changes its `m.rooom.join_rules` from `knock` to `public`? Do they have to `/leave` and then `/join` again or can they go straight to `/join` (which would be the same transition as added by this PR)?

richvdh:
> Tangentially related: what happens to knocking users when a room changes its m.rooom.join_rules from knock to public? Do they have to /leave and then /join again or can they go straight to /join (which would be the same transition as added by this PR)?

This is a great point. According to the auth rules, you can go straight to `join`. And that needs adding to the diagram.

static/diagrams/membership.drawio
```diff
@@ -1 +1,159 @@
...
+          <mxGeometry x="-0.6" relative="1" as="geometry">
+            <mxPoint as="offset" />
+          </mxGeometry>
+        </mxCell>
+        <mxCell id="Je66lSOIGBRFVx_LtrYI-3" value="/knock if&lt;br&gt;knock-able&lt;br&gt;join_rules&lt;br&gt;" style="edgeStyle=orthogonalEdgeStyle;rounded=0;orthogonalLoop=1;jettySize=auto;html=1;exitX=0.75;exitY=0;exitDx=0;exitDy=0;entryX=0.75;entryY=1;entryDx=0;entryDy=0;" parent="1" source="P05zALBlSnZwpnZ5Cu9t-4" target="P05zALBlSnZwpnZ5Cu9t-24" edge="1">
+          <mxGeometry x="-0.4" y="-5" relative="1" as="geometry">
+            <mxPoint as="offset" />
+          </mxGeometry>
+        </mxCell>
+        <mxCell id="q5SATBPCpZFuv_SF1K4v-1" value="/join if join_rules&lt;br&gt;knock_restricted&lt;br&gt;and using a &lt;br&gt;restricted join" style="edgeStyle=orthogonalEdgeStyle;rounded=0;orthogonalLoop=1;jettySize=auto;html=1;exitX=0.25;exitY=0;exitDx=0;exitDy=0;entryX=0;entryY=0.25;entryDx=0;entryDy=0;jumpStyle=arc;" edge="1" parent="1" source="P05zALBlSnZwpnZ5Cu9t-24" target="P05zALBlSnZwpnZ5Cu9t-2">
```
richvdh, line 146: > /join if join_rules knock_restricted and using a restricted join

I assume by "restricted join", we mean "a join satisfying one of the conditions for [restricted rooms](https://spec.matrix.org/v1.9/client-server-api/#conditions)"?

The transition is also permitted if the join_rules are `restricted` (and the restriction conditions are satisfied), or as @Xiretza points out, if the join rules are `public`.

Indeed, I think the rules for the knock -> join transition are identical to those for the leave -> join and external -> join transitions (which makes sense intuitively, and makes sense because neither `leave` nor `knock` are specifically mentioned in auth rule 4.3, which covers the transition to `join` generally). So, can we just make it use the same label "/join if allowed by join_rules" ?

richvdh (changes requested):

Kladki:
The [`m.room.member`](https://spec.matrix.org/v1.10/client-server-api/#mroommember) table also needs to be updated, since it currently states that "From `knock` to `join` Must never happen."

richvdh:
@turt2live are you still working on this?

turt2live:
I forgot this existed, sorry. I'd suggest someone else takes it over (if needed).

richvdh:
I'm going to close this, because it's now bitrotted. Have opened https://github.com/matrix-org/matrix-spec/issues/2053 to track the problem.

PR #1712, opened 2024-01-16
Clarify that the key backup MAC is implemented incorrectly

sumnerevans:
---
name: Clarify that the key backup MAC is implemented incorrectly
about: Clarify that the key backup MAC is implemented incorrectly
title: Clarify that the key backup MAC is implemented incorrectly
labels: ''
assignees: ''

---

Due to a bug in libolm, all implementations of the
m.megolm_backup.v1.curve25519-aes-sha2 key backup algorithm incorrectly
pass an empty string through HMAC-SHA-256 to generate the `mac` property
of the `session_data`.

It was intended for the entire raw encrypted data to be passed through
HMAC-SHA-256, but the issue was caught too late in the process, and thus
we are stuck with this until a new key backup algorithm is introduced.

This commit clarifies the real-world behavior of all current
implementations.

Signed-off-by: Sumner Evans <sumner@beeper.com>

### Pull Request Checklist

Preview: https://pr1712--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1712.clarification
+++ b/changelogs/client_server/newsfragments/1712.clarification
@@ -1,0 +1,1 @@
+Clarify that the key backup MAC is implemented incorrectly and does not pass the ciphertext through HMAC-SHA-256.
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1368,14 +1368,14 @@ The `session_data` field in the backups is constructed as follows:
     The first 8 bytes of the resulting MAC are base64-encoded, and become the
     `mac` property of the `session_data`.
 
-    {{% boxes/warning %}}
-    This was intended to pass the raw encrypted data, but due to a bug in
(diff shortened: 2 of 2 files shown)


tulir:
This was found while implementing key backup in mautrix-go (https://github.com/mautrix/go/pull/154)

For reference, this is implemented in matrix-rust-sdk for libolm compatibility: [decrypt](https://github.com/matrix-org/matrix-rust-sdk/blob/0.7.0/crates/matrix-sdk-crypto/src/backups/keys/compat.rs#L112-L113), [encrypt](https://github.com/matrix-org/matrix-rust-sdk/blob/0.7.0/crates/matrix-sdk-crypto/src/backups/keys/compat.rs#L161-L163) (it never calls update on the hmac). I think the libolm implementation is somewhere around https://gitlab.matrix.org/matrix-org/olm/-/blob/3.2.16/src/pk.cpp#L408-427, but I don't understand C++ well enough to tell why it's wrong.

uhoreg (approved):
Thanks

Issue #1681, opened 2023-11-29
NPM publishing for emoji translations still broken


turt2live:
https://github.com/matrix-org/matrix-spec/actions/runs/7036057140/job/19147844038

Previous fix: https://github.com/matrix-org/matrix-spec/pull/1648

turt2live:
Looks like it's still failing: https://github.com/matrix-org/matrix-spec/actions/runs/8393215998

New error, though.

turt2live:
Another quarter, another error message: https://github.com/matrix-org/matrix-spec/actions/runs/9601021575/job/26478620278

t3chguy:
![image](https://github.com/matrix-org/matrix-spec/assets/2403652/8b353e49-70e0-4249-8b8d-d831c25399de)

Looks like a missing secret, I lack access to investigate

turt2live:
yea, we forgot to add the `NPM_TOKEN` secret, oops. Adding it and re-running the job did the thing: https://www.npmjs.com/package/@matrix-org/spec/v/1.11.0

PR #1713, opened 2024-01-17
Fix npm release script

t3chguy:
Fixes https://github.com/matrix-org/matrix-spec/issues/1681

Preview: https://pr1713--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1713.misc
+++ b/changelogs/internal/newsfragments/1713.misc
@@ -1,0 +1,1 @@
+Fix npm release script for `@matrix-org/spec`.
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -27,7 +27,7 @@ jobs:
         run: "yarn install --frozen-lockfile"
 
       - name: 🎖 Bump package.json version
-        run: "yarn version $VERSION"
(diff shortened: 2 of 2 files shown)


richvdh:
I have a vague memory of us discussing this before, but I've forgotten everything. For me, and for future archaeologists, could you add some context here: has this always been broken, or did it break recently?

richvdh:
Is this the same as https://github.com/matrix-org/matrix-spec/issues/1681? Or different?

t3chguy:
@richvdh it likely has always been broken, all packages thus far have been manually deployed by me. Good spot on https://github.com/matrix-org/matrix-spec/issues/1681 - updated OP

t3chguy:
Manually published 1.9.0 https://www.npmjs.com/package/@matrix-org/spec

changelogs/internal/newsfragments/1713.misc
```diff
@@ -0,0 +1 @@
+Fix npm release script for `@matrix-org/spec`.
```
turt2live, line 1: somehow both CI and I missed this in review: `.misc` doesn't exist in the spec - we use `.clarification`

PR #1715, opened 2024-01-18
clarify otk and fallback key types in examples

uhoreg:
- remove unsigned curve25519 keys from examples because we don't use those for otks and fallback keys
- add missing `device_unused_fallback_key_types` property, which is required

Preview: https://pr1715--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Thank you!

Issue #1710, opened 2024-01-16
Re-clarify the auth rules around invite->knock transitions


kegsay:
This is a mess so strap yourself in.

# Background

Knocking was introduced in MSC2403 and got merged in March 2021. It introduced a new membership state "knock" along with a set of membership transitions to and from it. A puzzling question revolved around the transition from "invite" to "knock". Generally, this makes no sense. A user knocks on a room _to receive an invite_. If the inviting user wishes to disinvite the user, there is an existing transition from "invite" to "leave" to reflect this. Adding extra transitions increases complexity of the already complex membership state transition diagram. Extra complexity increases the risk of implementation errors or protocol errors.

# Timeline

MSC2043, at the point it is accepted in March 2021, does *not* permit a  transition from invite to knock in the auth rules: https://github.com/matrix-org/matrix-spec-proposals/blob/8f304ca9e567ca5b5630992025453b9a21a81b60/proposals/2403-knock.md#auth-rules.

Likewise, the April 2021 [spec PR](https://github.com/matrix-org/matrix-spec-proposals/pull/3154) which adopted MSC2043 into the spec, forbade `invite->knock` transitions in both the auth rules and the state transition table.

The [Synapse implementation](https://github.com/element-hq/synapse/blame/69637f8bac0959386108edeea11ba456b9334486/synapse/event_auth.py#L703-L704) also did not allow this transition.

Now, this is what happened next..
 - Feb 2022: the membership state transition diagram in the spec is updated to _allow_ `invite->knock`: https://github.com/matrix-org/matrix-spec-proposals/pull/3730. The auth rules are *not* changed at this point.
 - June 2022: an issue is filed https://github.com/matrix-org/matrix-spec/issues/1142 stating that this transition is invalid by reading the auth rules. It is claimed that this is an error in the auth rules and (wrongly) that the Synapse implementation _allows_ this transition. Due to the implementation apparently allowing it, consensus forms around letting the transition be allowed, despite it being "a fairly silly thing to do".
 - July 2022: the spec is further updated to retrospectively change the auth rules in room versions 7, 8 and 10 to allow the transition: https://github.com/matrix-org/matrix-spec/pull/1175.
   At the same time, MSC2043 is [retrospectively updated](https://github.com/matrix-org/matrix-spec-proposals/pull/3850) to claim that invite->knock transitions are legal (despite the auth rules specified elsewhere in the MSC still not allowing it).

## Current situation

[...]

PR #1717, opened 2024-01-23
Clarify that invite->knock is *not* a valid transition

kegsay:
This reverts https://github.com/matrix-org/matrix-spec/pull/1175

See https://github.com/matrix-org/matrix-spec/issues/1710

Signed-off-by: Kegan Dougal <kegan@matrix.org>

Preview: https://pr1717--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1717.clarification
+++ b/changelogs/room_versions/newsfragments/1717.clarification
@@ -1,0 +1,1 @@
+For room versions 7 through 11: Clarify that `invite->knock` is not a legal transition.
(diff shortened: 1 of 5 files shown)


richvdh (approved):
As with https://github.com/matrix-org/matrix-spec-proposals/pull/4091: I think this is the right thing to do, but I would like opinions from others.

richvdh:
Also worth linking here: a complement test that asserts that what is implemented matches what this PR says: https://github.com/matrix-org/complement/pull/702.

dbkr (approved):
Thanks, I agree this is the correct way out of the mess, and doubly so with the complement test.

PR #1718, opened 2024-01-24
Add some clarifications around implementation requirements for MSCs

uhoreg:
- clarify that implementations don't need to be merged to be considered
- implementations may be required on both client-side and server-side
- explain labels, and how to tell us about implementations

Preview: https://pr1718--matrix-spec-previews.netlify.app

### DIFF
--- a/content/proposals.md
+++ b/content/proposals.md
@@ -381,7 +381,7 @@ of the MSC working in order for it to move into FCP. This can usually be
 a branch/pull request to whichever implementation of choice that proves
 the MSC works in practice, though in some cases the MSC itself will be
 small enough to be considered proven. Implementations do not need to be
-merged no released, but must be of sufficient quality to show that the
+merged or released, but must be of sufficient quality to show that the
 MSC works. Where it's unclear if an MSC will require an implementation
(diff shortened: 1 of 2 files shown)


uhoreg:
At some point, we may want to say some more things about what we want/expect from an implementation, but that's probably the topic of a future PR.

content/proposals.md
```diff
@@ -380,9 +380,18 @@ As part of the proposal process the Spec Core Team will require evidence
...
+merged no released, but must be of sufficient quality to show that the
+MSC works. Where it's unclear if an MSC will require an implementation
+proof, ask in [\#matrix-spec:matrix.org](https://matrix.to/#/#matrix-spec:matrix.org).
+Proposals may require both server-side and client-side implementations.
+
+Proposals that have not yet been implemented will have the
+`needs-implementation` label. After an implementation has been made, add a
+comment in the GitHub issue indicating so. After an implementation has been
+made, we will check it to verify that it implements the MSC. Proposals that
+have implementations that we have not yet been checked will have the
```
richvdh, line 393:
```suggestion
have implementations that have not yet been checked will have the
```

richvdh (approved):
LGTM otherwise

PR #1721, opened 2024-02-01
Use a POST request in Federation request signing example

reivilibre:
Otherwise the GET request having a body is confusing, since it is not valid HTTP

Preview: https://pr1721--matrix-spec-previews.netlify.app


uhoreg (approved):
Thanks.  Good catch

PR #1724, opened 2024-02-14
Update HTML templates to link to object definitions

richvdh:
This PR is a series of commits which add hyperlinks to the definitions of subschemas. I'm hoping this might help with navigation for complex data structures like that returned by `/sync`.

Each of the commits, with the exception of "Link to HTML anchors, where they exist" has been tested to have no effect n the output. The "Link..." commit just adds `<href a="...">` links where possible.

[Before](https://spec.matrix.org/v1.9/client-server-api/#get_matrixclientv3sync):

image

[After](https://pr1724--matrix-spec-previews.netlify.app/client-server-api/#get_matrixclientv3sync):

image

Preview: https://pr1724--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/json-schema/resolve-additional-types.html
+++ b/layouts/partials/json-schema/resolve-additional-types.html
@@ -70,9 +70,6 @@
             {{ $this_object = merge $this_object (dict "anchor" (printf "%s_%s" $anchor_base (anchorize $this_object.title))) }}
         {{ end }}
 
-        /* Add the object we were passed into the $all_objects array */
-        {{ $all_objects = $all_objects | append $this_object }}
-
         /* Add any nested objects referenced in this object's `additionalProperties` */
         {{ if $this_object.additionalProperties }}
             {{ if reflect.IsMap $this_object.additionalProperties }}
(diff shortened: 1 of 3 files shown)


layouts/partials/json-schema/resolve-additional-types.html
```diff
@@ -142,20 +129,17 @@
     {{ $this_object := partial "json-schema/resolve-allof" .this_object }}
 
     {{ $more_objects := partial "json-schema/resolve-additional-types" (dict "schema" $this_object "anchor_base" .anchor_base "name" $name) }}
-    {{/*
-      As far as I know we don't have something like Array.concat(), so add them one at a time
-    */}}
```
richvdh, line 147: I removed this comment, since I found it misleading. We *do* have something like `Array.concat` -- it's called `append`, which does concatenate arrays if you pass an array as its parameter. However, it's not very useful here, because we want to call `clean-object` on each item before adding it to `$all_objects`.

uhoreg (approved):
Seems to do what it says it does

PR #1728, opened 2024-02-22
Allow `/versions` to optionally accept authentication

zecakeh:
According to [MSC4026](https://github.com/matrix-org/matrix-spec-proposals/pull/4026).

I followed the [OpenAPI 3.1 spec](https://spec.openapis.org/oas/v3.1.0#operation-object) to declare that the authentication is optional, so I also had to adapt the partials to not show `Requires authentication: Yes`.

Preview: https://pr1728--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/openapi/render-operation.html
+++ b/layouts/partials/openapi/render-operation.html
@@ -60,21 +60,22 @@
  <tr>
   <th>Requires authentication:</th>
   {{/*
-    Authentication is optional if one of these is true:
-      - the security key is not set
-      - the security value contains an empty object
+    Authentication is defined with the `security` key. We assume that the
+    key is not set if no authentication is necessary. If the key is set,
+    authentication is required unless it contains an item that is an empty
+    object.
   */}}
-  {{ $requires_authentication := 1 }}
+  {{ $requires_authentication := "Yes" }}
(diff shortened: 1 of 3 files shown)


layouts/partials/openapi/render-operation.html
```diff
@@ -59,7 +59,22 @@ <h1 id="{{ lower $method }}{{ $anchor }}">
   </tr>
  <tr>
   <th>Requires authentication:</th>
-  <td>{{ if $operation_data.security }}Yes{{ else }}No{{ end }}</td>
+  {{/*
```
uhoreg, line 62: Since you're rewriting the logic here, how much of a pain would it be to make it say "Requires authentication: optional" if it's optional?

zecakeh: It wouldn't be complicated, but isn't this wording weird?

turt2live: "No" feels like there's *no* option for authentication, even if technically correct. "Optional" or "conditional" would be more suitable.

zecakeh: Done

uhoreg (approved):
Thanks!

PR #1733, opened 2024-02-27
Add server support discovery endpoint

### LINKED ISSUE #58: We should mandate homeservers declare a postmaster@ equivalent so we can talk to their admins if needed (SPEC-87)
(Imported from https://matrix.org/jira/browse/SPEC-87)

(Reported by @ara4n)

zecakeh:
As per [MSC1929](https://github.com/matrix-org/matrix-spec-proposals/pull/1929).

Fixes #58.

Preview: https://pr1733--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -363,6 +363,8 @@ specify parameter values. The flow for this method is as follows:
 
 {{% http-api spec="client-server" api="versions" %}}
 
+{{% http-api spec="client-server" api="support" %}}
+
 ## Client Authentication
 
 Most API endpoints require the user to identify themselves by presenting
--- a/data/api/client-server/support.yaml
+++ b/data/api/client-server/support.yaml
@@ -1,0 +1,136 @@
+# Copyright 2018 New Vector Ltd
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+openapi: 3.1.0
+info:
+  title: Matrix Client-Server Support Discovery API
+  version: 1.0.0
+paths:
+  /matrix/support:
+    get:
+      summary: Gets Matrix server admin contact and support page of the domain.
+      description: |-
+        Gets server admin contact and support page of the domain.
+
+        Like the [well-known discovery URI](/client-server-api/#well-known-uri),
+        this should be accessed with the hostname of the homeserver, by making a
+        GET request to `https://hostname/.well-known/matrix/support`.
+
+        Note that this endpoint is not necessarily handled by the homeserver,
+        but by another webserver, to be used for discovering support information
+        for the homeserver.
+      operationId: getWellknownSupport
+      x-addedInMatrixVersion: "1.10"
+      responses:
+        "200":
+          description: Server support information.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  contacts:
+                    type: array
+                    description: |-
+                      Ways to contact the server administrator.
+
+                      At least one of `contacts` or `support_page` is required.
+                      If only this is set, it must contain at least one item.
+                    items:
+                      type: object
+                      title: Contact
+                      description: A way to contact the server administrator.
+                      properties:
+                        matrix_id:
+                          type: string
+                          description: |-
+                            A Matrix User ID allowing to contact the
+                            administrator.
+
+                            It could be an account registered on a different
+                            homeserver so the administrator can be contacted
+                            when the homeserver is down.
+
+                            At least one of `matrix_id` or `email_address` is
+                            required.
+                        email_address:
+                          type: string
+                          description: |-
+                            An email address allowing to contact the
+                            administrator.
+
+                            At least one of `matrix_id` or `email_address` is
+                            required.
+                        role:
+                          type: string
+                          enum:
+                            - "m.role.admin"
+                            - "m.role.security"
+                          description: |-
+                            An informal description of what the contact methods
+                            are used for.
+
+                            `m.role.admin` is a catch-all user for any queries
+                            and `m.role.security` is intended for sensitive
+                            requests.
+
+                            Unspecified roles are permitted through the use of
+                            [Namespaced Identifiers](/appendices/#common-namespaced-identifier-grammar).
+                      required:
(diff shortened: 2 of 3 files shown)


data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
+# Copyright 2018 New Vector Ltd
```
turt2live, line 1: *ahem* ;)

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+openapi: 3.1.0
+info:
+  title: Matrix Client-Server Support Discovery API
+  version: 1.0.0
+paths:
+  /matrix/support:
+    get:
+      summary: Gets Matrix server admin contact and support page of the domain.
```
turt2live, line 21:
```suggestion
      summary: Gets homeserver contacts and support details.
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+      description: |-
+        Gets server admin contact and support page of the domain.
+
+        Like the [well-known discovery URI](/client-server-api/#well-known-uri),
+        this should be accessed with the hostname of the homeserver, by making a
```
turt2live, line 26:
```suggestion
        this should be accessed with the hostname of the homeserver by making a
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+        GET request to `https://hostname/.well-known/matrix/support`.
+
+        Note that this endpoint is not necessarily handled by the homeserver,
+        but by another webserver, to be used for discovering support information
+        for the homeserver.
```
turt2live, line 31:
```suggestion
        Note that this endpoint is not necessarily handled by the homeserver.
        It may be served by another webserver, used for discovering support information
        for the homeserver.
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+          description: Server support information.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  contacts:
+                    type: array
+                    description: |-
+                      Ways to contact the server administrator.
```
turt2live, line 45: We should define "administrator" somewhere here. I believe the MSC had some notion of what the admin's contact could be used for - can we incorporate that into the schema somewhere?

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+                    items:
+                      type: object
+                      title: Contact
+                      description: A way to contact the server administrator.
+                      properties:
+                        matrix_id:
+                          type: string
+                          description: |-
+                            A Matrix User ID allowing to contact the
+                            administrator.
```
turt2live, line 58:
```suggestion
                            A Matrix User ID representing the administrator.
```

turt2live: a link to the appendices would also be good here.

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+                            homeserver so the administrator can be contacted
+                            when the homeserver is down.
+
+                            At least one of `matrix_id` or `email_address` is
+                            required.
+                        email_address:
+                          type: string
+                          description: |-
+                            An email address allowing to contact the
+                            administrator.
```
turt2live, line 70:
```suggestion
                            An email address to reach the administrator.
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+                        role:
+                          type: string
+                          enum:
+                            - "m.role.admin"
+                            - "m.role.security"
+                          description: |-
+                            An informal description of what the contact methods
+                            are used for.
+
+                            `m.role.admin` is a catch-all user for any queries
```
turt2live, line 83:
```suggestion
                            `m.role.admin` is a catch-all role for any queries
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+              schema:
+                type: object
+                properties:
+                  contacts:
+                    type: array
+                    description: |-
+                      Ways to contact the server administrator.
+
+                      At least one of `contacts` or `support_page` is required.
+                      If only this is set, it must contain at least one item.
```
turt2live, line 48:
```suggestion
                      If only `contacts` is set, it must contain at least one item.
```

data/api/client-server/support.yaml
```diff
@@ -0,0 +1,136 @@
...
+        "404":
+          description: No server support information available.
+      tags:
+        - Server administration
+servers:
+  - url: "{protocol}://{hostname}{basePath}"
+    variables:
+      protocol:
+        enum:
+          - http
```
turt2live, line 130: we don't technically allow http here or on the other .well-known endpoints. Is this a copy/paste artifact?

turt2live (commented):
generally this looks good to me - I have a bunch of minor editorial comments.

Thank you!

data/api/client-server/support.yaml
zecakeh: Yes, this file started as a duplicate of wellknown.yaml

data/api/client-server/support.yaml
zecakeh: I don't see anything on the MSC about why the admins would be contacted, other than the description of the `role` property.

data/api/client-server/support.yaml
zecakeh: Removed

data/api/client-server/support.yaml
zecakeh: Fixed

content/client-server-api/_index.md
```diff
@@ -363,6 +363,8 @@ specify parameter values. The flow for this method is as follows:
 
 {{% http-api spec="client-server" api="versions" %}}
 
+{{% http-api spec="client-server" api="support" %}}
```
richvdh, line 366: I'm not convinced this belongs under "Server discovery". I'd be inclined to put it in a new "Module", but I'm open to other suggestions

turt2live: Rationale behind the suggestion to put it here is you're *discovering* information about the server without necessarily needing to speak the rest of Matrix. Both authenticated and unauthenticated clients can use this, and it's "before" the homeserver APIs really start to do anything.

Arguably though, we should rename or introduce a section for `.well-known` stuff. I think this is best left to a future PR at this stage.

turt2live (approved):
Thanks! Apologies for this taking so long to review.

If we do want to move it to another section/module, that can be done with a follow-up PR imo.

richvdh:
It's a bit of an anomaly that https://spec.matrix.org/unstable/client-server-api/#well-known-uri now talks about "Servers hosting the `.well-known` JSON file", and now there are *two* `.well-known` JSON files. Generally I don't feel that section is particularly coherent now. But yes, we can improve this later.

PR #1735, opened 2024-02-28
Add support for multi-stream VoIP

zecakeh:
As per [MSC3077](https://github.com/matrix-org/matrix-spec-proposals/pull/3077).

This is built on top of #1602 and incorporates the feedback from the first review, as was suggested in the [Matrix Spec and Docs Authoring Room](https://matrix.to/#/!KzgjYxOASBbBMrtCXC:matrix.org/$b0xGoyZ1Jf7Jx53jaoNVmoEX5yylSBtMTsgT08NNCR4?via=matrix.org&via=element.io&via=envs.net).

Threads which are not resolved in the other PR are reopened here.

Preview: https://pr1735--matrix-spec-previews.netlify.app


content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
 
 ##### Streams
```
zecakeh, line 172: @richvdh [said](https://github.com/matrix-org/matrix-spec/pull/1602/files#r1287259846):

> can we add something to https://pr1602--matrix-spec-previews.netlify.app/client-server-api/#common-fields about the new field?

content/client-server-api/modules/voip_events.md
```diff
@@ -171,18 +171,34 @@ In response to an incoming invite, a client may do one of several things:
...
+
+*  `m.usermedia` - stream that contains the webcam and/or microphone tracks
+*  `m.screenshare` - stream with the screen-sharing tracks
+
+If an incoming stream is not described in `sdp_stream_metadata` and
+`sdp_stream_metadata` is present, the stream should be ignored. If a stream has
+a `purpose` of an unknown type (i.e. not `m.usermedia` or `m.screenshare`), it
+should be ignored.
+
+Clients implementing this specification will ignore any streamless tracks. Note
```
zecakeh, line 192: @richvdh [said](https://github.com/matrix-org/matrix-spec/pull/1602/files#r1287281183):

> I know this is existing text, but it now seems even more confusing than before, so I think we need to clarify.
>
> 1. What is the mechanism by which streamless tracks are ignored? Or is this an instruction to clients to ignore such tracks?
> 2. what is that to do with `addTrack`'s parameters?
> 3. is it related to `sdp_stream_metadata`? if not, it should probably not be in the middle of the text about it?

zecakeh (commented):
Threads still opened from the previous review.

content/client-server-api/modules/voip_events.md
zecakeh: I'm not sure about it, it only applies to 3 out of 7 event types, contrary to the other fields.

zecakeh, line 172: I am not a fan of the formulation in places. For example there are several "i.e." which should not be needed if the sentences were reworked.

However there were no comments about it in the first review, so maybe it was not deemed necessary. Should I still take a crack at it?

data/event-schemas/schema/core-event-schema/sdp_stream_metadata.yaml
```diff
@@ -0,0 +1,27 @@
+type: object
```
richvdh, line 1: I don't think this belongs in `core-event-schema`

content/client-server-api/modules/voip_events.md
richvdh: I think the wording is *okay* as it stands, but there is certainly room for improvement. If you'd like to do so, that would be great!

content/client-server-api/modules/voip_events.md
richvdh: 1. Indeed. Can you change `will` to `MUST` (or, maybe more readably, `should`)?
2. Right. Yes. I think we should remove it.
3. Yes, I think it would make more sense at the end of the section, after the note on backwards-compatibility.

data/event-schemas/schema/core-event-schema/sdp_stream_metadata.yaml
```diff
@@ -0,0 +1,27 @@
...
+        - m.usermedia
+        - m.screenshare
+      description: |-
+        The purpose of the stream.
+
+        The possible values are:
+
+        * `m.usermedia` - stream that contains the webcam and/or microphone
+          tracks
+        * `m.screenshare` - stream with the screen-sharing tracks
```
richvdh, line 25:
```suggestion
        * `m.usermedia`: Stream that contains the webcam and/or microphone
          tracks.
        * `m.screenshare`: Stream with the screen-sharing tracks.
```

content/client-server-api/modules/voip_events.md
richvdh: Mmmhmm. Makes sense.

zecakeh: I have reworked part of the text.

content/client-server-api/modules/voip_events.md
zecakeh: Done

data/event-schemas/schema/core-event-schema/sdp_stream_metadata.yaml
zecakeh: I agree. Since it doesn't seem to belong directly in `schemas` either, because it only contains full event schemas, I took the liberty to create a `components` directory (named like this for similarity with the property that includes reusable bits in the OpenAPI spec)

data/event-schemas/schema/core-event-schema/sdp_stream_metadata.yaml
zecakeh: Applied

richvdh (approved):
looks great!

PR #1737, opened 2024-03-04
MSC4041 - Add Retry-After header notice.

Half-Shot:
### Pull Request Checklist

Signed-off-by: Will Hunt <matrix@half-shot.uk>

Preview: https://pr1737--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1737.feature
+++ b/changelogs/client_server/newsfragments/1737.feature
@@ -1,1 +1,1 @@
-Specify that the `Retry-After` header may be used to rate-limit a client.
+Specify that the `Retry-After` header may be used to rate-limit a client, as per [MSC4041](https://github.com/matrix-org/matrix-spec-proposals/pull/4041).
--- a/changelogs/client_server/newsfragments/1737.feature
+++ b/changelogs/client_server/newsfragments/1737.feature
@@ -1,0 +1,1 @@
+Specify that the `Retry-After` header may be used to rate-limit a client.
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -106,7 +106,7 @@ No resource was found for this request.
 
 `M_LIMIT_EXCEEDED`
 Too many requests have been sent in a short period of time. Wait a while
-then try again. See [Rate limiting section](#rate-limiting)
+then try again. See [Rate limiting](#rate-limiting).
 
 `M_UNRECOGNIZED`
 The server did not understand the request. This is expected to be returned with
@@ -226,13 +226,13 @@ return a standard error response of the form:
 }
 ```
 
-{{% added-in v="1.10" %}}
+{{% changed-in v="1.10" %}}: `retry_after_ms` property deprecated in favour of `Retry-After` header.
 
-The [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
-HTTP header SHOULD be specified by the server on any 429 code response.
+Homeservers SHOULD include a [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+for any response with a 429 status code.
 
 The `retry_after_ms` key MAY be included to tell the client how long
-they have to wait in milliseconds before they can try again. This field is
+they have to wait in milliseconds before they can try again. This property is
 deprecated, in favour of the `Retry-After` header.
 
 ### Transaction identifiers
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -226,13 +226,14 @@ return a standard error response of the form:
 }
 ```
 
-The `retry_after_ms` key SHOULD be included to tell the client how long
-they have to wait in milliseconds before they can try again.
-
 {{% added-in v="1.10" %}}
 
-The `Retry-After` HTTP header SHOULD be specified by the server on any 429
-code response. It SHOULD be preferred by clients, falling back to `retry_after_ms`.
+The [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+HTTP header SHOULD be specified by the server on any 429 code response.
+
+The `retry_after_ms` key MAY be included to tell the client how long
+they have to wait in milliseconds before they can try again. This field is
+deprecated, in favour of the `Retry-After` header.
 
 ### Transaction identifiers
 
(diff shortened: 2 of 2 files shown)


richvdh:
For links: MSC4041.

where did the template for that rather bizarre description come from?

Half-Shot:
> For links: [MSC4041](https://github.com/matrix-org/matrix-spec-proposals/issues/4041).
> 
> where did the template for that rather bizarre description come from?

Err, I attempted to use GitHub codespaces for the first time, and I won't be using it again if it does things like this.

content/client-server-api/_index.md
```diff
@@ -2540,6 +2540,9 @@ return a standard error response of the form:
 The `retry_after_ms` key SHOULD be included to tell the client how long
 they have to wait in milliseconds before they can try again.
 
+The `Retry-After` HTTP header MAY be specified by the server. When this header
```
Half-Shot, line 2543: It might be nice to link to the specification for this header?

```suggestion
The [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after) HTTP header MAY be specified by the server. When this header
```

richvdh, line 2543: My reading of the MSC is rather stronger than this MAY wording. It actually uses the word "shall". In any case, it seems like compliant homeservers now *need* to include the header. I'd also like us to explicitly use the word "deprecated" for the old property.

content/client-server-api/_index.md
```diff
@@ -2540,6 +2540,9 @@ return a standard error response of the form:
...
+is set, it SHOULD be preferred by clients over the `retry_after_ms` value.
```
richvdh, line 2544: The MSC also says that the header should be on *any* 429 response, not just `M_LIMIT_EXCEEDED`. I'm not sure if, in practice, that makes any difference, but it feels like the requirement to use `Retry-After` belongs under the "API standards" section rather than buried here where nobody will read it. (Possibly an *informative* reminder in this "Security" section could be useful too, but I feel like the *normative* requirement doesn't belong here.)

richvdh, line 2544: Final point: please could you add a paragraph starting `{{% changed-in v="1.10" %}}` (or possibly `added-in`, if you give it its own section) to document when this changed.

richvdh (changes requested):
Thank you!

content/client-server-api/_index.md
Half-Shot: Adjusted.

content/client-server-api/_index.md
Half-Shot: I've moved the section to API standards, agree it makes sense. The security section only had a bit on ratelimiting, so I've just purged the section rather than leaving it empty.

Half-Shot: Done.

content/client-server-api/_index.md
```diff
@@ -212,6 +212,29 @@ only read state (e.g.: `/sync`, get account data, etc).
...
+
+```json
+{
+  "errcode": "M_LIMIT_EXCEEDED",
+  "error": "string",
+  "retry_after_ms": integer (optional, deprecated)
+}
+```
+
+{{% added-in v="1.10" %}}
```
richvdh, line 229: I don't find this very clear about what exactly was added.

Better to add instead, at the end of the section:

```
{{% changed-in v="1.10" %}}: `retry_after_ms` property deprecated in favour of `Retry-After` header.
```

content/client-server-api/_index.md
```diff
@@ -212,6 +212,29 @@ only read state (e.g.: `/sync`, get account data, etc).
...
+
+The [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+HTTP header SHOULD be specified by the server on any 429 code response.
```
richvdh, line 232: Avoid the passive voice where possible. "Homeservers should include a `Retry-After` header ...".

content/client-server-api/_index.md
```diff
@@ -212,6 +212,29 @@ only read state (e.g.: `/sync`, get account data, etc).
...
+
+The `retry_after_ms` key MAY be included to tell the client how long
+they have to wait in milliseconds before they can try again. This field is
```
richvdh, line 235: Per https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#general: prefer "property" over "key" or "field".

richvdh (changes requested):

content/client-server-api/_index.md
```diff
@@ -212,6 +212,29 @@ only read state (e.g.: `/sync`, get account data, etc).
...
+
+{{% changed-in v="1.10" %}}: `retry_after_ms` property deprecated in favour of `Retry-After` header.
+
+Homeservers SHOULD include a [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
+for any response with a 429 status code.
+
+The `retry_after_ms` property MAY be included to tell the client how long
+they have to wait in milliseconds before they can try again. This property is
+deprecated, in favour of the `Retry-After` header.
+
```
richvdh, line 237: Two problems:

* Turns out this isn't formatted right with `{{% %}}`. My bad.
* It's random having this in the middle of the section. There's an argument for a simple "this section has been changed"-style thing at the start of the section (just the tag, with no extra words); but if we're going to explain what's changed, let's put it at the end.

```suggestion
Homeservers SHOULD include a [`Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after)
for any response with a 429 status code.

The `retry_after_ms` property MAY be included to tell the client how long
they have to wait in milliseconds before they can try again. This property is
deprecated, in favour of the `Retry-After` header.

{{< changed-in v="1.10" >}}: `retry_after_ms` property deprecated in favour of `Retry-After` header.
```

richvdh (changes requested):
I'd land this myself, but I don't seem to have permission. So, please can you make the requested change?

richvdh (approved):
thank you!

changelogs/client_server/newsfragments/1737.feature
```diff
@@ -0,0 +1 @@
+Specify that the `Retry-After` header may be used to rate-limit a client.
```
richvdh, line 1:
```suggestion
Specify that the `Retry-After` header may be used to rate-limit a client, as per [MSC4041](https://github.com/matrix-org/matrix-spec-proposals/pull/4041).
```

PR #1739, opened 2024-03-05
Deprecate the `font` HTML tag

zecakeh:
#### Rationale:

MSC4077 allows to deprecate HTML tags that are deprecated in the WHATWG standard without requiring an MSC, if they can be replaced by tags with the same feature.

[`font` is deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font) and can be replaced by `span` with the `data-mx-bg-color` and `data-mx-color` attributes.

So, HTML like this:

```html
<font data-mx-bg-color="#FF0000" color="#00FF00">Some green text on a red background</font>
```

Can also be written like this:

```html
<span data-mx-bg-color="#FF0000" data-mx-color="#00FF00">Some green text on a red background</span>
```

Preview: https://pr1739--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1739.clarification
+++ b/changelogs/client_server/newsfragments/1739.clarification
@@ -1,0 +1,1 @@
+The [font](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/font) element is deprecated in the HTML spec. Clients should prefer [span](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/span) with the `data-mx-bg-color` and `data-mx-color` attributes instead.
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -55,6 +55,14 @@ requiring a [Spec Change Proposal](/proposals) when they are deprecated in the
 [WHATWG HTML Living Standard](https://html.spec.whatwg.org/multipage/).
 {{% /boxes/note %}}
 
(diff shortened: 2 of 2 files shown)


KitsuneRal (dismissed):

tulir:
`<font>` is used more than `<strike>`. Strikethrough was always done using `<del>` in most clients, but the `/rainbow` command in the latest element web version outputs `<font color="...">` tags. This also effectively reverts MSC2422

It might still be a good idea to remove it, but I think it'd be good to have a note saying it was previously allowed. Maybe mark it as deprecated somehow instead of removing entirely

KitsuneRal:
A note saying that clients should stop using  `font` and switch to `span` would be nice indeed. As for `<font color=…>`, I surmise sticking to one way of colouring characters is a good thing.

zecakeh:
I added an info box for the deprecation.

tulir (approved):
This looks fine now, but I'm not completely sure if it should require an MSC due to being used by many clients

KitsuneRal (approved):
LGTM

turt2live (approved):
I'm not a terribly active client dev myself these days, but the change seems reasonable given the context. Clients are asked to *prefer* `span`, but we aren't breaking anything.

Opening issues on a few clients would be appreciated, but not required. This might be an SCT responsibility to encourage the ecosystem to upgrade anyways.

PR #1740, opened 2024-03-05
Disambiguate uses of `PublicRoomsChunk`

### LINKED ISSUE #1729: Conflicting use of PublicRoomsChunk for space hierarchy over federation
**Link to problem area**:
https://spec.matrix.org/v1.9/server-server-api/#get_matrixfederationv1hierarchyroomid

**Issue**
There are two different types with the name `PublicRoomsChunk`, with the only difference being that the latter has a `children_state` field which the other doesn't. In [ruma](https://ruma.io), it is assumed that the [former is for the `children` field](https://docs.rs/ruma/latest/ruma/api/federation/space/struct.SpaceHierarchyChildSummary.html), and the [latter is for the `room` field](https://docs.rs/ruma/latest/ruma/api/federation/space/struct.SpaceHierarchyParentSummary.html). However, in the example request in the spec, both the `children` and `room` fields use the `children_state` field, so it is unclear whether the example is correct and they should both have the `children_field`, ruma is correct and only the `room` field should have that, or even maybe the `children` field should have that and the `room` field shouldn't.
Either way, they really should either be merged (for the former solution), or have different names (for the others).

zecakeh:
Make sure that different objects don't share the same title.

Also make sure that the titles are not confusing: in the Client-Server API, the items in the space hierarchy response were named `ChildRoomsChunk`, but the first item returned should actually be the parent.

Fixes #1729.

Preview: https://pr1740--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1740.clarification
+++ b/changelogs/client_server/newsfragments/1740.clarification
@@ -1,0 +1,1 @@
+Disambiguate uses of `PublicRoomsChunk` in the `GET /hierarchy` endpoint.
--- a/changelogs/server_server/newsfragments/1740.clarification
+++ b/changelogs/server_server/newsfragments/1740.clarification
@@ -1,0 +1,1 @@
+Disambiguate uses of `PublicRoomsChunk` in the `GET /hierarchy` endpoint.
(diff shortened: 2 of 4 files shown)


clokep (approved):
🚀 Thanks!

PR #1741, opened 2024-03-05
Clarify that extra fields of space hierarchy children are not required

zecakeh:
There is no `children_state` field, the `room_type` is only set for spaces and the description of `allowed_room_ids` says that the field can be omitted.

Preview: https://pr1741--matrix-spec-previews.netlify.app


clokep (approved):
This looks correct.

clokep:
Kind of related to #1729.

Issue #1078, opened 2022-05-22
Only one of `sdpMLineIndex` or `sdpMid` should be required in `Candidate`


zecakeh:
**Link to problem area**: https://spec.matrix.org/v1.2/client-server-api/#mcallcandidates

**Issue**
According to the WebRTC spec, both [`sdpMid`](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-sdpmid) and [`sdpMLineIndex`](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-sdpmlineindex) are nullable, but one of those two values must be set:

> If both the [sdpMid](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidateinit-sdpmid) and [sdpMLineIndex](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidateinit-sdpmlineindex) members of candidateInitDict are null, [throw](https://webidl.spec.whatwg.org/#dfn-throw) a [TypeError](https://webidl.spec.whatwg.org/#exceptiondef-typeerror). ([*source*](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-sdpmlineindex))

Currently in the spec they are both marked as required.

**Expected behaviour**
Both fields should be nullable, explaining that at least one of the two is required.

In practice, both the [matrix-js-sdk](https://github.com/matrix-org/matrix-js-sdk/blob/25afb7cb3cc612656458c00569cf03c42277aae2/src/webrtc/call.ts#L2234-L2240) and [Element Android](https://github.com/vector-im/element-android/blob/109b381e2edea9463f8e10629ce983146b1ea1c1/vector/src/main/java/im/vector/app/features/call/webrtc/WebRtcCall.kt#L928-L930) check that at least one of those is set, respecting the WebRTC spec, which makes this a spec bug rather than requiring an MSC.

richvdh:
Still the case in https://spec.matrix.org/v1.6/client-server-api/#mcallcandidates.

@dbkr: can you confirm this?

zecakeh:
Fixed with #1742.

PR #1742, opened 2024-03-05
Clarify that sdpMid and sdpMLineIndex are not required in `m.call.candidates`

### LINKED ISSUE #1738: Inconsistent `m.call.candidates` candidate definition
**Issue**
The spec expects [m.call.candidates](https://spec.matrix.org/v1.9/client-server-api/#mcallcandidates) events to contain an array of `Candidates`, where each one has 3 **required** fields: `candidate`, `sdpMLineIndex` and `sdpMid`

[The spec also says](https://spec.matrix.org/v1.9/client-server-api/#end-of-candidates) that end of discovery should be marked by `An ICE candidate whose value is the empty string`, which contradicts the Candidate definition.

This in turn causes problems on implementations that are strict about required fields e.g. [Ruma](https://github.com/ruma/ruma/blob/f24cae17f50d140a0ff112cb3dc74a2053aa1df4/crates/ruma-events/src/call/candidates.rs#L10-L37) which throws deserialization errors in this case.

**Expected behaviour**
The spec should be clear about what Candidates can look like. Ideally they would've been typed but I guess that ship has sailed.

zecakeh:
MSC2746, merged in v1.7, introduced the [end-of-candidates candidate](https://spec.matrix.org/v1.9/client-server-api/#end-of-candidates), where only the `candidate` property is set to an empty string. However, the schema for the event was untouched in the [spec PR](https://github.com/matrix-org/matrix-spec/pull/1511).

Besides, the [WebRTC specification](https://www.w3.org/TR/webrtc/) says that [only one of those two fields is required when `candidate` is not empty](https://www.w3.org/TR/webrtc/#dom-peerconnection-addicecandidate) (step 3).

Fixes #1738 and #1078.

Preview: https://pr1742--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/schema/m.call.candidates.yaml
+++ b/data/event-schemas/schema/m.call.candidates.yaml
@@ -1,45 +1,40 @@
-{
-    "type": "object",
-    "description": "This event is sent by callers after sending an invite and by the callee after answering. Its purpose is to give the other party additional ICE candidates to try using to communicate.",
-    "allOf": [{
-        "$ref": "core-event-schema/room_event.yaml"
-    }],
-    "properties": {
-        "content": {
-            "type": "object",
-            "allOf": [{
-                "$ref": "core-event-schema/call_event.yaml"
-            }],
-            "properties": {
-                "candidates": {
-                    "type": "array",
-                    "description": "Array of objects describing the candidates.",
-                    "items": {
-                        "type": "object",
-                        "title": "Candidate",
-                        "properties": {
-                            "sdpMid": {
-                                "type": "string",
-                                "description": "The SDP media type this candidate is intended for."
-                            },
-                            "sdpMLineIndex": {
-                                "type": "number",
-                                "description": "The index of the SDP 'm' line this candidate is intended for."
-                            },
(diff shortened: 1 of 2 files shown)


data/event-schemas/schema/m.call.candidates.yaml
```diff
@@ -1,45 +1,48 @@
...
+                The index of the SDP 'm' line this candidate is intended for.
+
+                At least one of `sdpMid` or `sdpMLineIndex` is required, unless
+                this an end-of-candidates candidate.
+            candidate:
+              type: string
+              description: |-
+                The SDP 'a' line of the candidate.
+
+                If this is an end-of-candidates candidate, this is empty.
```
uhoreg, line 40: Looks good.  Just a couple minor nit-picks:

```suggestion
                If this is an [end-of-candidates](/client-server-api/#end-of-candidates) candidate, this the empty string.
```

link to the "End-of-candidates" section, and clarify what "empty" means

uhoreg (approved):
Thanks!

PR #1744, opened 2024-03-12
Specify that appservice login and register fail on incorrect as_tokens

tulir:
This is implemented in both Synapse and Dendrite, but it turned out Conduit falls back to UIA if it doesn't recognize the `as_token`

Preview: https://pr1744--matrix-spec-previews.netlify.app

### DIFF
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -437,8 +437,8 @@ an application service-defined namespace will receive the same
 defined the namespace as `exclusive`.
 
 If either endpoint is called with the `m.login.application_service` login type,
-but without a valid `as_token`, the endpoints will return `M_MISSING_TOKEN` or
-`M_UNKNOWN_TOKEN`.
+but without a valid `as_token`, the endpoints will return an error with the
+`M_MISSING_TOKEN` or `M_UNKNOWN_TOKEN` error code.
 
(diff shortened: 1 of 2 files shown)


content/application-service-api.md
```diff
@@ -436,6 +436,10 @@ an application service-defined namespace will receive the same
 `M_EXCLUSIVE` error code, but only if the application service has
 defined the namespace as `exclusive`.
 
+If either endpoint is called with the `m.login.application_service` login type,
```
richvdh, line 439: Might help to be a bit more explicit

```suggestion
If `/register` or `/login` is called with the `m.login.application_service` login type,
```

content/application-service-api.md
```diff
@@ -436,6 +436,10 @@ an application service-defined namespace will receive the same
...
+but without a valid `as_token`, the endpoints will return an error with the
+`M_MISSING_TOKEN` or `M_UNKNOWN_TOKEN` error code.
```
richvdh, line 441: should we specify the HTTP error code? 401?

tulir: Specified 401 and also added a link to the C-S spec which has the same code

richvdh (approved):
Thanks!

PR #1745, opened 2024-03-12
Factor out all the common parameters of the various /relations apis

dbkr:
This doesn't currently work, perhaps because the template needs to resolve refs?

Things this PR needs:
 * [ ] https://github.com/matrix-org/matrix-spec/pull/1749 merged
 * [ ] Replace other instances of params
 * [ ] Do the same to responses
 * [ ] CI happy

Preview: https://pr1745--matrix-spec-previews.netlify.app

PR #1746, opened 2024-03-12
Spec for MSC3981

dbkr:
This writes up https://github.com/matrix-org/matrix-spec-proposals/pull/3981

~~Hopefully this is relatively straightforward, apart from having to add the parameters and response field in all three places. I tried to factor these out but it seems references just aren't supported in the right places currently (see https://github.com/matrix-org/matrix-spec/pull/1745 for my efforts). Path parameters can't be optional, so it can't be done that way either.~~

Preview: https://pr1746--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1746.feature
+++ b/changelogs/client_server/newsfragments/1746.feature
@@ -1,1 +1,1 @@
-Add support for recursion on the /relations endpoints ([MSC3981](https://github.com/matrix-org/matrix-spec-proposals/pull/3981)).
+Add support for recursion on the /relations endpoints (MSC3981).
--- a/changelogs/client_server/newsfragments/1746.feature
+++ b/changelogs/client_server/newsfragments/1746.feature
@@ -1,0 +1,1 @@
+Add support for recursion on the /relations endpoints ([MSC3981](https://github.com/matrix-org/matrix-spec-proposals/pull/3981)).
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -2156,6 +2156,19 @@ following endpoint.
 This endpoint is particularly useful if the client has lost context on the aggregation for
 a parent event and needs to rebuild/verify it.
 
+When using the `recurse` parameter, note that there no way for a client to
+affect how much the server recurses. If the client decides that the server's
+recursion level is insufficient, it could, for example, perform the recursion
+manually, or disable whatever feature requires more recursion.
+
+Filters specified via `event_type` or `rel_type` will be applied to all events
+returned, whether direct or indirect relations. Events that would match the filter,
+but whose only relation to the original given event is through a non-matching
+intermediate event, will not be included. This means that supplying a `rel_type`
+parameter of `m.thread` is not appropriate for fetching all events in a thread since
+relations to the threaded events would be filtered out. For this purpose, clients should
+omit the `rel_type` parameter and perform any necessary filtering on the client side.
+
 {{% boxes/note %}}
 Because replies do not use `rel_type`, they will not be accessible via this API.
 {{% /boxes/note %}}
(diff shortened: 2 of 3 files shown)


richvdh:
I think I've fixed #1745, so let's hang fire hire for a minute

dbkr:
Updated with ref version

data/api/client-server/relations.yaml
```diff
@@ -294,6 +297,27 @@ components:
...
       schema:
         type: string
+    recurse:
+      in: query
+      name: recurse
+      x-addedInMatrixVersion: "1.10"
+      required: false
+      description: |-
+        Whether to additionally include events which only relate indirectly to the
+        given event, ie. events related to the root events via one or more direct relationships.
```
richvdh, line 307: https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#general

```suggestion
        given event, i.e. events related to the root events via one or more direct relationships.
```

richvdh: also: *two* or more relationships, no?

richvdh: also also: we don't seem to use the term "root event" elsewhere: should this be "parent event" ?

data/api/client-server/relations.yaml
```diff
@@ -294,6 +297,27 @@ components:
...
+
+        If set to `false`, only events which have direct a relation with the given
+        event will be included.
+
+        If set to `true`, all events which relate to the given event, or relate to
+        events that relate to the given event, will be included.
+
+        It is recommended that at least 3 levels of relationships are traversed.
```
richvdh, line 315: By whom? also, avoid passive voice.

```suggestion
        It is recommended that homeservers traverse at least 3 levels of relationships.
```

data/api/client-server/relations.yaml
```diff
@@ -308,6 +332,12 @@ components:
           description: |-
             An opaque string representing a pagination token. The absence of this token
             means this is the start of the result set, i.e. this is the first batch/page.
+        recursion_depth:
+          type: integer
+          description: |-
+            If the `recurse` parameter was supplied by the client, this response field is
+            mandatory and gives the actual depth to which the server recursed. The the client
```
richvdh, line 339:
```suggestion
            mandatory and gives the actual depth to which the server recursed. If the client
```

content/client-server-api/_index.md
```diff
@@ -2156,6 +2156,19 @@ following endpoint.
 This endpoint is particularly useful if the client has lost context on the aggregation for
 a parent event and needs to rebuild/verify it.
 
+When using the `recurse` parameter, note that there no way for a client to
+affect how much the server recurses. If the client decides that the server's
```
richvdh, line 2160:
```suggestion
control how far the server recurses. If the client decides that the server's
```

content/client-server-api/_index.md
```diff
@@ -2156,6 +2156,19 @@ following endpoint.
...
+recursion level is insufficient, it could, for example, perform the recursion
+manually, or disable whatever feature requires more recursion.
```
richvdh, line 2162:
```suggestion
itself, or disable whatever feature requires more recursion.
```

richvdh (approved):
a few nits but lgtm otherwise

data/api/client-server/relations.yaml
dbkr: Done in 1ba0ab9d. It's worth noting that the original was the text that the MSC proposed to add, so hopefully it's clear here that this is actually what it meant, but ideally we should have caught this in MSC review.

changelogs/client_server/newsfragments/1746.feature
```diff
@@ -0,0 +1 @@
+Add support for recursion on the /relations endpoints (MSC3981).
```
turt2live, line 1: I've applied the change to `main` as part of release checklist steps, but for future PRs please use a consistent changelog style (and apologies if these aren't documented anywhere - we should fix that if undocumented).

* Use backticks around endpoints, and indicate the request method.
* Use `, as per [MSC3981](...).` at the end of your changelog, linking to the MSC in question.

```diff
Add support for recursion on the `GET /relations` endpoints, as per [MSC3981](https://github.com/matrix-org/matrix-spec-proposals/pull/3981).
```

PR #1749, opened 2024-03-13
Support reference objects in operation parameters

richvdh:
According to https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#operation-object, the entries in `parameters` lists can be reference objects. We don't currently support that in our templates, so let's fix that.

Preview: https://pr1749--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/openapi/resolve-ref-object.html
+++ b/layouts/partials/openapi/resolve-ref-object.html
@@ -13,16 +13,12 @@ The input parameter is a dict with the following keys:
 * `path`: The path of the schema file containing the (potential) ref; used for resolving
   relative references.
 
-* `root_schema`: The top-level schema which contains the potential ref; use for resolving `#/`
-  references.
-
 Ref: https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.1.0.md#reference-object.
 
 */}}
 
 {{ $schema := .schema }}
 {{ $path := .path }}
-{{ $root_schema := .root_schema }}
 
 {{ $ret := $schema }}
 
--- a/layouts/partials/openapi/render-api.html
+++ b/layouts/partials/openapi/render-api.html
@@ -21,7 +21,7 @@
 
     {{/* note that a `paths` entry can be a $ref */}}
 
-    {{ $params := dict "endpoint" $endpoint "path" $path "root_schema" $api_data }}
+    {{ $params := dict "endpoint" $endpoint "path" $path }}
 
     {{ with $path_data.get }}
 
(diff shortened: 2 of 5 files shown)


layouts/partials/openapi/resolve-ref-object.html
```diff
@@ -0,0 +1,74 @@
+{{/*
+
+Handles OpenAPI "Reference Objects".
+
+Reference objects are JSON objects with a single property, `$ref` (and
```
zecakeh, line 5: It's actually a bit more complicated than that… Depending on the property it is used in, more fields might be overriden.

[This table](https://learn.openapis.org/referencing/overview.html#a-taxonomy-of-references) summarizes things.

richvdh: Right. I guess we are *specifically* targetting "reference objects" here, so it's correct that `summary` and `description` are preserved?

I guess we'd need to make this logic more generic if we wanted to use it in other places, but that could be a future change.

zecakeh: Sure, what made me react here is more the comment in the resolve-refs partial that says it should be replaced by this one, but in its current state it's not possible.

richvdh: ah ok, maybe we can clarify that comment.

layouts/partials/openapi/resolve-ref-object.html
```diff
@@ -0,0 +1,69 @@
...
+    {{/*
+    OpenAPI spec says that "summary" and "description" from the reference object override
+    the values from the referenced component.
+    */}}
+    {{ if isset $schema "summary" }}
+        {{ $ret = merge $ret (dict "summary" $schema.summary) }}
+    {{ end }}
+    {{ if isset $schema "description" }}
+        {{ $ret = merge $ret (dict "description" $schema.summary) }}
+    {{ end }}
```
zecakeh, line 66: If we want to be more strict with how we resolve refs, should we throw an error or log a warning if `$schema` and `$ret` have properties of the same name, other than those? That would allow to know sooner what is wrong if someone expects the ref to be overwritten by the schema, which we often do, but requires the use of `allOf`.

richvdh:
Having discussed this again with @zecakeh:

As things currently stand, we implement $ref by calling `resolve-refs`, which recursively expands `$ref` without concern about where the `$ref` actually is. The problem with that is then it's unclear whose job it is to call resolve-refs. I was hoping to resolve that by moving towards a model where we only expand `$ref` in places where we know it is valid. But I'm now wondering if that is actually worthwhile.

Instead, maybe we should just call resolve-refs in the top-level template (`layouts/shortcodes/http-api.html`, etc),  and rely on the validator to actually enforce the rules. This is also what happens in `scripts/dump-openapi.py`.

It's also worth noting that we currently use `$ref` in a place where OpenAPI doesn't permit it (in the bodies of examples), and, assuming we want to keep that power (I think we do) we probably need the recursive `resolve-refs` for that.

So, I'm going to abandon this approach, in favour of @zecakeh's solution over at #1751.

PR #1751, opened 2024-03-13
Add support for `$ref` URIs containing fragments in OpenAPI definitions and JSON schemas

zecakeh:
Based on #1745.

There is a first commit to add support for fragments in `$ref` URIs, to solve (simple) JSON Pointers. It also adds support for [Reference Objects](https://spec.openapis.org/oas/v3.1.0#reference-object) for OpenAPI request parameters and responses.

Preview: https://pr1751--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/json-schema/resolve-refs.html
+++ b/layouts/partials/json-schema/resolve-refs.html
@@ -1,7 +1,10 @@
 {{/*
 
   Resolves the `$ref` JSON schema keyword, by recursively replacing
-  it with the object it points to.
+  it with the object it points to, given:
+
+  * `schema`: the schema where the references should be resolved
+  * `path`: the path of the file containing the schema
 
   This template uses [`Scratch`](https://gohugo.io/functions/scratch/)
   rather than a normal `dict` because with `dict` you can't replace key values:
@@ -20,8 +23,12 @@
     {{ $scratch.Set "result_map" dict }}
 
     {{ $ref_value := index $schema "$ref"}}
-    {{ if $ref_value}}
-        {{ $full_path := path.Join $path $ref_value }}
+    {{ if $ref_value }}
+        {{ $uri := urls.Parse $path }}
+        {{ $ref_uri := urls.Parse $ref_value }}
+        {{ $full_uri := $uri.ResolveReference $ref_uri }}
+
+        {{ $full_path := strings.TrimPrefix "/" $full_uri.Path }}
         {{/*
           Apparently Hugo doesn't give us a nice way to split the extension off a filename.
         */}}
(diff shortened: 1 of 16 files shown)


richvdh:
oh no :facepalm: 

This is the same thing as I did in https://github.com/matrix-org/matrix-spec/pull/1749 :(

zecakeh:
I saw that, well it is kind of completely different though

richvdh:
@zecakeh As I've written over at #1749, this approach seems sensible, but I'd like to land it independently of #1745. Could you rebase it without that?

zecakeh:
Alright, I removed the commit with the OpenAPI definition change. I added a commit to also support refs with fragments in the CI scripts.

layouts/partials/json-schema/resolve-refs.html
```diff
@@ -1,7 +1,10 @@
 {{/*
```
zecakeh, line 1: Should we do like in #1749 and allow to override properties named `description` and `summary`?

scripts/helpers.py
```diff
@@ -0,0 +1,87 @@
...
+            ref = load_file_from_uri(full_ref_uri)
+
+            if fragment:
+                # The fragment should be a JSON Pointer
+                keys = fragment.strip('/').split('/')
+                for key in keys:
+                    ref = ref[key]
+
+            # Check that the references in *this* file are valid
+            result = resolve_references(urllib.parse.urlsplit(full_ref_uri).path, ref)
```
zecakeh, line 56: Contrary to the `resolve-refs` partial, this does not keep non-duplicate keys in the original schema, should it?

layouts/partials/json-schema/resolve-refs.html
richvdh: No, I don't think that's necessary. The validator will complain if anyone tries to override other properties, I think?

layouts/shortcodes/http-api.html
```diff
@@ -21,6 +21,6 @@
 
 {{ $api_data := index .Site.Data.api .Params.spec .Params.api }}
 {{ $base_url := (index $api_data.servers 0).variables.basePath.default }}
-{{ $path := delimit (slice "api" $spec) "/" }}
+{{ $path := delimit (slice "api" $spec $api) "/" }}
```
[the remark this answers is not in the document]
richvdh: well, probably, yes. But we don't necessarily need to change that today.

richvdh (approved):
lgtm

layouts/partials/json-schema/resolve-refs.html
zecakeh: I doesn't, afaict.

richvdh: oh. bother. Well, maybe something to think about adding into our validation scripts in future.

PR #1754, opened 2024-03-19
Fix `event-group` shortcode for changes in `resolve-refs` partial

zecakeh:
This was broken because "path" needs to point to the file now, and results in `sdp_stream_metadata` having an unresolved ref in `m.call.invite`, `m.call.answer` and `m.call.negotiate`.

cc @richvdh 

Preview: https://pr1754--matrix-spec-previews.netlify.app


richvdh (approved):
Oops. Thanks.

PR #1756, opened 2024-03-19
Break out non-JSON request/response content types as tables

turt2live:
Currently we display this as a table like "image/png|image/jpeg" and description on a single line, but we're using a table. This breaks the join out to individual rows.

See `/client-server-api/#get_matrixmediav3thumbnailservernamemediaid` and `/client-server-api/#post_matrixmediav3upload` for examples.

Preview: https://pr1756--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1756.clarification
+++ b/changelogs/client_server/newsfragments/1756.clarification
@@ -1,0 +1,1 @@
+Clearly indicate that each `Content-Type` may have distinct behaviour on non-JSON requests/responses.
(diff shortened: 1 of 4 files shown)


anoadragon453 (approved):
I don't visually see a difference for `/client-server-api/#post_matrixmediav3upload`, but `/client-server-api/#get_matrixmediav3thumbnailservernamemediaid` is a marked improvement.

Thank you!

turt2live:
hmm, good point on `/upload`: it has a request body table, but only one type 🤦 
image

PR #1757, opened 2024-03-19
Spec `?animated` on `/thumbnail`

turt2live:
Better with https://github.com/matrix-org/matrix-spec/pull/1756

See https://github.com/matrix-org/matrix-spec-proposals/pull/2705

Preview: https://pr1757--matrix-spec-previews.netlify.app


data/api/client-server/content-repo.yaml
```diff
@@ -633,6 +633,28 @@ paths:
           schema:
             type: boolean
             default: false
+        - in: query
+          name: animated
+          x-addedInMatrixVersion: "1.10"
```
richvdh, line 638:
```suggestion
          x-addedInMatrixVersion: "1.11"
```

richvdh (commented):
Could you merge the latest changes from `main` so we can see how this looks in the preview?

data/api/client-server/content-repo.yaml
turt2live: applied manually - thanks!

data/api/client-server/content-repo.yaml
```diff
@@ -644,14 +666,33 @@ paths:
...
           content:
             image/jpeg:
               schema:
                 # This is a workaround for us not being able to say the response is required.
                 description: "**Required.** The bytes for the thumbnail."
             image/png:
               schema:
-                description: "**Required.** The bytes for the thumbnail."
+                x-changedInMatrixVersion:
+                  "1.11": The PNG may be of the APNG variety if animation is supported and requested.
```
richvdh, line 680: For the record (no action required): these don't seem to show up in the rendered spec, though that's no reason not to add them.

richvdh (approved):

data/api/client-server/content-repo.yaml
turt2live: https://github.com/matrix-org/matrix-spec/issues/1774

PR #1759, opened 2024-03-19
Update release documentation (Q2 2024 edition)

turt2live:
All of this stuff is pretty subjective, but it's good to keep it updated, I think. Much of it will change again.

Preview: https://pr1759--matrix-spec-previews.netlify.app


meta/releasing.md
```diff
@@ -9,45 +9,44 @@ machinery works.
 The spec is released each calendar quarter. The target release dates are within the
 following ranges:
 
-* Q1: January 20-27 (critically, before FOSDEM).
-* Q2: May 20-27.
-* Q3: August 20-27.
+* Q1: January 20-27 (if needed before FOSDEM) or February 21-28.
```
richvdh, line 12: well I guess that means we're only a month late, so yay?

Honestly I question the value of attempting to set dates like this. Personally I would just say "we aim for each release to happen at most four months after the previous release, but they may happen sooner to deal with things like FOSDEM" and be done with it.

meta/releasing.md
```diff
@@ -115,12 +114,16 @@ release.
...
 
-From time to time we'll need to update a release in the wild. Examples include fixing typos,
-updating build machinery, etc. Typically it is not considered a good idea to patch a release
-more than 1 month after the original release date - this is because the administrative effort
-is typically best reserved for the next release cycle.
+Patch releases are used to fix the most recent release on record. Typically a patch
+release will be deployed if there is an issue with the build machinery, a factual
+error is introduced by the release, or there are notable clarity issues introduced
+by the release which may affect implementation. In all cases, patch releases should
+*not* be used if more than 2-4 weeks have passed since the release.
```
richvdh, line 121: This is a bit strong. I preferred the previous formulation of "they aren't a great idea".

meta/releasing.md
```diff
@@ -9,45 +9,44 @@ machinery works.
...
+Spec releases do not currently have attached themes, though when planning a release
+a broad theme may be considered. Ideally, each release contains a "hero feature"
+which is highlighted in the later blog post.
 
-* Q1: Massive feature release, if possible. This generally happens thanks to FOSDEM.
-* Q2: Regular feature release, if possible.
-* Q3: Momentum-continuing feature release, if possible.
-* Q4: Preferably a maintenance release, but will accept features per normal.
+Maintenance-only releases are discouraged, but typical in Q4 to give implementations
+a small bit of reprieve from feature development.
```
richvdh, line 49: Are they discouraged? I mean, sure, a headline feature is nice for marketing and publicity, but *discouraging* a release without such a feature seems a bit harsh.

richvdh (approved):
Some thoughts, but nothing blocking.

meta/releasing.md
turt2live: There's still value in releasing every quarter instead of on an interval, though yes, the dates are increasingly becoming lies. We generally aim for the middle of the quarter.

meta/releasing.md
turt2live: Discouraged might be the wrong word here. This is trying to say there should always be a plan of some sort to avoid organic releases. Maybe we discourage organic releases?

meta/releasing.md
turt2live: the language is deliberately stronger, yea. I don't think we want to do patch releases all that often, and the previous guidance allowed things like typos to go through. When each release is 3-4 months away (or less, in some cases), 1 month of patch time is already pretty significant.

turt2live: correction: 2 months between releases, because quarters are weird.

meta/releasing.md
richvdh: I'm not entirely sure what you mean by an organic release, but what's so bad about them?

turt2live: Organic releases are the ones where we didn't really have a plan and just cut the spec on schedule with whatever random changes happened to be there. Preferably, we put a bit of thought and planning into the spec's progression instead of letting it coast through releases with just clarification changes.

richvdh: > Preferably, we put a bit of thought and planning into the spec's progression instead of letting it coast through releases with just clarification changes.

I mean sure, planning for the spec's progression is a good thing, but if, for whatever reason, we end up with a release of mostly maintenance stuff, that seems fine?

I just don't really feel like this belongs in the release process.

But not a strong opinion.

meta/releasing.md
turt2live: I've dropped the dates and fixed the intro sentence

meta/releasing.md
turt2live: paragraph dropped

meta/releasing.md
turt2live: I've softened it a bit

PR #1763, opened 2024-03-21
Do not allow setting m.push_rules account data directly.

clokep:
Spec PR for https://github.com/matrix-org/matrix-spec-proposals/pull/4010

See https://pr1763--matrix-spec-previews.netlify.app/client-server-api/#server-behaviour-10

Preview: https://pr1763--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1763.clarification
+++ b/changelogs/client_server/newsfragments/1763.clarification
@@ -1,1 +1,1 @@
-Clarify that the `m.push_rules` account data type cannot be set using the `/account_data` API.
+Clarify that the `m.push_rules` account data type cannot be set using the `/account_data` API, as per [MSC4010](https://github.com/matrix-org/matrix-spec-proposals/pull/4010).
--- a/changelogs/client_server/newsfragments/1763.clarification
+++ b/changelogs/client_server/newsfragments/1763.clarification
@@ -1,0 +1,1 @@
+Clarify that the `m.push_rules` account data type cannot be set using the `/account_data` API.
--- a/content/client-server-api/modules/account_data.md
+++ b/content/client-server-api/modules/account_data.md
@@ -36,6 +36,4 @@ This applies to both global and room-specific account data.
 list.
 {{% /boxes/note %}}
 
-Servers must allow clients to read the above event types as normal. Note that
-`m.fully_read` only exists for room-specific account data and `m.push_rules`
-only exists for global account data.
+Servers must allow clients to read the above event types as normal.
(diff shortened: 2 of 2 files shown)


content/client-server-api/modules/account_data.md
```diff
@@ -28,4 +28,9 @@ These events can also be received in a `/events` response or in the
 
 Servers MUST reject clients from setting account data for event types
 that the server manages. Currently, this only includes
-[m.fully\_read](#mfully_read).
+[`m.fully_read`](#mfully_read) and [`m.push_rules`](#mfully_read).
+This applies to both global and room-specific account data.
```
clokep, line 32: Should we clarify that this corresponds to the 405 error code?

turt2live, line 32:
```suggestion
[`m.fully_read`](#mfully_read) and [`m.push_rules`](#mfully_read).
This applies to both global and room-specific account data.

{{% boxes/note %}}
{{% changed-in v="1.10" %}} `m.push_rules` was added to the rejection
list.
{{% /boxes/note %}}
```

changelogs/client_server/newsfragments/1763.clarification
```diff
@@ -0,0 +1 @@
+Clarify that the `m.push_rules` account data type cannot be set using the `/account_data` API.
```
turt2live, line 1:
```suggestion
Clarify that the `m.push_rules` account data type cannot be set using the `/account_data` API, as per [MSC4010](https://github.com/matrix-org/matrix-spec-proposals/pull/4010).
```

content/client-server-api/modules/account_data.md
turt2live: I think so, yes.

content/client-server-api/modules/account_data.md
```diff
@@ -28,4 +28,9 @@ These events can also be received in a `/events` response or in the
...
+
+Servers must allow clients to read the above event types as normal. Note that
+`m.fully_read` only exists for room-specific account data and `m.push_rules`
+only exists for global account data.
```
turt2live, line 36: This is a bit confusing when paired with the line above about how *both* types of account data are banned. I think we can get away with links to the respective modules (as above).

```suggestion
Servers must allow clients to read the above event types as normal.
```

turt2live (commented):
generally lgtm - thanks for writing it up! A few nitpicks and different word suggestions.

clokep:
I think this is ready for another look!

turt2live (approved):
Thanks!

PR #1765, opened 2024-03-22
Fix npm publishing being broken in CI

t3chguy:
Fixes https://github.com/matrix-org/matrix-spec/issues/1681

Preview: https://pr1765--matrix-spec-previews.netlify.app


changelogs/internal/newsfragments/1765.misc
turt2live, line 1: we don't have `.misc` changelogs - use `.clarification` as confusing as that is please :)

turt2live (approved):
otherwise lgtm - thanks!

changelogs/internal/newsfragments/1765.misc
t3chguy: omfg how do I keep doing this

PR #1772, opened 2024-03-29
Fix security schemes in OpenAPI definitions

zecakeh:
There are not a lot of different changes but there is a lot of search & replace.

The changes are:
- Add the `Authorization: Bearer` security scheme as a different possibility to the query parameter for the client-server and identity APIs.
- Fix the definition of the security scheme of the application service, to take advantage of the built-in OpenAPI schemes
- Use `$ref` properly to reuse the security schemes. It's a bit more verbose as each scheme needs to be imported individually.

This can be reviewed per commit.

There is no difference in the spec output, only in the OpenAPI output.

Preview: https://pr1772--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/application_service/newsfragments/1772.clarification
+++ b/changelogs/application_service/newsfragments/1772.clarification
@@ -1,0 +1,1 @@
+Fix the OpenAPI definition of the security schemes.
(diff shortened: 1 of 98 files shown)


KitsuneRal (approved):
Overall, LGTM; I wonder though if we should deprecate usage of the query to pass access tokens, because security. I admit it’s a subject for an MSC, so just an aside.

PR #1773, opened 2024-03-29
Simplify uses of `resolve-refs` partial

zecakeh:
As discussed in https://github.com/matrix-org/matrix-spec/pull/1749#issuecomment-2003742301.

- Since it is a recursing partial, we can call it once we get the data, and it will solve everything in the tree. In the end only the `http-api` shortcode was not doing that already.
- Enable a lint in the OpenAPI validator to only allow `$ref` where it is supposed to be. It depends on #1772 for the lint to pass.
- Document our use of `$ref` inside examples to allow to compose them from other examples, although it is not supported by the OpenAPI and JSON Schema specs.
- Fix a call to `resolve-refs`, even if it doesn't seem to actually be needed because calls to this shortcode point to schemas without `$ref` (which is why the bug was unnoticed). 

Preview: https://pr1773--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/shortcodes/http-api.html
+++ b/layouts/shortcodes/http-api.html
@@ -24,6 +24,5 @@
 {{ $path := delimit (slice "api" $spec $api) "/" }}
 
 {{ $api_data = partial "json-schema/resolve-refs" (dict "schema" $api_data "path" $path) }}
-{{ $api_data := partial "json-schema/resolve-allof" $api_data }}
 
 {{ partial "openapi/render-api" (dict "api_data" $api_data "base_url" $base_url) }}
--- a/layouts/shortcodes/http-api.html
+++ b/layouts/shortcodes/http-api.html
@@ -23,4 +23,7 @@
 {{ $base_url := (index $api_data.servers 0).variables.basePath.default }}
 {{ $path := delimit (slice "api" $spec $api) "/" }}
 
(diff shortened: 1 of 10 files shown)


layouts/shortcodes/http-api.html
```diff
@@ -23,4 +23,7 @@
 {{ $base_url := (index $api_data.servers 0).variables.basePath.default }}
 {{ $path := delimit (slice "api" $spec $api) "/" }}
 
-{{ partial "openapi/render-api" (dict "api_data" $api_data "base_url" $base_url "path" $path) }}
+{{ $api_data = partial "json-schema/resolve-refs" (dict "schema" $api_data "path" $path) }}
+{{ $api_data := partial "json-schema/resolve-allof" $api_data }}
```
richvdh, line 27: why do we need to call `resolve-allof` here now?

zecakeh:
Rebased on main to pass CI

layouts/shortcodes/http-api.html
zecakeh: Sorry, it must be a leftover of when I was trying stuff.

zecakeh: I removed it.

richvdh:
> Rebased on main to pass CI

could you merge rather than rebase next time, please. Rebasing makes it hard to see what has changed since an earlier review.

richvdh (approved):
lgtm

PR #1776, opened 2024-04-03
Remove enum for `POST /login` `type` definition

zecakeh:
Since the enum is not exhaustive, improve the description of the property instead.

### DIFF
--- a/data/api/client-server/login.yaml
+++ b/data/api/client-server/login.yaml
@@ -109,10 +109,13 @@ paths:
               properties:
                 type:
                   type: string
-                  enum:
-                    - m.login.password
-                    - m.login.token
-                  description: The login type being used.
+                  description: |-
+                    The login type being used.
+
+                    This must be a type returned in one of the flows of the
+                    response of the [`GET /login`](/client-server-api/#get_matrixclientv3login)
+                    endpoint, like `m.login.password`, `m.login.token` or
+                    `m.login.sso`.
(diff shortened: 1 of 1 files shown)


data/api/client-server/login.yaml
```diff
@@ -109,10 +109,13 @@ paths:
...
-                    - m.login.password
-                    - m.login.token
-                  description: The login type being used.
+                  description: |-
+                    The login type being used.
+
+                    This must be a type returned in one of the flows of the
+                    response of the [`GET /login`](/client-server-api/#get_matrixclientv3login)
+                    endpoint, like `m.login.password`, `m.login.token` or
+                    `m.login.sso`.
```
KitsuneRal, line 118: I might be confusing something but I think `m.login.sso` is never to be found in `POST /login`, clients should redirect to a dedicated endpoint instead. Am I smoking stuff?

KitsuneRal (commented):
I agree with the motion (the list is certainly open-ended here) but `m.login.sso` is probably a poor example still.

data/api/client-server/login.yaml
richvdh: Yes, I think that's right.

zecakeh: Indeed, I got influenced by the issue without thinking. It is fixed now.

richvdh (approved):

PR #1781, opened 2024-04-09
Update github-labels.rst

richvdh:
Fix formatting

Preview: https://pr1781--matrix-spec-previews.netlify.app


turt2live (approved):
don't forget an internal or misc changelog (whatever we called it)

PR #1786, opened 2024-04-16
Upgrade CI scripts dependencies

zecakeh:
Will allow us to benefit from future fixes in JSON Schema validation.

Preview: https://pr1786--matrix-spec-previews.netlify.app

### DIFF
--- a/scripts/check-event-schema-examples.py
+++ b/scripts/check-event-schema-examples.py
@@ -43,6 +43,12 @@ except ImportError as e:
     raise
 
 try:
+    import referencing
+except ImportError as e:
+    import_error("referencing", "referencing", "referencing", e)
+    raise
+
+try:
     import yaml
 except ImportError as e:
     import_error("yaml", "PyYAML", "yaml", e)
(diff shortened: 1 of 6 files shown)


scripts/check-event-schema-examples.py
```diff
@@ -42,6 +42,12 @@ def import_error(module, package, debian, error):
     import_error("jsonschema", "jsonschema", "jsonschema", e)
     raise
 
+try:
+    import referencing
```
richvdh, line 46: Since we depend on it, we should add this to `requirements.txt`, even though it will be pulled in as a transitive dependency in practice.

scripts/helpers.py
```diff
@@ -84,4 +85,14 @@ def load_file_from_uri(path):
         else:
             # We have to assume it's YAML because some of the YAML examples
             # do not have file extensions.
-            return yaml.safe_load(f)
\ No newline at end of file
+            return yaml.safe_load(f)
+
+def load_resource_from_uri(path):
+    """Load a JSON or YAML JSON Schema from a file:// URI.
```
richvdh, line 91:
```suggestion
    """Load a JSON or YAML JSON Schema, as a `referencing.Resource` object, from a file:// URI.
```

richvdh (commented):
Looks sensible other than a couple of nits

scripts/check-event-schema-examples.py
zecakeh: Added

scripts/helpers.py
zecakeh: Applied

PR #1788, opened 2024-04-16
Bump minimum Hugo version in README

zecakeh:
To match the one in `config.toml` changed in #1775.

Preview: https://pr1788--matrix-spec-previews.netlify.app

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ place after an MSC has been accepted, not as part of a proposal itself.
 
 1. Install the extended version (often the OS default) of Hugo:
    <https://gohugo.io/getting-started/installing>. Note that at least Hugo
-   v0.110.0 is required.
+   v0.113.0 is required.
 
    Alternatively, use the Docker image at
    https://hub.docker.com/r/klakegg/hugo/. (The "extended edition" is required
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -61,7 +61,7 @@ place after an MSC has been accepted, not as part of a proposal itself.
 
 1. Install the extended version (often the OS default) of Hugo:
    <https://gohugo.io/getting-started/installing>. Note that at least Hugo
-   v0.110.0 is required.
+   v0.113.0 is required.
```
richvdh, line 64: Is it though? As I said in the chat, there's no need to bump this unless we're aware of problems on v0.110.0.

zecakeh: The idea here is to have the same version as in config.toml. The goal of both is to declare the minimum supported version of Hugo for the website.

zecakeh: What I'm saying is either we need to downgrade the version in config.toml, or we need to upgrade it in the README.

richvdh: Right, sorry, I misunderstood the purpose of the config.toml setting.

richvdh (approved):
Thanks!

PR #1789, opened 2024-04-16
Fix property type resolution in render-object-table

zecakeh:
The split was not clear between property-type and type-or-title, so it was not obvious which partial should be called for recursion. That resulted in an error where type-or-title was only called for objects and array items, even if it also resolves arrays of types.

This makes the split clearer. property-type must be called for any schema, and object-type-or-title is only called for object schemas.

This fixes some rendering in the CS API, for example the `displayname` of `m.room.member`:

Before:

image

After:

image

Also the `value` of a `PushCondition`:

Before:

image

After:

image

Preview: https://pr1789--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1789.clarification
+++ b/changelogs/internal/newsfragments/1789.clarification
@@ -1,0 +1,1 @@
+Fix property type resolution in `render-object-table` partial.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
Thanks!

PR #1791, opened 2024-04-16
Define opaque identifier grammar

richvdh:
Since we already have three of these, and I'm about to add a fourth, let's pull
it out to a common definition.

We could, of course, keep defining the grammar each time it's used, but
defining it in an appendix helps us be consistent for future API design.

Preview: https://pr1791--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/appendices/newsfragments/1791.clarification
+++ b/changelogs/appendices/newsfragments/1791.clarification
@@ -1,0 +1,1 @@
+Define 'Opaque Identifier Grammar'.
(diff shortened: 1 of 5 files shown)


clokep (approved):
Looks reasonable!

clokep:
Needs a newsfragment though.

richvdh:
> Needs a newsfragment though.

Sigh. I even wrote one, then forgot to `git add` it :/

PR #1793, opened 2024-04-17
Factor out common definition of `Tag` type

richvdh:
... and remove spurious `additionalProperties: true`. Aside from that being the default for all JSON objects, it's also misleading since we don't normally add additional properties to tags.


zecakeh (approved):
LGTM

PR #1797, opened 2024-04-23
Clean up unecessary `allOf`s

zecakeh:
Most of them were unnecessary to begin with. A few of them were necessary given the definition, but it turns out that the fields were duplicates to the ones in the `$ref`.

Preview: https://pr1797--matrix-spec-previews.netlify.app


richvdh (approved):
Yup. I think an earlier version of our tooling only supported `$ref` within an `allOf`.

PR #1798, opened 2024-04-23
Include information about additionalProperties in object tables

richvdh:
Currently, if we have an object which has `additionalProperties` in addition to `properties`, that information gets lost. This PR seeks to address that.

The individual commits should be separately reviewable.

Examples:

[old](https://spec.matrix.org/v1.10/client-server-api/#well-knownmatrixclient_discovery-information):
image

[new](https://pr1798--matrix-spec-previews.netlify.app/client-server-api/#well-knownmatrixclient_discovery-information):
image

[old](https://spec.matrix.org/v1.10/client-server-api/#_matrixclientv1loginget_token_authentication-data):
image

[new](https://pr1798--matrix-spec-previews.netlify.app/client-server-api/#_matrixclientv1loginget_token_authentication-data)
image

[old](https://spec.matrix.org/v1.10/client-server-api/#mroompower_levels_notifications)
image

[new](https://pr1798--matrix-spec-previews.netlify.app/client-server-api/#mroompower_levels_notifications)
image

Some of those descriptions could use improvement, but that is a separate job.

Preview: https://pr1798--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/partials/json-schema/resolve-additional-types.html
+++ b/layouts/partials/json-schema/resolve-additional-types.html
@@ -29,6 +29,8 @@
  *
  *     * title
  *     * properties
+ *     * additionalProperties
+ *     * patternProperties
  *     * required
  *     * enum
  *     * anchor
@@ -45,27 +47,31 @@
     "schema" .schema
     "anchor_base" .anchor_base
     "name" (.name | default .schema.title | default "<untitled object>")
+    "top_level" true
 ) }}
 {{ return $res.objects }}
 
 /*
  * A helper for the resolve-additional-types partial.
  *
- * Takes the same inputs as resolve-additional-types itself, except that `name` is mandatory.
+ * Takes the same inputs as resolve-additional-types itself, except:
+ *   * `name` is mandatory.
+ *   * `top_level`, if set, indicates that this is the top-level object.
  *
  * Returns a dict containing:
  *
  *   * `objects`: The array of object schema definitions.
  *
- *   * `schema`: An updated copy of the `schema` input (eg, nested `allOf`
- *     entries are expanded, and an `anchor` may be added). If the input
- *     `schema` was itself an object, this will be the same as the first entry
- *     in `objects`.
+ *   * `schema`: An updated copy of the `schema` input (eg, an `anchor` may be added).
+ *     If the input `schema` was itself an object that we should create a table for,
+ *     this will be the same as the first entry in `objects`.
  */
 {{ define "partials/resolve-additional-types-inner" }}
     {{ $this_object := .schema }}
     {{ $anchor_base := .anchor_base }}
     {{ $name := .name }}
+    {{ $top_level := .top_level }}
+
     {{ $all_objects := slice }}
 
     {{ if eq $this_object.type "object" }}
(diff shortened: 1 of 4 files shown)


layouts/partials/json-schema/resolve-additional-types.html
```diff
@@ -45,27 +47,31 @@
...
- * Takes the same inputs as resolve-additional-types itself, except that `name` is mandatory.
+ * Takes the same inputs as resolve-additional-types itself, except:
+ *   * `name` is mandatory.
+ *   * `top_level`, if set, indicates that this is the top-level object.
  *
  * Returns a dict containing:
  *
  *   * `objects`: The array of object schema definitions.
  *
- *   * `schema`: An updated copy of the `schema` input (eg, nested `allOf`
```
richvdh, line 60: Since #1787, `allOf` entries are now resolved at the top level before we get here, rather than inside `resolve-additional-types`.

layouts/partials/openapi/render-object-table.html
```diff
@@ -9,11 +9,11 @@
   * `properties`: optional dictionary of the properties to list, each given as:
     `property_name` : `property_data`
 
-  * `additionalProperties`: optional dictionary for properties with undefined
-    names, in the same format as `property_data`
+  * `additionalProperties`: an OpenAPI schema document for additional properties
```
zecakeh, line 12: The term "OpenAPI schema document" sounds weird to me. I have seen "OpenAPI Document" or "OpenAPI Definition", which are not right here. I believe that "JSON schema" is both correct here and easier to understand.

layouts/partials/openapi/render-object-table.html
```diff
@@ -52,9 +51,40 @@
...
+      `additionalProperties: false` means "you're not allowed to include any unspecced properties in your object". We
+      may want to consider how to display that; for now we just ignore it.
+
+      TODO: support `patternProperties` here.
+    */}}
+    {{ if reflect.IsMap .additionalProperties }}
+
+ <tr>
+  <td>&lt;Other properties&gt;</code></td>
+  <td><code>{{ partial "partials/property-type" .additionalProperties }}</code></td>
```
zecakeh, line 72: This needs to be marked as safe HTML for links to work:

```suggestion
  <td><code>{{ partial "partials/property-type" .additionalProperties | safeHTML }}</code></td>
```

zecakeh (commented):
I almost worked on that today, until I realized we needed a way to show `patternProperties` too, which is why I worked on #1796 first :wink:.

 There is at least one TODO comment to remove at:

https://github.com/matrix-org/matrix-spec/blob/26ce3929b4ae238f080bfc368fe3da092a890f2c/data/api/identity/v2_terms.yaml#L55

richvdh:
> I almost worked on that today, until I realized we needed a way to show patternProperties too, which is why I worked on https://github.com/matrix-org/matrix-spec/pull/1796 first 😉.

Great minds think alike :innocent:. It's not the first time the two of us have ended up working on the same thing. I think we should probably try to mention what we're planning to work on in `#matrix-docs:matrix.org` in future.

layouts/partials/openapi/render-object-table.html
richvdh: Thanks. Addressed in f2ebb400.

layouts/partials/openapi/render-object-table.html
richvdh: Thanks. Addressed in f2ebb400.

anoadragon453 (approved):

anoadragon453:
Sorry for leaving this to sit for so long. These changes LGTM.

PR #1799, opened 2024-04-24
Fix anchors for schemas under `oneOf`

zecakeh:
The subschemas were not updated in the parent schema, so the anchor would not be forwarded to the parent schema and the link to the subschema won't be rendered.

This results in 2 changes in the rendered spec:

For `m.room.encrypted`:

[Before](https://spec.matrix.org/v1.10/client-server-api/#mroomencrypted):

image

[After](https://pr1799--matrix-spec-previews.netlify.app/client-server-api/#mroomencrypted):

image

For `POST /_matrix/federation/v1/user/keys/claim`:

[Before](https://spec.matrix.org/v1.10/server-server-api/#post_matrixfederationv1userkeysclaim):

image

[After](https://pr1799--matrix-spec-previews.netlify.app/server-server-api/#post_matrixfederationv1userkeysclaim):

image

Preview: https://pr1799--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1799.clarification
+++ b/changelogs/internal/newsfragments/1799.clarification
@@ -1,0 +1,1 @@
+Fix anchors for schemas under `oneOf`.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
Thank you!

PR #1800, opened 2024-04-24
Use `OneTimeKeys` schema

zecakeh:
This was commented prior to the port to OpenAPI 3.1 for technical reasons (#1127). Now we can use it just fine.

I am wondering if the `oneOf: [string, KeyObject]` is right, since the section about [one-time and fallback keys](https://spec.matrix.org/v1.10/client-server-api/#one-time-and-fallback-keys) clearly says that only `signed_curve25519` is used in this case.

This adds the `KeyObject` definition in [several](https://pr1800--matrix-spec-previews.netlify.app/client-server-api/#_matrixclientv3keysclaim_keyobject) [places](https://pr1800--matrix-spec-previews.netlify.app/client-server-api/#_matrixclientv3keysupload_keyobject):

image

Preview: https://pr1800--matrix-spec-previews.netlify.app


richvdh:
This isn't really rendering right in the preview - it's referring to a `OneTimeKeys` type which is then not defined. Is that something that is fixed by one of the other PRs in flight?


image


> I am wondering if the oneOf: [string, KeyObject] is right, since the section about [one-time and fallback keys](https://spec.matrix.org/v1.10/client-server-api/#one-time-and-fallback-keys) clearly says that only signed_curve25519 is used in this case.

Yeah. I think the problem is that, previously, `curve25519` was used here, where the type is just `string`. I think it's fair to say that nobody should be using unsigned `curve25519` keys here, but maybe that is a separate change?

PR #1801, opened 2024-04-24
Do not use `title` for objects containing only `additionalProperties` or `patternProperties`

zecakeh:
Previously, titles would appear that do not link to a subschema definition. It would also mean that named subschemas would appear without being clearly referenced.

Now, the type clearly shows the nesting of objects and subschema definitions should be clearly referenced.

Examples:

[Before](https://spec.matrix.org/v1.10/client-server-api/#_matrixclientv3sync_rooms):

image

After:

Capture d’écran du 2024-04-24 12-31-37

[Before](https://spec.matrix.org/v1.10/client-server-api/#_matrixclientv3search_result-room-events):

image

After:

image

[Before]():

image

After:

image

Alternatively, we could just remove the `title` from the objects manually, allowing us to show a title when we really want to.

The rendered types can be improved later by using formats as introduced in #1796. 

Preview: https://pr1801--matrix-spec-previews.netlify.app


layouts/partials/openapi/render-object-table.html
```diff
@@ -155,12 +155,10 @@
       adhering to a regex pattern
 
     * `anchor`: optional HTML element id for the target type, which will be used to link to it.
-
-    The title has a higher priority than anything else.
 */}}
 {{ define "partials/object-type-or-title" }}
     {{ $type := "object" }}
-    {{ if .title }}
+    {{ if and .title (or .properties (not (or .additionalProperties .patternProperties))) }}
```
richvdh, line 161: could you add a comment explaining what this logic does in words, and why?

richvdh (commented):
Yes, I'd noticed this too. Suppressing the title when we aren't going to make a separate table sounds right to me.

layouts/partials/openapi/render-object-table.html
zecakeh: Oops, yes. I wanted to an forgot. Docs added.

layouts/partials/openapi/render-object-table.html
```diff
@@ -155,15 +155,17 @@
...
```
richvdh, line 161: we should mention `.properties` in the comment for this partial.

layouts/partials/openapi/render-object-table.html
```diff
@@ -155,15 +155,17 @@
...
-            If the property has a `title`, use that rather than `type`.
-            This means we can write things like `EventFilter` rather than `object`.
+          We only want to use the title in two cases:
+
+          * The object is rendered as a separate table that will use the same
+            title, which means that the object must have `properties`.
+          * The object doesn't define any properties, because showing the title
+            (like `EventFilter`) is better than showing `object`.
         */}}
         {{ $type = .title | htmlEscape }}
```
richvdh, line 170: I must say this logic makes my head explode, though I think it's probably right. 

What if we did something like this instead:

```
{{ $title = or .title "object" }}
{{ if .anchor }}
    {{ $title = printf "<a href=\"#%s\">%s</a>" (htmlEscape .anchor) $type }}
{{ end }}

{{ if .properties }} 
    {{/*
        The object has its own (regular) properties, so we will make a separate table for it. 
        Refer to it using its title, if it has one.
    */}}
    {{ $type = $title }}
{{ else if reflect.IsMap .additionalProperties }}
    // as before
{{ else if reflect.IsMap .patternProperties }}
    // as before
{{ else }}
    {{/*
        No properties, so there won't be a separate table. We use the title anyway, because showing the title
        (like `EventFilter`) is better than showing `object`.
    */}}
    {{ $type = $title }}
{{ end }}
```

zecakeh: Yes, that's probably easier to read. I'll change it.

zecakeh: Changed

layouts/partials/openapi/render-object-table.html
zecakeh: Done

richvdh (approved):
Thanks!

PR #1803, opened 2024-04-25
Update most CI actions

zecakeh:
This should get rid of most of the warnings in CI.

Note that [Beakyn/gha-comment-pull-request](https://github.com/Beakyn/gha-comment-pull-request) is unmaintained and should probably be replaced because it still declares using node 12.

Preview: https://pr1803--matrix-spec-previews.netlify.app


richvdh (approved):
Thanks

PR #1805, opened 2024-04-25
Set python version for the Towncrier CI job

zecakeh:
Otherwise the version might change depending on the runner. We just use the same version as other jobs.

This removes a GitHub warning.

Preview: https://pr1805--matrix-spec-previews.netlify.app


richvdh (approved):
ok

PR #1806, opened 2024-04-25
Replace `set-output` with environment files in CI

zecakeh:
It has been deprecated for a while: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/.


richvdh (approved):
thanks!

PR #1808, opened 2024-04-26
Deprecate authentication via a query string

zecakeh:
As per MSC4126.

It seems that OpenAPI 3.1 doesn't allow to mark a security scheme as deprecated, which is why I put it in the description.

Preview: https://pr1808--matrix-spec-previews.netlify.app


richvdh (approved):
Thanks!

changelogs/client_server/newsfragments/1808.clarification
Johennes, line 1: @zecakeh should this be a `.deprecation` as per the list in https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog?

turt2live: (we'll fix it during the changelog cleanup at release time - if there's other notable issues, please mention them in [#sct-office:matrix.org](https://matrix.to/#/#sct-office:matrix.org))

Johennes: Ok, yeah sorry for commenting on a closed PR. I wasn't sure what the proper process was. Would it help at all if I sent a PR renaming the file?

turt2live: Unfortunately not, as that PR then requires a changelog itself. Our [release process](https://github.com/matrix-org/matrix-spec/blob/main/meta/releasing.md) has a "normalize the changelog" step which takes care of this for us.

PR #1813, opened 2024-05-03
Use `patternProperties` in more places with supported formats

zecakeh:
Allows to have more places where the property name's type is better defined.

[For example](https://pr1813--matrix-spec-previews.netlify.app/client-server-api/#_matrixclientv3sync_rooms):

image

Preview: https://pr1813--matrix-spec-previews.netlify.app


richvdh (approved):
Thank you!

PR #1814, opened 2024-05-03
Add support for rendering string formats

zecakeh:
Similar to #1796, but for property values instead of property names. Currently only non-custom formats are defined for strings in OpenAPI definitions so I decided to add support for the `uri` format to have something to see.

[For example](https://pr1814--matrix-spec-previews.netlify.app/client-server-api/#well-knownmatrixclient_homeserver-information):

image

Preview: https://pr1814--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1814.clarification
+++ b/changelogs/internal/newsfragments/1814.clarification
@@ -1,0 +1,1 @@
+Add support for rendering string formats.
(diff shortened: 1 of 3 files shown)


uhoreg (approved):
Looks like it does what it says it does.  Thanks

PR #1816, opened 2024-05-07
Spec markup for mathematical messages

zecakeh:
As per MSC2191.

Preview: https://pr1816--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1816.clarification
+++ b/changelogs/client_server/newsfragments/1816.clarification
@@ -1,0 +1,1 @@
+Add support for mathematical messages, as per [MSC2191](https://github.com/matrix-org/matrix-spec-proposals/pull/2191).
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -429,6 +429,15 @@ representation instead. Clients SHOULD, however, aim to support, at minimum, the
 basic LaTeX2e maths commands and the TeX maths commands, with the possible
 exception of commands that could be security risks.
 
+{{% boxes/warning %}}
+Certain commands, such as [those that can create macros](https://katex.org/docs/supported#macros),
+are potentially dangerous. Clients should either decline to process those
+commands, or should take care to ensure that they are handled in safe ways (such
+as by limiting recursion). In general, LaTeX commands should be filtered by
+allowing known-good commands rather than forbidding known-bad commands. Some
+LaTeX libraries may have options for doing this.
+{{% /boxes/warning %}}
+
 #### Server behaviour
 
 Homeservers SHOULD reject `m.room.message` events which don't have a
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -430,12 +430,17 @@ basic LaTeX2e maths commands and the TeX maths commands, with the possible
 exception of commands that could be security risks.
 
 {{% boxes/warning %}}
-Certain commands, such as [those that can create macros](https://katex.org/docs/supported#macros),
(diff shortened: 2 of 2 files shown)


content/client-server-api/modules/instant_messaging.md
```diff
@@ -385,6 +386,49 @@ An example of a media message with a caption is:
...
+        sin(<i>x</i>)=<sup><i>a</i></sup>/<sub><i>b</i></sub>
+      </span>"
+}
+```
+
+The LaTeX format is poorly defined and has several extensions, so if a client
+encounters syntax that it cannot render, it SHOULD present the fallback
+representation instead. Clients SHOULD, however, aim to support, at minimum, the
+basic LaTeX2e maths commands and the TeX maths commands, with the possible
+exception of commands that could be security risks.
```
uhoreg, line 430: I think we need to call out the security risk more here, maybe by putting [this paragraph](https://github.com/matrix-org/matrix-spec-proposals/pull/2191/files#diff-9f3ff21f46a094d6bbf0f87639fa549aa118651ce79ac01bbae8668bfa5c4f70R119-R125) from the MSC in a WARNING box.

zecakeh: I agree, I added the warning box.

sumnerevans: I think it might also be worth calling out that calling out to a full-fledged LaTeX compiler such as `xelatex` is not recommended. Instead, using libraries which support only a subset of the spec should be expected.

zecakeh: I made the warning more comprehensive to include that.

content/client-server-api/modules/instant_messaging.md
```diff
@@ -385,6 +386,63 @@ An example of a media message with a caption is:
...
+      <span data-mx-maths=\"\\sin(x)=\\frac{a}{b}\">
+        sin(<i>x</i>)=<sup><i>a</i></sup>/<sub><i>b</i></sub>
+      </span>"
+}
+```
+
+The LaTeX format is poorly defined and has several extensions, so if a client
+encounters syntax that it cannot render, it SHOULD present the fallback
+representation instead. Clients SHOULD, however, aim to support, at minimum, the
+basic LaTeX2e maths commands and the TeX maths commands, with the possible
```
clokep, line 429: Should LaTeX2e and TeX link off to other specifications?

zecakeh: I am not sure where we could link for TeX, the [TUG](https://www.tug.org) website? There doesn't seem to be a specification.

uhoreg: https://www.latex-project.org/ seems to be the official site for LaTeX, so you can link to that for LaTeX2e.  Probably the closest to thing to an official spec for TeX is Knuth's The TeXbook, but that's only available as a physical book, or as [the `.tex` sources](https://ctan.org/pkg/texbook), which probably isn't the easiest to read. I think that either TUG or the `.tex` source for The TeXbook would be the best thing to link to.

zecakeh: Links added

uhoreg (approved):
Thanks!

changelogs/client_server/newsfragments/1816.clarification
Johennes, line 1: @zecakeh shouldn't this rather be a `.feature`?

turt2live: (we'll fix it during the changelog cleanup at release time - if there's other notable issues, please mention them in [#sct-office:matrix.org](https://matrix.to/#/#sct-office:matrix.org))

PR #1823, opened 2024-05-16
MSC4132: deprecate linking to events in rooms identified by alias

ara4n:
Implements https://github.com/matrix-org/matrix-spec-proposals/pull/4132

Preview: https://pr1823--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/appendices/newsfragments/1823.deprecation
+++ b/changelogs/appendices/newsfragments/1823.deprecation
@@ -1,0 +1,1 @@
+Deprecate linking to events in rooms identified by alias.
--- a/changelogs/appendices/newsfragments/1823.deprecation
+++ b/changelogs/appendices/newsfragments/1823.deprecation
@@ -1,1 +1,1 @@
-Deprecate linking to events in rooms identified by alias.
+Deprecate linking to events in rooms identified by alias, as per [MSC4132](https://github.com/matrix-org/matrix-spec-proposals/pull/4132).
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -745,7 +745,7 @@ Specifically, the following mappings are used:
 * `r` for room aliases.
 * `u` for users.
 * `roomid` for room IDs (note the distinction from room aliases).
-* `e` for events, when after a room reference (`roomid`).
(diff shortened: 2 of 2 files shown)


changelogs/appendices/newsfragments/1823.deprecation
```diff
@@ -0,0 +1 @@
+Deprecate linking to events in rooms identified by alias.
```
richvdh, line 1: For consistency with other changelog entries:

```suggestion
Deprecate linking to events in rooms identified by alias, as per [MSC4132](https://github.com/matrix-org/matrix-spec-proposals/pull/4132).
```

content/appendices.md
```diff
@@ -745,7 +745,7 @@ Specifically, the following mappings are used:
 * `r` for room aliases.
 * `u` for users.
 * `roomid` for room IDs (note the distinction from room aliases).
-* `e` for events, when after a room reference (`r` or `roomid`).
+* `e` for events, when after a room reference (`roomid`).
```
richvdh, line 748: The MSC specifically says, of event URIs using room aliases "Clients should ... continue to accept them"... so it seems a bit premature to rip them out here. (Yes, there's an INFO box, but nobody's going to read that.)

Perhaps we could move the INFO boxes right to the bottom of their respective sections, and just say:

```suggestion
* `e` for events, when after a room ID (`roomid`). Use of `e` after a room alias (`r`) is deprecated.
```

content/appendices.md
```diff
@@ -755,6 +755,13 @@ wish to consider handling them as `u`, `r`, and `e` respectively.
 `roomid` was otherwise unchanged.
 {{% /boxes/note %}}
 
+{{% boxes/note %}}
+{{% changed-in v="1.11" %}}
+Referencing event IDs within a room identified by room alias (`r`) rather than room ID
```
richvdh, line 760: Consider changing this to:

```
{{< changed-in v="1.11" >}} Referencing event IDs within a room identified by room alias (`r`) rather than room ID
```

(note the <> rather than %), which I think will give this whole thing a more compact representation.

richvdh (approved):
Thanks.

PR #1824, opened 2024-05-18
Clarify that redaction events are subject to auth rules

Kladki:
[Context](https://matrix.to/#/%23matrix-spec%3Amatrix.org/%24Y5gvmJK2fv-HwBRURrf6cHMtFvmqrTvfZ9Lur4fqDhE?via=ahouansou.cz&via=l1qu1d.net&via=0x1a8510f2.space&via=the-apothecary.club): there was a misunderstanding about how redactions interact with auth rules.

Signed-off-by: Matthias Ahouansou <matthias@ahouansou.cz>

Preview: https://pr1824--matrix-spec-previews.netlify.app

### DIFF
--- a/content/rooms/fragments/v8-auth-rules.md
+++ b/content/rooms/fragments/v8-auth-rules.md
@@ -24,7 +24,7 @@ level_ is **not** considered by the auth rules.
 
 The ability to send a redaction event does not mean that the redaction itself should
 be performed. Receiving servers must perform additional checks, as described in
-the [Handling Redactions](#handling-redactions) section below.
+the [Handling Redactions](#handling-redactions) section.
 {{% /boxes/note %}}
 
 The rules are as follows:
--- a/content/rooms/fragments/v8-auth-rules.md
+++ b/content/rooms/fragments/v8-auth-rules.md
@@ -1,11 +1,9 @@
 
 Events must be signed by the server denoted by the `sender` property.
 
-`m.room.redaction` events are not explicitly part of the auth rules.
-They are still subject to the minimum power level rules, but should always
-fall into "10. Otherwise, allow". Instead of being authorized at the time
-of receipt, they are authorized at a later stage: see the
-[Redactions](#redactions) section below for more information.
+While they are still subject to them, there are no auth rules specifically for
+`m.room.redaction` events. The actual redaction of content is authorized at a
+later stage: see the [Redactions](#redactions) section below for more information.
 
 The types of state events that affect authorization are:
 
--- a/content/rooms/fragments/v8-auth-rules.md
+++ b/content/rooms/fragments/v8-auth-rules.md
@@ -19,8 +19,8 @@ the default power level for users in the room.
 `m.room.redaction` events are subject to auth rules in the same way as any other event.
 In practice, that means they will normally be allowed by the auth rules, unless the
 `m.room.power_levels` event sets a power level requirement for `m.room.redaction`
-events via the `events` or `events_default` properties. In particular, the /redact
-level/ is **not** considered by the auth rules.
+events via the `events` or `events_default` properties. In particular, the _redact
+level_ is **not** considered by the auth rules.
 
 The ability to send a redaction event does not mean that the redaction itself should
 be performed. Receiving servers must perform additional checks, as described in
--- a/content/rooms/fragments/v8-auth-rules.md
+++ b/content/rooms/fragments/v8-auth-rules.md
@@ -1,10 +1,6 @@
 
 Events must be signed by the server denoted by the `sender` property.
 
-While they are still subject to them, there are no auth rules specifically for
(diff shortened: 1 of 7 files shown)


richvdh:
... though having read the diff I think I can guess what the misunderstanding was.

Kladki:
The misunderstanding was that 'are not explicitly part of the auth rules' and 'should always fall into "10. Otherwise, allow".' was interpreted as meaning that redaction events are not subject to the auth rules at all.

content/rooms/v6.md
```diff
@@ -40,11 +40,9 @@ in [room version 5](/rooms/v5).
...
 ### Authorization rules
 
-`m.room.redaction` events are not explicitly part of the auth rules.
-They are still subject to the minimum power level rules, but should always
-fall into "10. Otherwise, allow". Instead of being authorized at the time
-of receipt, they are authorized at a later stage: see the
-[Handling Redactions](#handling-redactions) section below for more information.
+While they are still subject to them, there are no auth rules specifically for
+`m.room.redaction` events. The actual redaction of content is authorized at a
+later stage: see the [Redactions](#redactions) section below for more information.
```
richvdh, line 45: I find this use of "they" and "them" to mean different things rather confusing. I think we can do better.

Also, you've changed the link, to refer to the wrong section.

```suggestion
`m.room.redaction` events are subject to auth rules in the same way as any other event.
In practice, that means they will normally be allowed by the auth rules, unless the
`m.room.power_levels` event sets a power level requirement for `m.room.redaction`
events via the `events` or `events_default` properties. In particular, the /redact
level/ is **not** considered by the auth rules.

The ability to send a redaction event does not mean that the redaction itself should
be performed. Receiving servers must perform additional checks, as described in
the [Handling Redactions](#handling-redactions) section below.
```

Suggest putting all this in a `{{% boxes/note %}}` box next to the other one, below.

richvdh: Likewise for the other room versions.

richvdh (changes requested):
Thanks for this. I think it helps, but I think it could be clearer.

Also, please add a sign-off to the PR, per https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off

content/rooms/v6.md
richvdh: yeah, I meant it to be italic. There are other references to "redact level" in the appendices, and they use italics.

Underscores, maybe?

Kladki:
Just to ensure you are aware, I have signed off the pull request now.

content/rooms/fragments/v8-auth-rules.md
```diff
@@ -21,6 +15,18 @@ For example, mentions of the `sender`'s power level can also refer to
...
+{{% boxes/note %}}
+`m.room.redaction` events are subject to auth rules in the same way as any other event.
+In practice, that means they will normally be allowed by the auth rules, unless the
+`m.room.power_levels` event sets a power level requirement for `m.room.redaction`
+events via the `events` or `events_default` properties. In particular, the _redact
+level_ is **not** considered by the auth rules.
+
+The ability to send a redaction event does not mean that the redaction itself should
+be performed. Receiving servers must perform additional checks, as described in
+the [Handling Redactions](#handling-redactions) section below.
```
richvdh, line 27: It's only "below" in room v8. It's "above" in room v9. Probably the easiest solution is just to remove "below":

```suggestion
the [Handling Redactions](#handling-redactions) section.
```

content/rooms/v7.md
```diff
@@ -57,6 +51,18 @@ For example, mentions of the `sender`'s power level can also refer to
...
+{{% boxes/note %}}
+`m.room.redaction` events are subject to auth rules in the same way as any other event.
+In practice, that means they will normally be allowed by the auth rules, unless the
+`m.room.power_levels` event sets a power level requirement for `m.room.redaction`
+events via the `events` or `events_default` properties. In particular, the _redact
+level_ is **not** considered by the auth rules.
+
+The ability to send a redaction event does not mean that the redaction itself should
+be performed. Receiving servers must perform additional checks, as described in
+the [Redactions](#redactions) section below.
```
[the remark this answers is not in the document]
richvdh: likewise in v10 and v11

richvdh (commented):
Thanks, a couple more things it would be nice to fix while you're here.

content/rooms/v10.md
```diff
@@ -99,7 +99,7 @@ level_ is **not** considered by the auth rules.
 
 The ability to send a redaction event does not mean that the redaction itself should
 be performed. Receiving servers must perform additional checks, as described in
-the [Redactions](#redactions) section below.
+the [Handling redactions](#handling-redactions) section.
```
richvdh, line 102: It *is* below in v10, but ... :man_shrugging:

richvdh (approved):
Thanks very much!

PR #1827, opened 2024-05-24
Fix typo of object being spelt as "obiect"

Kladki:
Preview: https://pr1827--matrix-spec-previews.netlify.app


turt2live (approved):
Thanks!

PR #1830, opened 2024-05-28
e2ee/device verification: clarifications

sumnerevans:
- **e2ee/device verification: move general error handling under framework**

  Most of the "Error and exception handling" section is generally applicable to other verification methods besides SAS, so I moved those bullet points under "Key verification framework".
- **e2ee/device verification/error handling: fix typo start -> request**

  It used to be possible to create a start request without a request, but this was deprecated.
- **m.key.verification.start: remove wording that it is typically to-device**

  It can be either a to-device event or an in-room-event.
- **e2ee/device verification: normalize all links**

   Changes the format of all links in the Device Verification section to be:

   ```
   [`m.key.verification.whatever`](#mkeyverificationwhatever)
   ```

   This involved adding links in a bunch of places, but also changing many links that did not have code within them to have the backticks.
-  **e2ee/device verification start: clarify required nature of transaction_id or m.relates_to**

   Add notes about the fact that the start event *may* be the first event sent during a verification process and that clients should handle other clients doing so, but not themselves send the start event first.

Signed-off-by: Sumner Evans <sumner.evans@automattic.com>

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

Preview: https://pr1830--matrix-spec-previews.netlify.app

PR #1831, opened 2024-05-28
Clean up pull request template

richvdh:
As far as I can tell, these header lines only encourage people to create badly-formatted PRs.

Also we only have one template so let's give it the default name.

Preview: https://pr1831--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,3 @@
----
-name: Spec clarification/not a proposal
-about: A change that's not a spec proposal, such as a clarification to the spec itself.
-title: ''
-labels: ''
-assignees: ''
-
----
 
 ### Pull Request Checklist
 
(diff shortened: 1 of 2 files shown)


erikjohnston:
Do you have an example of what this looks like? When I try and make a PR currently it seems to just give an empty box :/

richvdh:
The thing I'm trying to fix is stuff like https://github.com/matrix-org/matrix-spec/pull/1830.

erikjohnston (approved):
I mean sure why not.

Though I don't fully understand why I don't get PR templates :(

tulir:
> When I try and make a PR currently it seems to just give an empty box :/

I think that's because github's web UI doesn't support having multiple (named) PR templates, so when there's no default, it just doesn't show a template at all.

GitHub CLI does support those named templates, but it doesn't support the metadata section at the beginning

sumnerevans:
I wasn't really sure what to put in that section, but it just keeps showing up whenever I do `gh pr create` which is the main way that I create PRs. It didn't really make sense to just copy the title and description that I already had written, so I just left it.

PR #1833, opened 2024-05-31
Clarify that room ACLs apply to EDUs

Kladki:
### Pull Request Checklist

Preview: https://pr1833--matrix-spec-previews.netlify.app

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -1212,7 +1212,7 @@ of `M_FORBIDDEN`.
 
 The following endpoint prefixes MUST be protected:
 
--   `/_matrix/federation/v1/send` (on a per-PDU basis)
+-   `/_matrix/federation/v1/send` (on a per-PDU and EDU basis)
 -   `/_matrix/federation/v1/make_join`
 -   `/_matrix/federation/v1/make_leave`
 -   `/_matrix/federation/v1/send_join`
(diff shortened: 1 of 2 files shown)


content/server-server-api.md
```diff
@@ -1212,7 +1212,7 @@ of `M_FORBIDDEN`.
 
 The following endpoint prefixes MUST be protected:
 
--   `/_matrix/federation/v1/send` (on a per-PDU basis)
+-   `/_matrix/federation/v1/send` (on a per-PDU and EDU basis)
```
clokep, line 1215: I don't see this in the original [MSC1383](https://docs.google.com/document/d/1aiuROf1__7ZFkJvDdAZQfBNxyzjYd-ijiRAcHJYqJCM/edit#heading=h.t1ebd56v2ae6) nor any evidence that Synapse has historically done this.

It probably makes sense, but might need an MSC.

I'd be interested in the opinion of those around when ACLs were added, however.

tulir: Since the MSC doesn't say EDUs are forbidden and Synapse doesn't reject incoming EDUs, I don't think it can be added to the spec without a MSC

richvdh: Typing and read-receipts certainly existed before server ACLs, and yes, this would probably make sense, but unfortunately it's not what the MSC says, and not what the Synapse implementation does (indeed, there is an open issue about it: https://github.com/element-hq/synapse/issues/11456).

This seems like a great idea but we'd need an MSC before we can add it to the spec. Please do open one! But I'm going to have to close this for now.

Issue #1782, opened 2024-04-10
Use RFC 2119 keywords consistently


turt2live:
https://tools.ietf.org/html/rfc2119

See https://github.com/matrix-org/matrix-spec/issues/701
See https://github.com/matrix-org/matrix-spec/issues/682

clokep:
I'm in favor of standardizing this. It offloads any terminology discussion to the IETF and matches what people generally expect when reading specifications.

turt2live:
ftr, as this affects the "presentation layer" of the spec, I don't believe this requires an MSC. It's something we can do without actually changing the meaning of the specification.

Johennes:
Shall MSCs attempt to use the RFC2119 language as well or should that be reserved to the spec itself?

richvdh:
You've all said the same thing as me, but for posterity here's the GH comment I started writing 6 hours ago:

---

I think we've said that use of RFC2119 language is to be encouraged in MSCs, especially as an MSC gets closer to FCP.

In the past we've tended to be more relaxed about wording in MSCs "as long as the intention is clear" but that has had some unfortunate consequences.

A well-written MSC often has a bunch of informative sections around a normative proposal. Obviously, only the normative section needs formal RFC2119 keywords.

In other words: MSCs SHOULD express the proposal using RFC2119 keywords.

Johennes:
For reference: In https://github.com/matrix-org/matrix-spec/pull/1858#discussion_r1634728239 we discussed that the RFC keywords probably shouldn't be used in info or warning boxes because these usually should be descriptive rather than normative.

For what it's worth, this also matches the rules currently being used in gematik's extension of the spec. Below is an example of a normative requirement (delimited by `A_...` and `<=`) that uses keyword capitalization and an adjacent explanatory note ("Hinweis") that doesn't.

![Screenshot 2024-06-25 at 08 18 01](https://github.com/matrix-org/matrix-spec/assets/1137962/31206c32-8e1c-4480-818e-75512739f759)

HarHarLinks:
> ftr, as this affects the "presentation layer" of the spec, I don't believe this requires an MSC. It's something we can do without actually changing the meaning of the specification.

While I agree with the comments made, I can not find that the spec defines its use of RFC 2119 language at all so far? If this is correct, I think that we would also need to do that, and I wonder if that change could affect the meaning of the spec somewhere.

turt2live:
[...]

PR #1839, opened 2024-06-03
Clarifications around encoding of data within verification QR codes

hughns:
### Pull Request Checklist

Currently the [spec](https://spec.matrix.org/v1.10/client-server-api/#qr-code-format) is specific about the contents of the binary data to be *present* within a QR, but it does not specify how that data should be *encoded* within a QR.

I've checked the original [MSC1544](https://github.com/matrix-org/matrix-spec-proposals/pull/1544/files#diff-fd371cd950df67f5a748e49d367ce76d81c216f62feaf870a714cf43413265b4R171) but don't see any obvious discussion around this. Also, no discussion in the spec [PR](https://github.com/matrix-org/matrix-spec-proposals/pull/3149).

As such, I assume this is an oversight.

I do have a hesitation about linking out to the ISO spec as it is not "open".

Preview: https://pr1839--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1200,7 +1200,7 @@ The binary segment MUST be of the following form:
   request event, encoded as:
   - two bytes in network byte order (big-endian) indicating the length in
     bytes of the ID as a UTF-8 string
-  - the ID encoded as a UTF-8 string (i.e. one UTF-8 byte per character)
+  - the ID encoded as a UTF-8 string
 - the first key, as 32 bytes.  The key to use depends on the mode field:
   - if `0x00` or `0x01`, then the current user's own master cross-signing public key
   - if `0x02`, then the current device's Ed25519 signing key
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1206,11 +1206,11 @@ The binary segment MUST be of the following form:
   - if `0x02`, then the current device's Ed25519 signing key
 - the second key, as 32 bytes.  The key to use depends on the mode field:
   - if `0x00`, then what the device thinks the other user's master
-    cross-signing key is
+    cross-signing public key is
   - if `0x01`, then what the device thinks the other device's Ed25519 signing
+    public key is
+  - if `0x02`, then what the device thinks the user's master cross-signing public
     key is
-  - if `0x02`, then what the device thinks the user's master cross-signing key
-    is
 - a random shared secret, as a sequence of bytes.  It is suggested to use a secret
   that is about 8 bytes long.  Note: as we do not share the length of the
   secret, and it is not a fixed size, clients will just use the remainder of
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1179,10 +1179,16 @@ The process between Alice and Bob verifying each other would be:
 
 ###### QR code format
 
-The QR codes to be displayed and scanned using this format will encode binary
-strings in the general form:
+The QR codes to be displayed and scanned must be
+[ISO/IEC 18004:2015](https://www.iso.org/standard/62021.html) compatible and
+contain a single segment that uses the byte mode encoding.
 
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/end_to_end_encryption.md
sumnerevans, line 1: Do we also need to clarify what it means to encode the `transaction_id` as a UTF-8 string (line 1203)?

hughns: Yes, I've also pushed a change to try and remove other ambiguities arounds "strings".

sumnerevans (commented):
While you're in this area, you could clarify that the keys in key1 and key2 are public keys. For example:

> - the second key, as 32 bytes. The key to use depends on the mode field
>   - if `0x00`, then what the device thinks the other user's master cross-signing public key is
>   - if `0x01`, then what the device thinks the other device's Ed25519 signing public key is
>   - if `0x02`, then what the device thinks the user's master cross-signing public key is

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1179,10 +1179,16 @@ The process between Alice and Bob verifying each other would be:
 
 ###### QR code format
 
-The QR codes to be displayed and scanned using this format will encode binary
-strings in the general form:
+The QR codes to be displayed and scanned must be
```
Johennes, line 1182:
```suggestion
The QR codes to be displayed and scanned MUST be
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1179,10 +1179,16 @@ The process between Alice and Bob verifying each other would be:
...
+[ISO/IEC 18004:2015](https://www.iso.org/standard/62021.html) compatible and
+contain a single segment that uses the byte mode encoding.
 
-- the ASCII string `MATRIX`
+The error correction level can be chosen by the device displaying the QR code.
+
+The binary segment must be of the following form:
```
Johennes, line 1188:
```suggestion
The binary segment MUST be of the following form:
```

Johennes (commented):
Two minor suggestions with respect to https://github.com/matrix-org/matrix-spec/issues/1782.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1179,10 +1179,16 @@ The process between Alice and Bob verifying each other would be:
 
 ###### QR code format
 
-The QR codes to be displayed and scanned using this format will encode binary
-strings in the general form:
+The QR codes to be displayed and scanned must be
+[ISO/IEC 18004:2015](https://www.iso.org/standard/62021.html) compatible and
```
richvdh, line 1183:
```suggestion
compatible with [ISO/IEC 18004:2015](https://www.iso.org/standard/62021.html) and
```

richvdh, line 1183: [sadly yes, the spec for QR codes is copyrighted; this seems to be the best we can do.]

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1194,7 +1200,7 @@ strings in the general form:
   request event, encoded as:
   - two bytes in network byte order (big-endian) indicating the length in
     bytes of the ID as a UTF-8 string
-  - the ID as a UTF-8 string
+  - the ID encoded as a UTF-8 string (i.e. one UTF-8 byte per character)
```
richvdh, line 1203: I don't know what "one UTF-8 byte per character" means. Isn't that a contradiction?

Assuming there are non-ascii characters in the ID, then those characters will require more than one byte to encode.

richvdh (changes requested):

content/client-server-api/modules/end_to_end_encryption.md
hughns: I don't recall what I was thinking... this looks better to me:

```suggestion
  - the ID encoded as a UTF-8 string
```

richvdh (approved):
LGTM. Thanks!

sumnerevans (approved):
Thanks for getting these clarifications across the line!

PR #1840, opened 2024-06-05
Clarify when an event is returned from /send_join.

### LINKED ISSUE #1708: Clarify that join events only need to be signed by resident servers if using `join_authorised_via_users_server`
**Link to problem area**: https://spec.matrix.org/v1.9/server-server-api/#put_matrixfederationv2send_joinroomideventid

The join event only needs a signature from a resident server if joining via the restricted join rules flow, i.e. `join_authorised_via_users_server` is specified. 

c.f. [MSC3083](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/3083-restricted-rooms.md) and [Room Version 8 auth rules](https://spec.matrix.org/v1.9/rooms/v8/)

clokep:
Fixes #1708.

Signed-off-by: Patrick Cloke <clokep@patrick.cloke.us>

Preview: https://pr1840--matrix-spec-previews.netlify.app


data/api/server-server/joins-v2.yaml
```diff
@@ -207,9 +207,8 @@ paths:
                     title: SignedMembershipEvent
                     x-addedInMatrixVersion: "1.2"
                     description: |-
-                      Required if the `content` of the event in the request contained the `join_authorised_via_users_server`
-                        field. The signed copy of the membership event sent to other servers by the resident server,
-                        including the resident server's signature.
+                      The signed copy of the membership event sent to other servers by the resident server.
+                      Required if the event is signed by the resident server.
```
richvdh, line 211: Hrm. I think we need to clarify this a bit more. It just begs the question, "so when should the event be signed by the resident server".

richvdh (changes requested):

data/api/server-server/joins-v2.yaml
clokep: Would linking back to restricted joins be enough?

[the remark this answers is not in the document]
clokep: What do you mean by present? Present when you get the room state?

As the joining server I think the important things to do are:

1. Verify the signatures as with any event.
2. Ensure this copy of the event is persisted locally and not the copy that was used to make the `/send_join` call.

I'd have to double check what synapse does to see if there's anything else though.

richvdh: Probably?

clokep: I took another attempt at this, let me know what you think.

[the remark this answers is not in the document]
clokep: The signature needs to be checked (as is in the auth rules), Synapse does this @ https://github.com/matrix-org/synapse/blob/23740eaa3da94fbe2e25180376961979fc4e8cd3/synapse/federation/federation_base.py#L238-L259.

I'm still unsure if that's your question or not though.

richvdh (approved):
LGTM

PR #1841, opened 2024-06-05
Fix link to push rule condition kinds.

clokep:
As reported by https://matrix.to/#/!NasysSDfxKxZBzJJoE:matrix.org/$6qi3oaPnzhcBphCN57WC3PDsP9Y5juch2QJGp2ceMgQ?via=matrix.org&via=envs.net&via=element.io.

Signed-off-by: Patrick Cloke <clokep@patrick.cloke.us>


clokep:
I'm not sure why it thinks the link is broken? It also doesn't seem to have generated a preview. 😢

zecakeh:
The IDs of the page are deduplicated in the browser by manipulating the DOM with JavaScript. If the test doesn't do the same thing, it will probably not know the real IDs for the links.

zecakeh:
It looks like the easiest solution here is to use the `IgnoreURLs` [option](https://github.com/wjdp/htmltest?tab=readme-ov-file#basic-options) of htmltest.

zecakeh:
#1851 should fix the issue with the CI.

richvdh (approved):
Thanks. Could you merge in `develop` to check that it fixes CI?

PR #1842, opened 2024-06-05
Remove references to device-specific push rules.

### LINKED ISSUE #637: properly remove per-device push rules
the APIs are still hanging around, which confuses people. we should add them back to the spec deliberately in future

clokep:
Remove references to device-specific push rules, which have never been implemented and aren't really specified.

This might need an MSC, but #637 is marked as a clarification so I wasn't sure. 🤷 

Fix #637

### DIFF
--- a/data/api/client-server/pushrules.yaml
+++ b/data/api/client-server/pushrules.yaml
@@ -16,228 +16,214 @@ info:
   title: Matrix Client-Server Push Rules API
   version: 1.0.0
 paths:
-  /pushrules/:
+  /pushrules/global/:
     get:
-      summary: Retrieve all push rulesets.
+      summary: Retrieve all push rules.
       description: |-
-        Retrieve all push rulesets for this user. Clients can "drill-down" on
-        the rulesets by suffixing a `scope` to this path e.g.
-        `/pushrules/global/`. This will return a subset of this data under the
-        specified key e.g. the `global` key.
+        Retrieve all push rules for this user.
       operationId: getPushRules
       security:
         - accessTokenQuery: []
         - accessTokenBearer: []
       responses:
         "200":
-          description: All the push rulesets for this user.
+          description: All the push rules for this user.
           content:
             application/json:
               schema:
-                type: object
-                required:
-                  - global
-                properties:
-                  global:
-                    type: object
-                    description: The global ruleset.
-                    title: Ruleset
-                    allOf:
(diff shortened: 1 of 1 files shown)


clokep:
This still has a dangling reference to "Push Rulesets" (see https://pr1842--matrix-spec-previews.netlify.app/client-server-api/#push-notifications), unsure how to rephrase that.

data/api/client-server/pushrules.yaml
```diff
@@ -16,228 +16,214 @@ info:
   title: Matrix Client-Server Push Rules API
   version: 1.0.0
 paths:
-  /pushrules/:
+  /pushrules/global/:
```
richvdh, line 19: This removes the `GET /_matrix/client/v3/pushrules/` API, which would need an MSC - clients rely on it existing. Unfortunately I think we need to keep both `GET /_matrix/client/v3/pushrules/` and `GET /_matrix/client/v3/pushrules/global` for now.

clokep: You're correct. I thought I tested this and the endpoint wasn't actually supported, I think I forgot the trailing slash. :sad:

clokep: I think it is useful to split this out to a separate endpoint still, although copying most of the example over is unfortunate.

data/api/client-server/pushrules.yaml
```diff
@@ -237,7 +235,214 @@ paths:
...
+        "200":
+          description: All the push rules for this user.
+          content:
+            application/json:
+              schema:
+                $ref: definitions/push_ruleset.yaml
+              examples:
+                response:
+                  value: {
+                    "content": [
```
richvdh, line 257: I believe, if we move this example to `definitions/push_ruleset.yaml`, it will magically be used for both endpoints, without the need to duplicate it.

richvdh (changes requested):
@clokep: would you be able to update this to remove the duplication?

clokep:
> @clokep: would you be able to update this to remove the duplication?

Yes, sorry about that! 👍

clokep:
Compare https://pr1842--matrix-spec-previews.netlify.app/client-server-api/#get_matrixclientv3pushrules to https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3pushrules

richvdh (approved):
LGTM!

PR #1843, opened 2024-06-05
Clarify that an access token is optional on `/account/password` and `/account/deactivate`

### LINKED ISSUE #1807: `POST /_matrix/client/v3/account/password` has authentication set as required but description says it is optional
**Link to problem area**: https://spec.matrix.org/v1.10/client-server-api/#post_matrixclientv3accountpassword

**Issue**

The table of the endpoint shows:

> Requires authentication: Yes

However the description has these 2 sentences:

> An access token should be submitted to this endpoint if the client has an active session.
>
> The homeserver may change the flows available depending on whether a valid access token is provided.

Which means that the access token is actually optional.

**Expected behaviour**

Although that was not a thing before, https://github.com/matrix-org/matrix-spec/pull/1728 has added the possibility to display:

> Requires authentication: Optional

A simple change in the OpenAPI definition should be enough to fix that.

I have not done a thorough search to know if other endpoints are affected.

zecakeh:
The descriptions of the endpoints clearly state that the access token is not required, but it also says:

> Requires Authentication: Yes

This changes it to show:

> Requires Authentication: Optional

Fixes #1807.

### Pull Request Checklist

Preview: https://pr1843--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1843.clarification
+++ b/changelogs/client_server/newsfragments/1843.clarification
@@ -1,0 +1,1 @@
+Clarify that authentication is optional on the `POST /account/password` and `POST /account/deactivate` endpoints.
--- a/changelogs/client_server/newsfragments/1843.clarification
+++ b/changelogs/client_server/newsfragments/1843.clarification
@@ -1,1 +1,1 @@
-Clarify that authentication is optional on the `POST /account/password` and `POST /account/deactivate` endpoints.
+Clarify that an access token is optional on the `POST /account/password` and `POST /account/deactivate` endpoints.
--- a/data/api/client-server/registration.yaml
+++ b/data/api/client-server/registration.yaml
@@ -387,6 +387,7 @@ paths:
         access token provided in the request. Whether other access tokens for
         the user are revoked depends on the request parameters.
       security:
+        - {}
         - accessTokenQuery: []
         - accessTokenBearer: []
       operationId: changePassword
@@ -592,6 +593,7 @@ paths:
         parameter because the homeserver is expected to sign the request to the
         identity server instead.
       security:
+        - {}
         - accessTokenQuery: []
         - accessTokenBearer: []
       operationId: deactivateAccount


uhoreg:
Looking at the synapse code, `/account/password` indeed doesn't seem to require authentication, but `/account/deactivate` does, AFAICT.  I haven't checked what other servers do yet.  I think we may need to untangle whether the discrepancy in `/account/deactivate` is a spec bug or a synapse bug.

zecakeh:
From what I can tell both Dendrite and Conduit seem to require authentication for both endpoints.

richvdh:
It took me quite a while to grok what was going on here. Clearly, both endpoints require authentication of some form -- anything else would be anarchy where people go around changing other peoples' passwords - so the phrasing of the PR is a bit confusing.

So, for the record: both endpoints require authentication via UIA. The question at stake is whether they *also* need an `Authorization` header with an access token. The spec is inconsistent. As @uhoreg says, Synapse's implementation of `/account/deactivate` requires an access token, but that of `/account/password` does not.

richvdh:
Doing some archaeology:

`/password` was implemented in https://github.com/matrix-org/synapse/pull/126/files#diff-20445d423fa1657115c589560de173027dbba8b9a866b21fb3592cd55e30dc50R31
 and specced in https://github.com/matrix-org/matrix-spec-proposals/pull/190. We can see from the start that the access token was not required if you did UIA with `m.login.email.identity`, which remains the case today; however the swagger implied that an access token was required. @zecakeh's change seems correct here and corrects a 9-year-old spec bug!

`/deactivate` was implemented in https://github.com/matrix-org/synapse/pull/921 and specced in https://github.com/matrix-org/matrix-doc/pull/361. The spec implies that an access token may not be required, but in practice, an access token *was* required (https://github.com/matrix-org/synapse/pull/921/files#diff-20445d423fa1657115c589560de173027dbba8b9a866b21fb3592cd55e30dc50R150). 

My impression is that a server *could* implement `/deactivate` without the access token, for example by allowing UIA with `m.login.token`, and be spec-compliant. The fact that no implementation has yet done so isn't important.

In short: LGTM

changelogs/client_server/newsfragments/1843.clarification
```diff
@@ -0,0 +1 @@
+Clarify that authentication is optional on the `POST /account/password` and `POST /account/deactivate` endpoints.
```
richvdh, line 1:
```suggestion
Clarify that an access token is optional on the `POST /account/password` and `POST /account/deactivate` endpoints.
```

maybe?

richvdh (approved):

zecakeh:
> It took me quite a while to grok what was going on here. Clearly, both endpoints require authentication of some form -- anything else would be anarchy where people go around changing other peoples' passwords -- so the phrasing of the PR is a bit confusing.

In that case, should we consider that the phrasing of the spec is also confusing? Do we need to change "Requires Authentication" to "Requires Access Token"?

changelogs/client_server/newsfragments/1843.clarification
zecakeh: LGTM

richvdh:
> In that case, should we consider that the phrasing of the spec is also confusing? Do we need to change "Requires Authentication" to "Requires Access Token"?

I think something like that would be good, yes. Not sure exactly what words to use though.

richvdh (approved):

PR #1847, opened 2024-06-07
Spec `unsigned.membership` property, per MSC4115

richvdh:
cf MSC4115

Preview: https://pr1847--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/client_event_without_room_id.yaml
+++ b/data/api/client-server/definitions/client_event_without_room_id.yaml
@@ -118,10 +118,11 @@ properties:
           The room membership of the user making the request, at the time of the event.
 
           This property is the value of the `membership` property of the
-          requesting user's [`m.room.member`](#mroommember) state at the point
-          of the event, including any changes caused by the event. If the user
-          had yet to join the room at the time of the event (i.e, they have no
-          `m.room.member` state), this property is set to `leave`.
+          requesting user's [`m.room.member`](/client-server-api#mroommember)
+          state at the point of the event, including any changes caused by the
+          event. If the user had yet to join the room at the time of the event
+          (i.e, they have no `m.room.member` state), this property is set to
+          `leave`.
 
           It is recommended that homeservers SHOULD populate this property
           wherever practical, but they MAY omit it if necessary (for example,
--- a/data/api/client-server/definitions/client_event_without_room_id.yaml
+++ b/data/api/client-server/definitions/client_event_without_room_id.yaml
@@ -124,7 +124,7 @@ properties:
           (i.e, they have no `m.room.member` state), this property is set to
           `leave`.
 
-          It is recommended that homeservers SHOULD populate this property
+          Homeservers SHOULD populate this property
           wherever practical, but they MAY omit it if necessary (for example,
           if calculating the value is expensive, servers might choose to only
           implement it in encrypted rooms). The property is *not* normally populated
--- a/data/api/client-server/definitions/client_event_without_room_id.yaml
+++ b/data/api/client-server/definitions/client_event_without_room_id.yaml
@@ -128,7 +128,8 @@ properties:
           wherever practical, but they MAY omit it if necessary (for example,
           if calculating the value is expensive, servers might choose to only
           implement it in encrypted rooms). The property is *not* normally populated
-          in the application service API (where there is no clear definition of "requesting user").
+          in events pushed to application services via the application service transaction API
+          (where there is no clear definition of "requesting user").
         type: string
         example: join
         x-addedInMatrixVersion: "1.11"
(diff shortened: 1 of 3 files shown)


richvdh:
ugh this has made it appear in the AS API, which is not the intention

richvdh:
oh, maybe it's fine. I'll add a couple more words.

data/api/client-server/definitions/client_event_without_room_id.yaml
```diff
@@ -112,3 +113,22 @@ properties:
...
+          state at the point of the event, including any changes caused by the
+          event. If the user had yet to join the room at the time of the event
+          (i.e, they have no `m.room.member` state), this property is set to
+          `leave`.
+
+          It is recommended that homeservers SHOULD populate this property
+          wherever practical, but they MAY omit it if necessary (for example,
+          if calculating the value is expensive, servers might choose to only
+          implement it in encrypted rooms). The property is *not* normally populated
+          in the application service API (where there is no clear definition of "requesting user").
```
clokep, line 131: I couldn't find this bit about application services in the MSC -- I would think you'd populate it with whoever the AS is acting as?

clokep: I was thinking the "virtual user" provided by the `user_id` query parameter: https://spec.matrix.org/v1.10/application-service-api/#identity-assertion

But now we're kind of making stuff up that's not in the MSC. So probably best to leave it and if someone else wants it for AS they can define that in a new MSC.

richvdh:
```suggestion
          implement it in encrypted rooms). The property is *not* normally populated
          in events pushed to application services via the application service transaction API
          (where there is no clear definition of "requesting user").
```

clokep: Oh yes, that makes sense!

clokep: That sounds good, can you add a cross link to the AS docs?

clokep (approved):

data/api/client-server/definitions/client_event_without_room_id.yaml
richvdh: yeees, but then it looks even more ridiculous in the AS docs itself

clokep: What do you mean that it looks ridiculous? I'm ok with it as is, if you think it looks best now.

richvdh: > What do you mean that it looks ridiculous? 

Currently, under https://pr1847--matrix-spec-previews.netlify.app/application-service-api/#put_matrixappv1transactionstxnid, we have:

image

That looks moderately ridiculous imho (why is it listed at all?), but it's hard to do much about. OTOH, if it links straight back to https://pr1847--matrix-spec-previews.netlify.app/application-service-api/#put_matrixappv1transactionstxnid, it might be even sillier.

Or maybe we should accept the awkwardness, and at least linking it would be clear?

data/api/client-server/definitions/client_event_without_room_id.yaml
```diff
@@ -112,3 +113,23 @@ properties:
...
+          The room membership of the user making the request, at the time of the event.
+
+          This property is the value of the `membership` property of the
+          requesting user's [`m.room.member`](/client-server-api#mroommember)
+          state at the point of the event, including any changes caused by the
+          event. If the user had yet to join the room at the time of the event
+          (i.e, they have no `m.room.member` state), this property is set to
+          `leave`.
+
+          It is recommended that homeservers SHOULD populate this property
```
Johennes, line 127: This feels slightly duplicative because SHOULD effectively already means RECOMMEND.

> [3](https://datatracker.ietf.org/doc/html/rfc2119#section-3). SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>   may exist valid reasons in particular circumstances to ignore a
>   particular item, but the full implications must be understood and
>   carefully weighed before choosing a different course.

Maybe it could be shortened similar to this?

```suggestion
          Homeservers SHOULD populate this property
```

data/api/client-server/definitions/client_event_without_room_id.yaml
clokep: Ah -- the same bit gets added in both endpoints. I understand now. 👍 So yes -- I agree that would be a silly link to include.

I think ideally it wouldn't even be listed on the AS spec, but I have no idea how much refactoring that would need.

data/api/client-server/definitions/client_event_without_room_id.yaml
richvdh: good spot, thanks!

PR #1849, opened 2024-06-07
Do not add empty arrays to examples

### LINKED ISSUE #1811: Empty objects and null in registration arrays
**Link to problem area**:
https://spec.matrix.org/v1.10/application-service-api/#registration
**Issue**
There are empty json objects and null where there is just supposed to be namespaces and strings respectively.
https://spec.matrix.org/v1.10/application-service-api/

image

Similar to https://github.com/matrix-org/matrix-spec/issues/1010

zecakeh:
Effectively, this only changes the rendering of the example of https://spec.matrix.org/v1.10/application-service-api/#registration

[Result](https://pr1849--matrix-spec-previews.netlify.app/application-service-api/#registration)

### Pull Request Checklist

Preview: https://pr1849--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1849.clarification
+++ b/changelogs/internal/newsfragments/1849.clarification
@@ -1,0 +1,1 @@
+Do not add empty arrays to examples.
--- a/layouts/partials/json-schema/resolve-example.html
+++ b/layouts/partials/json-schema/resolve-example.html
@@ -35,9 +35,9 @@
         */}}
         {{ if reflect.IsMap $this_object.items }}
             {{ $items_example := partial "json-schema/resolve-example" $this_object.items }}
-            {{ $example = slice $items_example }}
-        {{ else }}
-            {{ $example = slice }}
+            {{ if $items_example }}
+                {{ $example = slice $items_example }}
+            {{ end }}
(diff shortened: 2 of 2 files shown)


sumnerevans (commented):
image

this doesn't look like it was the intended result...

And are we sure that there aren't any places where an empty list is significant in the spec? I feel like I've seen some places where it says something like "an empty list represents explicitly nothing, whereas null/omitted represents use the defaults".

zecakeh:
It was the intended result for me. Especially getting rid of `[ null ]` which is almost always wrong. If we want to have a list it is always possible to provide it as an example in the schema, but without this change it is not possible to remove empty lists that have no examples.

The fact that this example is still wrong is another issue. A complete example should probably be provided since we cannot generate a valid one with the current schema.

IMO, this brings to light the issue that we don't check those generated examples in CI. To do that we would need to re-implement the example generation for CI. Or to stop relying on generated examples and always provide full examples, because they are easier to check. The thing though is that we rely heavily on them.

richvdh (approved):
Seems sensible. We should have a better example for "Registration" but that's a separate problem.

Issue #1826, opened 2024-05-23
Clarify Room ID requirements


HarHarLinks:
**Link to problem area**:
https://spec.matrix.org/latest/appendices/#room-ids

**Issue**
- According to https://matrix.to/#/%23urandom%3Amaunium.net/%24GC76GPFSf0n8VEHdZN2FxSgiRb2MGDcA6L0DRwAVR7s?via=sosnowkadub.de&via=exkc.moe&via=synapse.test.progval.net&via=ucmeet-server.business.celcom.com.my Room IDs need to be limited to 255 chars to be valid in state keys
- these 255 chars probably apply to URL encoded ASCII representation because room IDs get used in URIs? I'm not sure. it seems strange that character-set, encoding, etc appears to not be defined while user IDs at least speak of UTF-8 and explicitly state a 255 char limit.

richvdh:
Why are room IDs used in state keys?

deepbluev7:
> Why are room IDs used in state keys?

https://spec.matrix.org/v1.10/client-server-api/#mspacechild-relationship

HarHarLinks:
for some reason, size limits appear in this place https://spec.matrix.org/v1.10/client-server-api/#size-limits

this might answer my second question and is not referenced from the appendix section

richvdh:
Oh, that section seems pretty clear about the requirements here.

Seems like this should be defined in the appendix, and then referenced from the "size limits" section.

PRs to update this would be very welcome.

PR #1850, opened 2024-06-07
Move size limits for user, room and event IDs into the appendix and clarify that the length is to be measured in bytes

Johennes:
Fixes: #1826
Relates to: #1001

### Pull Request Checklist

Preview: https://pr1850--matrix-spec-previews.netlify.app

### DIFF
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -556,7 +556,7 @@ The `domain` of a user ID is the [server name](#server-name) of the
 homeserver which allocated the account.
 
 The length of a user ID, including the `@` sigil and the domain, MUST
-NOT exceed 255 characters.
+NOT exceed 255 bytes.
 
 The complete grammar for a legal user ID is:
 
@@ -663,6 +663,9 @@ Room IDs are case-sensitive. They are not meant to be
 human-readable. They are intended to be treated as fully opaque strings
 by clients.
 
+The length of a room ID, including the `!` sigil and the domain, MUST
+NOT exceed 255 bytes.
+
 #### Room Aliases
 
 A room may have zero or more aliases. A room alias has the format:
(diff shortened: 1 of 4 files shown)


content/appendices.md
```diff
@@ -556,7 +556,7 @@ The `domain` of a user ID is the [server name](#server-name) of the
 homeserver which allocated the account.
 
 The length of a user ID, including the `@` sigil and the domain, MUST
-NOT exceed 255 characters.
+NOT exceed 255 bytes.
```
richvdh, line 559: For future reference: this restriction comes from the fact that a user ID must fit inside a state key (in the `m.room.member` event), and state keys have always been limited to 255 bytes (since https://github.com/matrix-org/matrix-spec/commit/6a0595bc46c3751090d5d24e2ab94647c9573915)

content/appendices.md
```diff
@@ -663,6 +663,9 @@ Room IDs are case-sensitive. They are not meant to be
 human-readable. They are intended to be treated as fully opaque strings
 by clients.
 
+The length of a room ID, including the `!` sigil and the domain, MUST
+NOT exceed 255 bytes.
```
richvdh, line 667: This was also previously specified in the C-S API, also going back to https://github.com/matrix-org/matrix-spec/commit/6a0595bc46c3751090d5d24e2ab94647c9573915.

content/appendices.md
```diff
@@ -689,6 +692,8 @@ whereas more recent versions omit the domain and use a base64-encoded hash inste
 Event IDs are case-sensitive. They are not meant to be human-readable. They are
 intended to be treated as fully opaque strings by clients.
 
+The length of an event ID MUST NOT exceed 255 bytes.
+
```
richvdh, line 696: I feel like we should acknowledge that some room versions have much stricter requirements on event IDs which mean we can never get anywhere near 255 bytes:

```suggestion
In addition to the requirements of the room version, the length of an event ID,
including the `$` sigil and the domain where present, MUST NOT exceed 255 bytes.

Event IDs are case-sensitive. They are not meant to be human-readable. They are
intended to be treated as fully opaque strings by clients.
```

content/appendices.md
richvdh: no, that's fine. I just wanted to make a record so that when, in a few months, we come to look at this PR, we remember the thought process.

richvdh (approved):
Thanks!

PR #1851, opened 2024-06-08
Generate ToC with Hugo rather than JavaScript

zecakeh:
Thanks to the `.RenderShortcodes` function added in Hugo 0.117.0, we can include the CS modules and all the headings are included in the generated ToC, with unique IDs. This avoids to do it with JavaScript in the browser.

This should fix the issue in #1841.

This might however leave some duplicate IDs in the code generated from OpenAPI definitions.

### Pull Request Checklist

Preview: https://pr1851--matrix-spec-previews.netlify.app


content/client-server-api/_index.md
```diff
@@ -2805,42 +2805,42 @@ operations and run in a resource constrained environment. Like embedded
 applications, they are not intended to be fully-fledged communication
 systems.
 
-{{< cs-module name="instant_messaging" >}}
```
richvdh, line 2808: hrm. We switched to `{{<` deliberately (https://github.com/matrix-org/matrix-spec/pull/1205). Switching back seems like it will break things?

richvdh: Oh, this is ok because the `cs-module` shortcode is switching to `.RenderShortcodes` instead of `.Content`? But why is it necessary?

zecakeh: Hugo is lazy and uses the markdown renderer to detect headings and construct the headings tree.

`.Content` returns a page rendered as HTML, and `{{<` means "don't process the data returned by this shortcode with the renderer", so the renderer in the main page just replaces the shortcode with the module page rendered as HTML without processing anything, and Hugo is only aware of the headings in the main page.

`.RenderShortcodes` returns the raw markdown, with the shortcodes of the page already processed, and `{{%` means "process the data returned by this shortcode with the renderer", so the renderer processes the full mardown of the main page and the module pages as a single document, so Hugo is aware of all the headings in the final page.

.htmltest.yaml
```diff
@@ -4,3 +4,4 @@
 IgnoreDirectoryMissingTrailingSlash: true
 DirectoryPath: spec
 CheckExternal: false
+IgnoreInternalEmptyHash: true
```
richvdh, line 7: for the record: this stops `htmltest` complaining about links with `href="#"`

layouts/partials/toc.html
```diff
@@ -0,0 +1,15 @@
...
+
+  A modified version of the toc.html partial in Docsy.
+
+*/}}
+{{ $page := .Params }}
+{{ if not .Params.notoc -}}
+  {{ with .TableOfContents -}}
+    <hr>
+    <div id="toc">
+        <a id="toc-title" href="#">{{ $page.Title }}</a>
```
richvdh, line 11: is the `href="#"` here important? I know that the old ToC code generated one, but I think that might be more by accident than anything else.

richvdh:
Writing it here for want of somewhere better: I wondered if we could avoid cloning the whole of `sidebar-tree.html` by putting `toc.html` back in `baseof.html`. The answer is that we *can*, with some custom CSS, along the lines of:

baseof.html:
```patch
-            {{ partial "sidebar.html" . }}
+            <div class="sidebar-div td-sidebar-nav">
+              {{ partial "sidebar.html" . }}
+              {{ partial "toc.html" . }}
+            </div>
```

Custom CSS:
```css
.sidebar-div {
    overflow-y: auto;
    position: sticky;
    top: 4rem;
    height: calc(100vh - 5rem);
}

.td-sidebar__inner {
    height: auto !important;
    position: static !important;
}
```

(In other words, we have to move the `height` and `position` from `.td-sidebar__inner` to a higher-level div.)  But frankly, that looks more brittle than just forking `sidebar-tree.html`.

.htmltest.yaml
zecakeh: htmltest only tests the HTML source, the ToC was generated with JavaScript, so htmltest didn't check it.

layouts/partials/toc.html
zecakeh: It allows to go back to the very top of the page, I don't know if it should be considered an important feature :shrug:.

zecakeh: If we want to avoid changing the htmltest config, we could set an actual ID to go back to the top.

layouts/partials/toc.html
```diff
@@ -0,0 +1,15 @@
...
+        {{ . -}}
```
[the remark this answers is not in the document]
zecakeh: No, it tells the templating system to ignore whitespaces and newlines after it. It should only have an effect on the HTML source, not the rendering.

layouts/partials/sidebar-tree.html
```diff
@@ -0,0 +1,80 @@
+{{/*
+
+  A modified version of the siderbar-tree.html partial in Docsy to add the ToC.
```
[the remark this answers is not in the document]
zecakeh: The omission of the comment is not deliberate. I tried to document the change with the "to add the ToC" part. I thought it was clear enough since the partial is called `toc.html`.

layouts/partials/toc.html
```diff
@@ -0,0 +1,15 @@
...
+    </div>
+  {{ end -}}
+{{ end -}}
```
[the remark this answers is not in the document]
zecakeh: It would be easier if there wasn't a conspiracy to hide the trailing NLs in most places (GitHub being one of them)…

layouts/partials/sidebar-tree.html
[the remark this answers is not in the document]
zecakeh: I restored the comment and clarified the modification.

layouts/partials/toc.html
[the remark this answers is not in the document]
zecakeh: Fixed

layouts/partials/toc.html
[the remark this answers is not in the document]
zecakeh: Removed.

.htmltest.yaml
richvdh: oh of course

layouts/partials/toc.html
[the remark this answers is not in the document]
richvdh: Github seems to show them to me, at least in the diff view?

layouts/partials/toc.html
richvdh: oh, I hadn't realised that was what `href="#"` does. This is fine.

richvdh (approved):
Thanks!

PR #1852, opened 2024-06-10
Add missing word header

Johennes:
### Pull Request Checklist

Preview: https://pr1852--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1852.clarification
+++ b/changelogs/client_server/newsfragments/1852.clarification
@@ -1,0 +1,1 @@
+Fix various typos throughout the specification.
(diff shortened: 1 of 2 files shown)


Johennes:
On a related note, it seems like sending the `retry_after_ms` response field was downgraded from SHOULD to MAY in 1.10. I think for a proper deprecation it should have stayed SHOULD so that servers include it whenever they set the new header to support clients that haven't yet caught up with the spec. Given that this was already released, I suppose there's no point in changing it yet again though?

uhoreg (approved):
Good catch.  Thanks

Issue #1437, opened 2023-02-15
Missing modules in Feature Profiles


uhoreg:
**Link to problem area**: https://spec.matrix.org/unstable/client-server-api/#feature-profiles

**Issue**
The table should list all the modules, but it's missing:

- Secrets
- Third Party Invites (not to be confused with Third Party Networks, which is in the table)
- Room Tagging

Also the table is in a completely different order from the order the modules are actually in

PR #1855, opened 2024-06-10
Add missing modules to feature profiles

Johennes:
I also sorted the table by section occurrence and copied the section title in cases where it differed from the title in the table.

Fixes: #1437 

### Pull Request Checklist


turt2live (changes requested):
Thanks! I'd suggest we split out the adding and resorting to their own PRs. It's not immediately clear to me that this is the ordering we'd want (previously it was loosely ordered on "required-ness"). Adding the missing stuff is certainly important, though.

Johennes:
Ok, I can do that. Where would you suggest I insert the missing modules in the existing order?

turt2live:
We typically just append them to the end, for better or worse. If they're "required" (they shouldn't be), put them under the existing required ones.

Johennes:
I pushed #1860 which merely adds the three missing modules without touching the table ordering.

Johennes:
The brief discussion in the matrix-spec room was, sadly, inconclusive. The suggested options so far are:

1. Sort alphabetically
2. Sort by section occurrence
3. Sort by requiredness (this is more or less how the table is sorted currently)
4. Sort by requiredness first and alphabetically second

As a middleground, I updated the PR to match option 4 but also renamed those rows where the module name didn't match the section title. This should hopefully still make it easy enough to cross-reference the table with the sections.

For convenience, the PR can be reviewed by commit:

* https://github.com/matrix-org/matrix-spec/pull/1855/commits/aa93f8b7126cb44ea8fec9a44543c5a6440155ed does the renaming
* https://github.com/matrix-org/matrix-spec/pull/1855/commits/5afe6ed19f2559d1567f02c4be2d3eb38a852eab does the sorting

Johennes:
CC @uhoreg / @anoadragon453 since you participated in the chat in matrix-spec.

turt2live (commented):
I'm going to abstain from reviewing this one - the order doesn't bother me.

clokep (approved):

PR #1867, opened 2024-06-12
MSC4170: 403 error responses for profile APIs

Johennes:
Spec for https://github.com/matrix-org/matrix-spec-proposals/pull/4170.

### Pull Request Checklist

Preview: https://pr1867--matrix-spec-previews.netlify.app


data/api/client-server/profile.yaml
```diff
@@ -98,6 +98,20 @@ paths:
                   value: {
                     "displayname": "Alice Margatroid"
                   }
+        "403":
+          x-addedInMatrixVersion: "1.2"
```
Johennes, line 102: I don't know if this is correct but technically we're clarifying the 403 response introduced in Matrix v1.2.

richvdh (approved):
LGTM, but will await opinions from other SCT members on https://github.com/matrix-org/matrix-spec/pull/1867/files#r1636536644

data/api/server-server/query.yaml
```diff
@@ -170,6 +170,20 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: ../client-server/definitions/errors/error.yaml
+              examples:
+                response:
+                  value: {
+                    "errcode": "M_FORBIDDEN",
+                    "error": "Profile lookup over federation is disabled on this homeserver"
+                  }
```
richvdh, line 186: actually: given MSC3550 didn't mention the federation API at all, making this change without an MSC seems like a bit of a stretch.

richvdh (changes requested):

data/api/server-server/query.yaml
richvdh: (and while we're at it maybe we should discuss the additional client endpoints too, since they weren't *technically* covered by the MSC or the original spec PR)

clokep: > given MSC3550 didn't mention the federation API at all

It mentions that the reason the MSC is being written is for requests over federation, but then doesn't mention the federation endpoints at all.

I'm a bit torn on it.

Johennes: I am (not eager but definitely) willing to put up a short MSC for this. I would like to ensure that it's time well spent though. So I'll hold off of doing anything until you've found an agreement on whether an MSC is needed or not.

turt2live: We discussed it briefly at the time of richvdh's comment, and I think the thread here captures that discussion well. It feels like it needs an MSC because it's relatively unclear what the original MSC's intended scope was. 

As a formality, my vote is +1.0 to needing an MSC because the intention and history are different stories.

Johennes: Ok, makes sense. I'll try and put something together next week.

Johennes: This PR is now blocked on https://github.com/matrix-org/matrix-spec-proposals/pull/4170

Johennes: This PR has now been updated as per the accepted MSC.

Johennes:
I've optimistically used `x-addedInMatrixVersion: "1.12"` but this doesn't block the 1.12 release in any way. If the release happens before this PR lands, I'll bump the versions to 1.13.

richvdh (approved):
LGTM. THanks!

PR #1869, opened 2024-06-13
Replace references to obsolete RFC 1341 with RFC 2046

zecakeh:
Sorry to discover that only after #1858 was merged, but it turns out that if we look at [the whole RFC 1341 on the IETF's website](https://datatracker.ietf.org/doc/html/rfc1341), or even [the TOC of the RFC on the W3C's website](https://www.w3.org/Protocols/rfc1341/0_TableOfContents.html), we can see that it was obsoleted by [RFC 1521](https://datatracker.ietf.org/doc/html/rfc1521), which was in turn obsoleted by RFCs 2045-2049, and the part that interests us is now in [RFC 2046](https://datatracker.ietf.org/doc/html/rfc2046).

### Pull Request Checklist

Preview: https://pr1869--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/1869.feature
+++ b/changelogs/server_server/newsfragments/1869.feature
@@ -1,0 +1,1 @@
+Media downloads and thumbnails are now authenticated, as per [MSC3916](https://github.com/matrix-org/matrix-spec-proposals/pull/3916).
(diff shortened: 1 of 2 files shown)


turt2live (approved):
good spot - thanks!

Issue #1790, opened 2024-04-16
`a`nchor tag `name` attribute is deprectated


HarHarLinks:
**Link to problem area**:
https://spec.matrix.org/latest/client-server-api/#mroommessage-msgtypes

**Issue**
`a`nchor tag `name` attribute is deprectated

**Expected behaviour**
should be replaced by the global <del>`title`</del> `id` attribute or removed

HarHarLinks:
indeed, thanks, edited in OP

PR #1870, opened 2024-06-14
Remove the deprecated name attribute on HTML anchor elements

Johennes:
The `name` attribute has been allowed since the [r0.4.0](https://spec.matrix.org/legacy/client_server/r0.4.0.html) and was introduced with https://github.com/matrix-org/matrix-spec-proposals/pull/1562/files#diff-f90b924249c2799cc2d7eb25cba5418fa8b0412dc3b8a3f176964dab7b2d789bR88.

It has since been [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#name):

> Was required to define a possible target location in a page. In HTML 4.01, `id` and `name` could both be used on `<a>`, as long as they had identical values.
>
>    Note: Use the global attribute [`id`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#id) instead.

I think we should not allow clients to render HTML IDs from external content. So the only possible way forward appears to be removing the `name` attribute without a replacement.

CC @turt2live in case you recall the use case that led to `name` being permitted.

Fixes: #1790

### Pull Request Checklist

Preview: https://pr1870--matrix-spec-previews.netlify.app


changelogs/client_server/newsfragments/1870.deprecation
Johennes, line 1: I'm not sure if deprecation is the correct classification for this since we're removing it without deprecating it. However, none of the other categories from https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog seemed fitting.

content/client-server-api/modules/instant_messaging.md
```diff
@@ -74,7 +74,7 @@ the tag.
 | Tag    | Permitted Attributes                                                                                                                       |
 |--------|--------------------------------------------------------------------------------------------------------------------------------------------|
 | `span` | `data-mx-bg-color`, `data-mx-color`, `data-mx-spoiler` (see [spoiler messages](#spoiler-messages)), `data-mx-maths` (see [mathematical messages](#mathematical-messages)) |
-| `a`    | `name`, `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`) |
+| `a`    | `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`) |
```
turt2live, line 77: stylistic note: please ensure the table still lines up in ASCII when editing.
```suggestion
| `a`    | `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`)          |
```

turt2live (dismissed):
The `name` attribute is imported from Vector, which is before even my time with Matrix. I'm not sure why it was included, or what value it has being included. 

From a spec process perspective, I think we'd need to see the WHATWG spec on this deprecation before we can meaningfully decide if it needs an MSC. I suspect this PR will probably need an MSC if the feature is being removed with no replacement. From MSC4077 ([ref](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4077-process-deprecated-html.md#proposal)):

> It is left to the discretion of the spec PR reviewer as to whether an MSC is required. In general, an MSC should be requested when the replacement HTML feature is not quite the same as the old feature. The reviewer *may* also decide whether to keep supporting the old, deprecated, feature for some time. For example, replacing `strike` with `s` may mean that both tags are supported for 1-2 Matrix spec versions before `strike` can be dropped.

content/client-server-api/modules/instant_messaging.md
Johennes: Sorry, I missed that 🤦

Johennes:
WHATWG tracks it as obsolete: https://html.spec.whatwg.org/multipage/obsolete.html#obsolete-but-conforming-features

> Authors should not specify the name attribute on a elements. If the attribute is present, its value must not be the empty string and must neither be equal to the value of any of the IDs in the element's tree other than the element's own ID, if any, nor be equal to the value of any of the other name attributes on a elements in the element's tree. If this attribute is present and the element has an ID, then the attribute's value must be equal to the element's ID. In earlier versions of the language, this attribute was intended as a way to specify possible targets for fragments in URLs. The id attribute should be used instead.

Johennes:
I'm happy to propose the removal as an MSC first if that is required. I'm not sure what the name attribute could ever have been used for but if we're unsure an MSC might help get further perspectives on that.

content/client-server-api/modules/instant_messaging.md
```diff
@@ -74,7 +74,7 @@ the tag.
 | Tag    | Permitted Attributes                                                                                                                       |
 |--------|--------------------------------------------------------------------------------------------------------------------------------------------|
 | `span` | `data-mx-bg-color`, `data-mx-color`, `data-mx-spoiler` (see [spoiler messages](#spoiler-messages)), `data-mx-maths` (see [mathematical messages](#mathematical-messages)) |
-| `a`    | `name`, `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`) |
+| `a`    | `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`)          |
```
turt2live, line 77: sorry, I counted wrong

```suggestion
| `a`    | `target`, `href` (provided the value is not relative and has a scheme matching one of: `https`, `http`, `ftp`, `mailto`, `magnet`)         |
```

Johennes:
@turt2live would you like this to go through an MSC before continuing here?

turt2live:
I think so, yes. We can try to fast track the decision making on it though to avoid process burden.

Johennes:
Ok, no problem. I'll try to get a short proposal up.

Johennes:
This is now blocked on https://github.com/matrix-org/matrix-spec-proposals/pull/4159

Johennes:
Undrafting since the MSC has been merged.

changelogs/client_server/newsfragments/1870.deprecation
```diff
@@ -0,0 +1 @@
+Remove the deprecated name attribute on HTML anchor elements
```
uhoreg, line 1:
```suggestion
Remove the deprecated name attribute on HTML anchor elements as per [MSC4159](https://github.com/matrix-org/matrix-spec-proposals/pull/4159).
```

changelogs/client_server/newsfragments/1870.deprecation
uhoreg: There is a "Removed" section, which isn't listed in the `CONTRIBUTING.rst`

uhoreg (approved):

changelogs/client_server/newsfragments/1870.deprecation
Johennes: Oh, I see. I have opened https://github.com/matrix-org/matrix-spec/pull/1907 to add it to the contribution guide.

Thanks also for adding the MSC to the changelog. I had totally forgotten about that. 🤦‍♂️

PR #1882, opened 2024-06-20
Allow to specify a prefix for generated HTML IDs of API endpoints

zecakeh:
Based on #1881 because it touches the same code.

This allows to deduplicate IDs of duplicate endpoints, and with #1881 it gets rid of almost all duplicate IDs.

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

Preview: https://pr1882--matrix-spec-previews.netlify.app

PR #1881, opened 2024-06-20
Ensure more uniqueness for generated HTML IDs

zecakeh:
Given that the IDs are not "uniquified" with JavaScript anymore, we need to make sure the generated IDs are unique.

This solves most cases of duplicate IDs, except for the duplicate endpoints `POST /_matrix/client/v3/rooms/{roomId}/invite` and `GET /_matrix/client/v3/events` which will probably need a new solution. 

See each commit's message for details.

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

Preview: https://pr1881--matrix-spec-previews.netlify.app

PR #1883, opened 2024-06-20
Fix the rendering of the event format for room versions 1 and 2

zecakeh:
Before, `prev_events` and `auth_events` would show an empty array, and the `Event Hash` definition would be duplicated.

[Before on stable](https://spec.matrix.org/v1.10/rooms/v1/#event-format):

image

And on unstable:

image

[After](https://pr1883--matrix-spec-previews.netlify.app/rooms/v1/#event-format):

image

This also gets rid of the last duplicate IDs after #1881 and #1882.

### Pull Request Checklist

Preview: https://pr1883--matrix-spec-previews.netlify.app


turt2live (approved):
Thanks!

PR #1884, opened 2024-06-20
Fix the table of content for room versions

zecakeh:
We need to use `.RenderShortcodes` for room version pages like in #1851, otherwise the headings in the included fragments are not present in the ToC of the page.

However, that breaks the `withVersioning` trick, because the output we get has shortcodes replaced by temporary opaque strings, so we cannot detect them to remove them. The solution I found is to use a page parameter in the `added-in` and `changed-in` shortcodes to detect if the current version is the same as the `v` parameter, since it seems to be the only way to pass a variable from the rendered page to the shortcode in the fragment. 

### Pull Request Checklist

Preview: https://pr1884--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/room_versions/newsfragments/1884.clarification
+++ b/changelogs/room_versions/newsfragments/1884.clarification
@@ -1,0 +1,1 @@
+Fix the tables of contents for the room versions.
--- a/changelogs/room_versions/newsfragments/1884.clarification
+++ b/changelogs/room_versions/newsfragments/1884.clarification
@@ -1,1 +1,1 @@
-Fix the tables of contents for the room versions.
+Generate the Table of Contents with Hugo rather than JavaScript.
--- a/layouts/shortcodes/added-in.html
+++ b/layouts/shortcodes/added-in.html
@@ -1,8 +1,9 @@
-{{ $ver := .Params.v }}
-{{ $this := .Params.this }}
+{{- $ver := .Params.v -}}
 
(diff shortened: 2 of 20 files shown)


turt2live:
https://spec.matrix.org/unstable/rooms/v11/ (and others) appear to be working fine right now. Is there a place where we can see the bug?

turt2live:
oh my. That's dangerous.

Ideally we'd have CI for this, but that's not a problem for this PR to solve.

changelogs/room_versions/newsfragments/1884.clarification
```diff
@@ -0,0 +1 @@
+Fix the tables of contents for the room versions.
```
turt2live, line 1:
```suggestion
Generate the Table of Contents with Hugo rather than JavaScript.
```

layouts/shortcodes/added-in.html
```diff
@@ -1,8 +1,9 @@
-{{ $ver := .Params.v }}
-{{ $this := .Params.this }}
+{{- $ver := .Params.v -}}
 
-{{ if $this }}
-  <span><strong>[New in this version]</strong></span>
-{{ else }}
+{{- with page.Params.version -}}
```
turt2live, line 3: this block doesn't appear to be working:

image

turt2live (changes requested):
otherwise lgtm

PR #1890, opened 2024-06-25
Clarify that dont_notify and coalesce MUST be gracefully ignored

Johennes:
Relates to: https://github.com/matrix-org/matrix-spec/pull/1501/files#r1176822971.

I replaced the info box with a subsection akin to the one for historical user IDs to prevent using RFC2119 keywords inside an explanatory box.

I also attempted to be a little more verbose about what ignoring means by describing the filtering behavior [implemented in Synapse](https://github.com/element-hq/synapse/blob/f79dbd0f61194929585d7010a3ec1b9ee208f033/rust/src/push/evaluator.rs#L221).

CC @clokep 

### Pull Request Checklist

Preview: https://pr1890--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1890.clarification
+++ b/changelogs/client_server/newsfragments/1890.clarification
@@ -1,0 +1,1 @@
+Clarify that the deprecated dont_notify and coalesce push rule actions MUST be ignored, not rejected by homeservers.
--- a/changelogs/client_server/newsfragments/1890.clarification
+++ b/changelogs/client_server/newsfragments/1890.clarification
@@ -1,1 +1,1 @@
-Clarify that the deprecated dont_notify and coalesce push rule actions MUST be ignored, not rejected by homeservers.
+Clarify that the deprecated dont_notify and coalesce push rule actions MUST be ignored, not rejected.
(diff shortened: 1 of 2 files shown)


clokep:
I did check the original MSC and that uses (lowercase) "should" and "recommended", but I really think it is a requirement for backwards compatibility.

content/client-server-api/modules/push.md
```diff
@@ -184,11 +184,13 @@ they are represented as a dictionary with a key equal to their name and
...
+###### Historical Actions
+
 Older versions of the Matrix specification included the `dont_notify` and
-`coalesce` actions. These should both be considered no-ops (ignored, not
-rejected) if received from a client.
-{{% /boxes/note %}}
+`coalesce` actions. Homeservers MUST ignore these actions by stripping them
+from any actions array they encounter. This means that setting the actions
+on a rule to e.g. `["dont_notify"]` MUST be equivalent to setting an empty
+actions array.
```
Johennes, line 193: Not that I'm keen on making this more complicated but I'm now wondering if this clause should apply to clients as well since they might receive default push rules with `dont_notify` when connecting to an older server? 🤔

clokep: Yes it probably also applies to clients working with an old server.

Johennes: Ok, I have rephrased the paragraph slightly to include clients as well.

changelogs/client_server/newsfragments/1890.clarification
```diff
@@ -0,0 +1 @@
+Clarify that the deprecated dont_notify and coalesce push rule actions MUST be ignored, not rejected.
```
clokep, line 1:
```suggestion
Clarify that the deprecated `dont_notify` and `coalesce` push rule actions MUST be ignored, not rejected.
```

clokep (approved):

PR #1903, opened 2024-07-06
remove confusing description of restricted rooms with no valid conditions

### LINKED ISSUE #1891: Behaviour of restricted rooms with no valid conditions is not clear
## Link to problem area:
https://spec.matrix.org/v1.10/client-server-api/#restricted-rooms
## Issue:
Currently, the spec states that:
> If the room is `restricted` but no valid conditions are presented then the room is effectively invite only.

The problem here is that there seems to be [multiple interpretations of this](https://github.com/ruma/ruma/pull/1871#discussion_r1657351112), those being:
- Taking this literally, using the exact rules of `invite` join rule when this is the case
- Taking this as a figure of speech, as for the most the room should act as if it was `invite`, since servers shouldn't ever authorize joins

The key difference between these two cases is that with the latter, a `join` can still be authorized via `join_authorised_via_users_server`, despite the fact that servers shouldn't be doing this. With the former, such a join wouldn't be authorized.

So which interpretation does the spec intend to abide by?

Kladki:
Fixes  #1891

I tried changing the wording, but I couldn't think of something that isn't either verbose or confusing. The auth rules already cover the proper behavior anyways

### Pull Request Checklist

Preview: https://pr1903--matrix-spec-previews.netlify.app


changelogs/client_server/newsfragments/1903.changelog
Johennes, line 1: Should this be a `.clarification`? The `changelog` type doesn't appear to exist in https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog.

PR #1909, opened 2024-07-18
Clarify that the `User identifier` object in `POST /_matrix/client/v3/login` contains additional properties that depend on the identification type

Johennes:
The `UserIdentifier` object in https://spec.matrix.org/v1.11/client-server-api/#post_matrixclientv3login uses `additionalProperties: true` which, as far as I understand it, is equivalent to `additionalProperties: {}`. The rendering template only checks if `additionalProperties` is a map and, thus, omits rendering the `<Other properties>` row in this case. This PR maps `true` to `{}` before performing the `isMap` check.

### Pull Request Checklist

Preview: https://pr1909--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1909.clarification
+++ b/changelogs/client_server/newsfragments/1909.clarification
@@ -1,1 +1,1 @@
-The `UserIdentifier` object in `POST /_matrix/client/v3/login` can contain additional properties.
+The `UserIdentifier` object in `POST /_matrix/client/v3/login` contains additional properties that depend on the identification type.
--- a/changelogs/client_server/newsfragments/1909.clarification
+++ b/changelogs/client_server/newsfragments/1909.clarification
@@ -1,0 +1,1 @@
+The `UserIdentifier` object in `POST /_matrix/client/v3/login` can contain additional properties.
--- a/changelogs/client_server/newsfragments/1909.clarification
+++ b/changelogs/client_server/newsfragments/1909.clarification
@@ -1,1 +1,1 @@
-The `UserIdentifier` object in `POST /_matrix/client/v3/login` contains additional properties that depend on the identification type.
+The `User identifier` object in `POST /_matrix/client/v3/login` contains additional properties that depend on the identification type.
--- a/data/api/client-server/definitions/user_identifier.yaml
+++ b/data/api/client-server/definitions/user_identifier.yaml
@@ -21,4 +21,4 @@ properties:
     description: The type of identification.  See [Identifier types](/client-server-api/#identifier-types) for supported values and additional property descriptions.
 required:
   - type
-additionalProperties: true
+additionalProperties: {}
--- a/data/api/client-server/definitions/user_identifier.yaml
+++ b/data/api/client-server/definitions/user_identifier.yaml
@@ -18,7 +18,10 @@ type: object
 properties:
   type:
     type: string
-    description: The type of identification.  See [Identifier types](/client-server-api/#identifier-types) for supported values and additional property descriptions.
+    description: |-
+      The type of identification. See [Identifier types](/client-server-api/#identifier-types)
+      for supported values and additional property descriptions.
 required:
   - type
-additionalProperties: {}
+additionalProperties:
+  description: Keys dependent on the identification type
--- a/data/api/client-server/definitions/user_identifier.yaml
+++ b/data/api/client-server/definitions/user_identifier.yaml
@@ -24,4 +24,4 @@ properties:
 required:
   - type
 additionalProperties:
-  description: Keys dependent on the identification type
+  description: Keys dependent on the identification type.
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -67,12 +67,18 @@
 
       TODO: support `patternProperties` here.
     */}}
-    {{ if reflect.IsMap .additionalProperties }}
 
+    {{/* Map `additionalProperties: true` to the equivalent `additionalProperties: {}` */}}
+    {{ $additionalProperties := .additionalProperties }}
+    {{ if eq $additionalProperties true }}
+      {{ $additionalProperties = dict }}
+    {{ end }}
+
+    {{ if reflect.IsMap $additionalProperties }}
  <tr>
   <td>&lt;Other properties&gt;</td>
-  <td><code>{{ partial "partials/property-type" .additionalProperties | safeHTML }}</code></td>
-  <td>{{ partial "partials/property-description" (dict "property" .additionalProperties) }}</td>
+  <td><code>{{ partial "partials/property-type" $additionalProperties | safeHTML }}</code></td>
+  <td>{{ partial "partials/property-description" (dict "property" $additionalProperties) }}</td>
  </tr>
     {{ end }}
 </table>
--- a/layouts/partials/openapi/render-object-table.html
+++ b/layouts/partials/openapi/render-object-table.html
@@ -53,11 +53,6 @@
 
     {{ end }}
 
-    {{/* Enforce `additionalProperties: {}` instead of the equivalent `additionalProperties: true` for easier processing */}}
-    {{ if eq .additionalProperties true }}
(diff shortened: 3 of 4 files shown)


changelogs/client_server/newsfragments/1909.clarification
```diff
@@ -0,0 +1 @@
+The `UserIdentifier` object in `POST /_matrix/client/v3/login` can contain additional properties.
```
zecakeh, line 1: I believe that a better solution would be to change the definition of `UserIdentifier` rather than encourage the use of `true`.

It would probably be good to always have a `description` for `additionalProperties`.

In this particular case, we could also define that the values of the additional properties are always strings.

layouts/partials/openapi/render-object-table.html
```diff
@@ -67,12 +67,18 @@
 
       TODO: support `patternProperties` here.
     */}}
-    {{ if reflect.IsMap .additionalProperties }}
 
+    {{/* Map `additionalProperties: true` to the equivalent `additionalProperties: {}` */}}
```
zecakeh, line 71: If this change is accepted, the comments above need to be adjusted.

changelogs/client_server/newsfragments/1909.clarification
Johennes: I thought about that but without a technical way to prohibit the use of `additionalProperties: true` I was worried that it might easily creep in again elsewhere in the spec in the future?

anoadragon453: We could add a bright red error banner in the HTML if `additionalProperties` is set to something other than the desired type? That would at least alert content writers to their mistake.

Hugo doesn't look like it has any `assert` functionality to fail the build otherwise.

zecakeh: Hugo has one, it's [`errorf`](https://gohugo.io/functions/fmt/errorf/).

anoadragon453: @Johennes can you modify this PR to invoke `errorf` in this case?

Johennes: Apologies for the delay I was away for a few days. The `errorf` function sounds brilliant. I didn't know this existed. I have changed things around to enforce `additionalProperties: {}` instead.

layouts/partials/openapi/render-object-table.html
Johennes: This is now obsolete due to https://github.com/matrix-org/matrix-spec/pull/1909/files#r1687890480.

layouts/partials/openapi/render-object-table.html
```diff
@@ -53,6 +53,11 @@
 
     {{ end }}
 
+    {{/* Enforce `additionalProperties: {}` instead of the equivalent `additionalProperties: true` for easier processing */}}
+    {{ if eq .additionalProperties true }}
+      {{ errorf "Use `additionalProperties: {}` instead of `additionalProperties: true`" }}
+    {{ end }}
```
anoadragon453, line 59: To cover more use cases, we can check whether the type of `additionalProperties` is a map with [`reflect.isMap`](https://gohugo.io/functions/reflect/ismap/).

Johennes: You mean throw the error if `additionalProperties` is not a map, regardless of what it is?

richvdh (changes requested):
I kinda feel like we're missing the point here. As a general rule, *everything* in Matrix can have additional properties (as indeed the [comment](https://github.com/matrix-org/matrix-spec/blob/20c109c3a710ed68ef101b4a791a22b4c407b4bd/layouts/partials/openapi/render-object-table.html#L67-L68) says), so I'm not really sure what `additionalProperties: true`, or, for that matter, `additionalProperties: {}`, is trying to tell us.

IMHO the only thing worth doing here is to remove the redundant `additionalProperties: true` from `user_identifier.yaml`.

At the very least, the comment needs an update.

Johennes:
Hm, I think there are two different cases though.

The comment says "you're allowed to include random unspecced properties in your object". This one I can understand and it would mean that #1910 is indeed redundant. I guess this wasn't quite clear to me because it's already used contrarily in other places such as https://spec.matrix.org/v1.11/client-server-api/#getwell-knownmatrixclient which I suppose are equally redundant.

However, the `User identifier` object strikes me as a different case. It must contain other _specced_ properties dependent on the value of its `type` property. Wouldn't that be worth spelling out?

richvdh:
> other places such as https://spec.matrix.org/v1.11/client-server-api/#getwell-knownmatrixclient which I suppose are equally redundant.

Well, in that case, we are at least using the `additionalProperties` to say that the additional properties should use the "Java package naming convention". If all we're doing is adding `additonalProperties: {}`, then there is absolutely no information being transferred.

> However, the `User identifier` object strikes me as a different case. It must contain other _specced_ properties dependent on the value of its `type` property. Wouldn't that be worth spelling out?

Well, yeees... I'm just not convinced that `additionalProperties: {}` is a good way to do that. That's not a common convention in openapi specs, and if I just read the openapi doc without reading the spec, and saw `additionalProperties: {}`, I wouldn't know that the author was trying to say "there most be other specced properties dependent on the `type` property". Of course we could *define* it as a convention for the matrix openapi docs, and update the hugo templates to say something relevant, but it doesn't feel very intuitive to me.

What if we extended the `user_identifier` definition to include a `description` on the `additionalProperties`, saying that?

Johennes:
> Well, in that case, we are at least using the additionalProperties to say that the additional properties should use the "Java package naming convention".

Ok, let's maybe spin this one off into https://github.com/matrix-org/matrix-spec/pull/1910 because I think I fully agree with the rest of what you said.

> Well, yeees... I'm just not convinced that additionalProperties: {} is a good way to do that. That's not a common convention in openapi specs, and if I just read the openapi doc without reading the spec, and saw additionalProperties: {}, I wouldn't know that the author was trying to say "there most be other specced properties dependent on the type property". Of course we could define it as a convention for the matrix openapi docs, and update the hugo templates to say something relevant, but it doesn't feel very intuitive to me.

Yeah, you're right.

> What if we extended the user_identifier definition to include a description on the additionalProperties, saying that?

This sounds better, yes, and actually lines up with what we already do elsewhere:

Screenshot 2024-07-24 at 12 52 58

layouts/partials/openapi/render-object-table.html
```diff
@@ -67,6 +67,9 @@
 
       TODO: support `patternProperties` here.
     */}}
+    {{ if eq .additionalProperties true }}
+      {{ errorf "`additionalProperties: true` is de facto the default for Matrix and doesn't provide any information" }}
+    {{ end }}
```
richvdh, line 72: If we're going to do this, maybe update the comment above which claims "nothing to do"?

Johennes: Actually, I think we don't even need this anymore. My original premise was that we wanted to prevent `additionalProperties: true` because it's not getting rendered in the spec. However, as we discussed, `additionalProperties: true` doesn't add any information on its own. So not rendering it is not problematic. Therefore, I have removed this from the PR.

data/api/client-server/definitions/user_identifier.yaml
```diff
@@ -18,7 +18,10 @@ type: object
...
     type: string
-    description: The type of identification.  See [Identifier types](/client-server-api/#identifier-types) for supported values and additional property descriptions.
+    description: |-
+      The type of identification. See [Identifier types](/client-server-api/#identifier-types)
+      for supported values and additional property descriptions.
 required:
   - type
-additionalProperties: true
+additionalProperties:
+  description: Keys dependent on the identification type
```
richvdh, line 27:
```suggestion
  description: Keys dependent on the identification type.
```

richvdh (approved):
Thanks! Sorry for the slow turnaround.

Johennes:
No worries. Thanks a lot for the review everyone. This has turned out much better in the end than what I was trying to do initially.

PR #1910, opened 2024-07-18
Don't mention that `GET /_matrix/client/v3/profile/{userId}` can return additional properties because this is true for almost every endpoint

Johennes:
The endpoint is [documented](https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3profileuserid) as:

> This API may return keys which are not limited to `displayname` or `avatar_url`.

This is superfluous though because its true for pretty much every endpoint and calling it out only here can be irritating.

~The response schema doesn't include `additionalProperties` though and, thus, no `<Other properties>` row is rendered.~

~This pull requests requires #1909 for the change to be visible in the rendered spec.~

### Pull Request Checklist

Preview: https://pr1910--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/profile.yaml
+++ b/data/api/client-server/profile.yaml
@@ -222,7 +222,7 @@ paths:
                     type: string
                     description: The user's display name if they have set one, otherwise not
                       present.
-                additionalProperties: true
+                additionalProperties: {}
               examples:
                 response:
                   value: {
--- a/data/api/client-server/profile.yaml
+++ b/data/api/client-server/profile.yaml
@@ -222,6 +222,7 @@ paths:
                     type: string
                     description: The user's display name if they have set one, otherwise not
                       present.
+                additionalProperties: true
               examples:
                 response:
                   value: {
--- a/data/api/client-server/profile.yaml
+++ b/data/api/client-server/profile.yaml
@@ -195,8 +195,7 @@ paths:
       description: |-
         Get the combined profile information for this user. This API may be used
         to fetch the user's own profile information or other users; either
-        locally or on remote homeservers. This API may return keys which are not
-        limited to `displayname` or `avatar_url`.
+        locally or on remote homeservers.
       operationId: getUserProfile
       parameters:
         - in: path
@@ -222,7 +221,6 @@ paths:
                     type: string
                     description: The user's display name if they have set one, otherwise not
                       present.
-                additionalProperties: {}
               examples:
                 response:
                   value: {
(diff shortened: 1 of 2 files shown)


Johennes:
Given @zecakeh's comment [here](https://github.com/matrix-org/matrix-spec/pull/1909/files#r1682420410), I think this should probably block on #1909.

Johennes:
This now uses `additionalProperties: {}` in accordance with #1909 on which it no longer depends.

data/api/client-server/profile.yaml
```diff
@@ -222,6 +222,7 @@ paths:
                     type: string
                     description: The user's display name if they have set one, otherwise not
                       present.
+                additionalProperties: {}
```
richvdh, line 225: As I've written over at https://github.com/matrix-org/matrix-spec/pull/1909#pullrequestreview-2194861986, it's not obvious to me that we should add this. IMHO it adds no extra information.

Johennes: Continuing the discussion from https://github.com/matrix-org/matrix-spec/pull/1909#issuecomment-2247610235, I think I agree that `additionalProperties: {}` doesn't add real information when this is technically the case for almost everything in Matrix.

I have considered https://spec.matrix.org/v1.11/client-server-api/#getwell-knownmatrixclient as a counter example.

```
additionalProperties:
  type: object
  description: Application-dependent keys using Java package naming convention.
```

Using Java package naming feels like it should be the default for any custom property. However, in this case the `additionalProperties` schema also limits the type to be an object. So this feels worth keeping indeed.

Since the schema of `additionalProperties` in `/_matrix/client/v3/profile/{userId}` doesn't need any type restriction, I concur, that adding it is pointless.

However, this now makes me wonder if we should remove the remark that the response can contain additional fields when this is true for almost every response?

> This API may return keys which are not limited to `displayname` or `avatar_url`.

richvdh: > However, in this case the `additionalProperties` schema also limits the type to be an object. So this feels worth keeping indeed.

As an aside, I'm unconvinced that it is correct that other properties in the `.well-known` result must be objects, but that's a different problem.

> However, this now makes me wonder if we should remove the remark that the response can contain additional fields when this is true for almost every response?
>
>>    This API may return keys which are not limited to displayname or avatar_url.

Yes, removing that seems reasonable to me.

Johennes: > As an aside, I'm unconvinced that it is correct that other properties in the `.well-known` result must be objects, but that's a different problem.

Yeah, I was wondering about that, too. Have logged https://github.com/matrix-org/matrix-spec/issues/1913.

Johennes: > > However, this now makes me wonder if we should remove the remark that the response can contain additional fields when this is true for almost every response? [...]
>
> Yes, removing that seems reasonable to me.

Have made it so.

richvdh (approved):
Thanks!

PR #1915, opened 2024-08-04
Add recommendations for mobile clients

agx:
This came out of a discussion at https://matrix.to/#/!XXSJTvRPInupfUgQVb:matrix.org/$IUx1flZV7iLWIviEeJWrmIw8LHgv4DhKGDKnO1hU688?via=pixie.town&via=matrix.org&via=element.io

When looking into adding UnifiedPush support to a client I was looking around for RFC style "Best Comon Practice" implementation hints but couldn't find any. Maybe it makes sense to add something to the documentation? I couldn't find a more sensible place than the spec if there is one please point me to it?

### Pull Request Checklist

* [ ] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog) (left out as this repo maybe isn't the right location to add this?)

### DIFF
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -1363,3 +1363,12 @@ following conventions.
    the tokens returned by `next_batch` and `prev_batch` should contain enough
    information for subsequent calls to the API to know which page of results
    they should return.
+
+## Best Current Practice for Mobile Clients
+
+In order to save bandwidth and battery mobile clients should:
+
+- Use [push notifications][/client-server-api##push-notifications] to get notified about new events
+- Disable continuous `/sync` calls
+- Fetch the events indicted in the push notifications via the ``GET /rooms/{roomId}/event/{eventId}`` API call
+- In case the event is encrypted and the necessary keys aren't present, invoke `/sync` to get the needed encryption keys


content/appendices.md
```diff
@@ -1363,3 +1363,12 @@ following conventions.
    the tokens returned by `next_batch` and `prev_batch` should contain enough
    information for subsequent calls to the API to know which page of results
    they should return.
+
+## Best Current Practice for Mobile Clients
+
+In order to save bandwidth and battery mobile clients should:
+
+- Use [push notifications][/client-server-api##push-notifications] to get notified about new events
```
anoadragon453, line 1371:
```suggestion
- Use [push notifications][/client-server-api#push-notifications] to get notified about new events
```

content/appendices.md
```diff
@@ -1363,3 +1363,12 @@ following conventions.
...
+- Disable continuous `/sync` calls
+- Fetch the events indicted in the push notifications via the ``GET /rooms/{roomId}/event/{eventId}`` API call
```
anoadragon453, line 1373:
```suggestion
- Fetch the events included in the push notifications via the ``GET /rooms/{roomId}/event/{eventId}`` API call
```

anoadragon453: Also, could you linkify `GET /rooms/{roomId}/event/{eventId}`? Likewise with `/sync`?

anoadragon453 (changes requested):
I think this may go better under https://spec.matrix.org/v1.11/client-server-api/#receiving-notifications.

@matrix-org/rust could I confirm whether this is the best practice for mobile developers to receive (non-)encrypted events?

bnjbvr:
> @matrix-org/rust could I confirm whether this is the best practice for mobile developers to receive (non-)encrypted events?

Yep, something like this. For what it's worth, instead of using `/context` or `/event` for retrieving the event, we're doing a limited sliding sync, because the two former endpoints don't work for stripped events (coming from invited rooms). And if the event was encrypted, we're indeed trying to run a sliding sync limited to the to-device and encryption extensions; this may change with the new evolution of the sliding sync protocol.

richvdh:
This feels like it has stalled. @agx: are you able to make the requested changes?

agx:
> This feels like it has stalled. @agx: are you able to make the requested changes?

Yes, I hope to get to this in the not too distant future.

richvdh:
I'm going to close this for now. If you'd like to pick this up again, it would be much appreciated: please just ping us and we can reopen.

Issue #1125, opened 2022-06-14
Different font for ∪ and ∩


richvdh:
**Link to problem area**:
https://spec.matrix.org/v1.2/rooms/v9/#definitions

**Issue**
![image](https://user-images.githubusercontent.com/1389908/173621628-29b5f9a7-04c7-468f-be34-fbdce9c630d5.png)

We seem to use Arial for the intersection, but it lacks a glyph for the union and we fall back to a default sans-serif font

PR #1919, opened 2024-08-07
Use dedicated fonts for better support of mathematical symbols

Johennes:
Fixes: #1125

This injects various fonts _before_ Arial (which has limited support for mathematical symbols):

* STIX Two Math – included in [macOS Catalina](https://www.idownloadblog.com/2020/05/28/macos-catalina-fonts-list/) and [Sonoma](https://support.apple.com/en-us/108939)
* Cambria Math – included in [Windows 10](https://learn.microsoft.com/en-us/typography/fonts/windows_10_font_list) and [11](https://learn.microsoft.com/en-us/typography/fonts/windows_11_font_list)
* Noto Sans Math – included on some Linux variants (e.g. Fedora) but might be ignored by browsers to avoid fingerprinting
* Dejavu Sans – included on some Linux variants (e.g. [Ubuntu](https://github.com/adrg/os-font-list/blob/master/md/linux/Ubuntu-21.04_Hirsute-Hippo.md) but not Fedora)

This is how it looks on macOS Sonoma:

Screenshot 2024-08-19 at 13 31 27

### Alternatives

The only other alternative I considered was using the n-ary union and intersection symbols ⋃ and ⋂. Those wouldn't require any font changes but are way bigger which might look even worse:

Screenshot 2024-08-07 at 15 23 39

### Pull Request Checklist

Preview: https://pr1919--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -42,16 +42,15 @@ $td-enable-google-fonts: false;
 /*
  * The $font-family-sans-serif definition here overrides the default value set by docsy
  * (https://github.com/matrix-org/docsy/blob/66a4e61d2d34edc7196b9df83a7d09cd4af14b47/assets/scss/_variables.scss#L68)
- * and:
- *
- * - Adds "Inter" to the front (making it the default)
- * - Inserts "Arial Unicode MS" before "Arial" (because the former contains more mathematical glyphs)
- *
- * The Inter font itself is loaded via stylesheet links layouts/partials/hooks/head-end.html.
  */
-$font-family-sans-serif: "Inter", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
-  "Helvetica Neue", Arial Unicode MS, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji",
-  "Segoe UI Symbol";
+$font-family-sans-serif:
+  // Add "Inter" to the front, making it the default. The font itself is loaded via stylesheet
+  // links in layouts/partials/hooks/head-end.html.
+  "Inter",
+  -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue",
+  // Insert fonts suited for mathematical symbols on different platforms before "Arial"
+  "STIX Two Math", "Cambria Math", "Noto Sans Math", "Dejavu Sans",
+  Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 
 // Disable smooth scrolling as it makes TOC highlighting jump during the transition.
 $enable-smooth-scroll: false;
--- a/assets/scss/_variables_project.scss
+++ b/assets/scss/_variables_project.scss
@@ -40,15 +40,18 @@ $table-bg: $secondary-lightest-background;
 $td-enable-google-fonts: false;
 
 /*
- * Replace the default font with Inter.
+ * The $font-family-sans-serif definition here overrides the default value set by docsy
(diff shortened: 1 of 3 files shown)


assets/scss/_variables_project.scss
```diff
@@ -40,15 +40,18 @@ $table-bg: $secondary-lightest-background;
...
- * https://github.com/matrix-org/docsy/blob/66a4e61d2d34edc7196b9df83a7d09cd4af14b47/assets/scss/_variables.scss#L68
- * and adds "Inter" to the front.
+ * - Adds "Inter" to the front (making it the default)
+ * - Inserts "Arial Unicode MS" before "Arial" (because the former contains more mathematical glyphs)
  *
- * The font itself is loaded via stylesheet link layouts/partials/hooks/head-end.html.
+ * The Inter font itself is loaded via stylesheet links layouts/partials/hooks/head-end.html.
  */
-$font-family-sans-serif: "Inter", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
+$font-family-sans-serif: "Inter", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
```
Mikaela, line 52: Isn't -apple-system a deprecated name for system-ui nowadays supported by ~everyone?

- https://caniuse.com/?search=system-ui

Johennes: As far as I understand, we simply "inherit" this from upstream. Would system-ui help fix #1125?

Mikaela: That would depend on what the client browser considered as the system font, but having so exotic characters, personally I would attempt to request known good fonts such as Noto.

anoadragon453: Noto Sans Math [supports these symbols](https://fonts.google.com/noto/specimen/Noto+Sans+Math/tester), and looks as I would expect:

Screenshot from 2024-08-12 10-13-27

I suppose the downside is that some browsers may need to download the font instead of using a system one.

Johennes: Maybe we add both Noto and Arial Unicode in the CSS list then?

richvdh:
Yeah, this doesn't help for me here on Linux, and I don't have Noto either.

How about  "DejaVu Sans" (which is what my browser falls back to for the ∪, and looks fine, as long as we don't use Arial for ∩). Don't know how widely supported that is though.

richvdh:
Oh wait, I do have Noto Sans. But Firefox is refusing to render the ∪ and ∩ glyphs in it and still falling back to a mix of Arial and DejaVu Sans.

Johennes:
> Oh wait, I do have Noto Sans. But Firefox is refusing to render the ∪ and ∩ glyphs in it and still falling back to a mix of Arial and DejaVu Sans.

That might possibly be due to anti-fingerprinting settings in Firefox. 🤔

anoadragon453:
> Oh wait, I do have Noto Sans. But Firefox is refusing to render the ∪ and ∩ glyphs in it and still falling back to a mix of Arial and DejaVu Sans.

Do you have "Noto Sans Math"?

Johennes:
I've updated the PR to use different fonts to cover more platforms. See the issue description for more info.

anoadragon453:
Looks marginally improved from NixOS (font weights on each symbol now match). My browser has ended up using DejaVu Sans.

Before:

image

After:

Screenshot from 2024-08-19 13-02-38

anoadragon453 (approved):
Definitely an improvement over the current state of things. Thanks!

PR #1921, opened 2024-08-09
#1383 issue : Update edu.yaml

AllMightLegend:
### Pull Request Checklist

"m.presence" is changed to "m.typing" as here edu has no reason for presence.
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

Preview: https://pr1921--matrix-spec-previews.netlify.app

PR #1923, opened 2024-08-09
Update edu.yaml

AllMightLegend:
"m.presence" is changed to "m.typing" as here edu does not need "m.presence"

### Pull Request Checklist


turt2live:
This appears to be a duplicate of https://github.com/matrix-org/matrix-spec/pull/1921

AllMightLegend:
yes sir but the thing is I have changed the edu.yaml file but the error is for checking the newsfragments which I have not even seen

PR #1924, opened 2024-08-10
#1383 issue : Update edu.yaml

AllMightLegend:
### Pull Request Checklist

Updated edu.yaml with "m.typing" as "m.presence" is not required. Discarded the changes made to history folder according to the advice.
Fixes #1383 as edu_type is primarily used in the typing notification docs.
P.S. Thank you sir

Preview: https://pr1924--matrix-spec-previews.netlify.app


KitsuneRal:
Thanks, this is better. Just as an aside - you don't have to (actually, shouldn't) make a new pull request every time - it's better to simply add commits to the existing pull request. That makes it much easier to trace your changes.

Please do not remove items from the pull request checklist. And please add the news fragment as the contributing guidelines describe. We usually employ the following line as the content of newsfragment file for changes like yours, so that we can easily group them: "Fix various typos throughout the specification."

AllMightLegend:
oh thank you sir

AllMightLegend:
Sir, Where do I mention this line "Fix various typos throughout the specification." ?? In the description ??

AllMightLegend:
I have created 1924.clarification in newsfragments and pushed it here . Is it okay now ??

richvdh:
Please update the summary of the pull request to explain what this is changing. 

You don't need to mention the issue number in the summary  - instead you can write something like `fixes #1383` in the description (ie, the first comment).

changelogs/server_server/newsfragments/1924.clarification
```diff
@@ -0,0 +1 @@
+In data/api/server-server/definitions/edu.yaml , edu type : content : "m.presence" is changed to "m.typing" as "m.presence" is useless here and not required
```
anoadragon453, line 1: A small update to the changelog to make it more readable, and to make use of inline code (`) instead of quotations.

If I'm understanding this change correctly, I believe this is the intended reason for the fix?

```suggestion
Update the default EDU example to be `m.typing` rather than `m.presence`, as it is primarily used in the typing notification docs.
```

AllMightLegend: But quotation is used in the syntax not inline code

tulir:
I don't see the reason for this change. The [typing notification docs](https://spec.matrix.org/v1.11/server-server-api/#typing-notifications) already have their own example, defined in https://github.com/matrix-org/matrix-spec/blob/main/data/api/server-server/definitions/event-schemas/m.typing.yaml. This example is only used in the [generic EDU docs](https://spec.matrix.org/v1.11/server-server-api/#edus), where the type is not really relevant.

#1383 is not an issue, it's a pull request which fixed the examples

AllMightLegend:
> I don't see the reason for this change. The [typing notification docs](https://spec.matrix.org/v1.11/server-server-api/#typing-notifications) already have their own example, defined in https://github.com/matrix-org/matrix-spec/blob/main/data/api/server-server/definitions/event-schemas/m.typing.yaml. This example is only used in the [generic EDU docs](https://spec.matrix.org/v1.11/server-server-api/#edus), where the type is not really relevant.
> 
> #1383 is not an issue, it's a pull request which fixed the examples

Oh okay sir

richvdh:
Let's close this then, since no explanation about why it is helpful has been provided.

PR #1928, opened 2024-08-12
Document that the spec uses RFC 2119

### LINKED ISSUE #1927: Document that the spec uses RFC 2119
**Link to problem area**: Yes

**Issue**
The spec uses RFC 2119, but doesn't document that. I would expect based on [Travis' statement](https://github.com/matrix-org/matrix-spec/issues/1782#issuecomment-2284286870) that fixing this would be a clarification and thus not require the MSC process.

https://tools.ietf.org/html/rfc2119

HarHarLinks:
Contributed with my community hat on.

I added it to the index because the index itself already starts using RFC 2119 language in section 3.

### Pull Request Checklist

Resolves #1927

Preview: https://pr1928--matrix-spec-previews.netlify.app

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -99,6 +99,22 @@ services - be that for instant messages, VoIP call setups, or any other
 objects that need to be reliably and persistently pushed from A to B in
 an interoperable and federated manner.
 
+### Bindingness of requirements
+
+As a standards document, the Matrix Spec needs to signify the bindingness of the
+individual features, APIs, procedures, or other requirements it describes.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" across all parts of the
+specification are to be interpreted as described in
+[RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+Some entire sections of the specification might be optional depending on the
+circumstances, for example the
+[modules of the client-server API](/client-server-api/#modules) apply depending
+on the use case. The bindingness of the above key words appearing in such a
+section is only relevant if the section itself is applicable.
+
 ### Spec Change Proposals
 
 To propose a change to the Matrix Spec, see the explanations at
--- a/content/_index.md
+++ b/content/_index.md
@@ -99,10 +99,7 @@ services - be that for instant messages, VoIP call setups, or any other
 objects that need to be reliably and persistently pushed from A to B in
 an interoperable and federated manner.
 
-### Bindingness of requirements
-
-As a standards document, the Matrix Spec needs to signify the bindingness of the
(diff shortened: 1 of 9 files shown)


HarHarLinks:
could you clarify where I should put the newsfragment for a change to the index?

content/_index.md
```diff
@@ -99,6 +99,22 @@ services - be that for instant messages, VoIP call setups, or any other
 objects that need to be reliably and persistently pushed from A to B in
 an interoperable and federated manner.
 
+### Bindingness of requirements
+
+As a standards document, the Matrix Spec needs to signify the bindingness of the
+individual features, APIs, procedures, or other requirements it describes.
```
richvdh, line 105: "Bindingness" is rather an awkward word!

Can we just say something like:

```suggestion
### Use of RFC2119 keywords
```

... and omit the first paragraph, which I don't think adds much value?

content/_index.md
```diff
@@ -99,6 +99,22 @@ services - be that for instant messages, VoIP call setups, or any other
...
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" across all parts of the
+specification are to be interpreted as described in
+[RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
+
+Some entire sections of the specification might be optional depending on the
+circumstances, for example the
```
richvdh, line 113:
```suggestion
circumstances. For example, the
```

content/_index.md
```diff
@@ -99,6 +99,22 @@ services - be that for instant messages, VoIP call setups, or any other
...
+[modules of the client-server API](/client-server-api/#modules) apply depending
+on the use case. The bindingness of the above key words appearing in such a
+section is only relevant if the section itself is applicable.
```
richvdh, line 116:
```suggestion
on the use case. The requirement level expressed by the above key words appearing in such a
section is only relevant if the section itself is applicable.
```

richvdh (changes requested):
some minor edits

content/_index.md
HarHarLinks: I tried to be more generic and may have ended up with a mangled Germenglish version. How do you think about something slightly more general and introductory than jumping straight to RFC 2119 in the title?

```suggestion
### Requirement levels
```

---

I'm slightly confused by the spelling in your proposal: Is it correct that contrary to MSCs, RFCs use a space before the number? RFC 2119 does where it references itself.

richvdh: > I'm slightly confused by the spelling in your proposal: Is it correct that contrary to MSCs, RFCs use a space before the number?

No, that's just me being a crank. The space is correct.

richvdh (approved):
LGTM. Thanks!

Issue #1682, opened 2023-11-29
Consider publishing a pre-made Markdown checklist alongside changelog


turt2live:
Several projects use our changelog and add checkboxes in front of each item to track their support for that release. We can probably make this easier with a `https://spec.matrix.org/v1.9/checklist.md` sort of thing.

Examples:
* https://github.com/matrix-org/synapse/issues/16709
* https://github.com/ruma/ruma/issues/1626

I'd suggest we go with doing something similar to Ruma's template - individual projects can edit the template further if they choose (like how Synapse does it).

clokep:
FWIW I'd be fine with the synapse one having extra checkboxes to check off. I usually just remove the clarifications, but I'm a fan of checking things off. 😀 I think Dendrite usually copies the Synapse ones too.

Johennes:
I've tried making a rough PoC for this in #1936. It's sort of crappy though and I would welcome input if anyone has some.

Johennes:
#1937 is a variant that turns the changelog page into a section and outputs checklists under `/changelog/$VERSION/checklist.md`.

PR #1937, opened 2024-08-22
Provide MD checklists for the changelog

Johennes:
Fixes: #1682

This changes `/changelog` to be a section and adds a separate output format to provide Markdown checklists under `/changelog/$VERSION/checklist.md`.

Example:

- HTML: https://pr1937--matrix-spec-previews.netlify.app/changelog/v1.11/
- MD: https://pr1937--matrix-spec-previews.netlify.app/changelog/v1.11/checklist.md

I've also added links to the checklists in each version's header table.

Screenshot 2024-09-09 at 09 33 35

Since each version's changelog has a separate page now, inlining all changelogs on the index page seemed superfluous. Similarly, having the index page only list historical spec versions felt odd. Therefore, I've moved the info about historical versions to its own page and made the index page redirect to the latest version's changelog page.

Screenshot 2024-09-09 at 09 34 37

### Pull Request Checklist

Preview: https://pr1937--matrix-spec-previews.netlify.app


zecakeh:
It seems strange to me to have the same content twice in "Changelog" and in the different versions. This also complicates the rendering because the levels of the headings should be different in both cases.

Maybe clicking on "Changelog" could just open the latest version?

Johennes:
> Maybe clicking on "Changelog" could just open the latest version?

Have made it so.

layouts/shortcodes/changelog/changelog-description.html
Johennes, line 1: I've removed this entirely because the added info felt superfluous. Every changelog lists changes made since the last release and the version is apparent via the URL and page heading.

scripts/generate-changelog.sh
```diff
@@ -22,12 +22,15 @@ towncrier --yes
...
+title: $VERSION
+type: docs
+outputs:
+  - html
+  - checklist
 date: $(date -Iseconds)
 ---
 EOF
-    if [ "$VERSION" = "vUNSTABLE" ]; then
-        echo "## Changes since last release"
```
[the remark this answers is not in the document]
Johennes: Yeah, it looks a bit odd. I've tried tweaking it along the lines of what you proposed in https://github.com/matrix-org/matrix-spec/pull/1937/commits/a120a427baf759f1c3a45389f964843fb71643c4.

zecakeh:
Thanks, it looks good. I like that when we hover a version in the sidebar, there's the full title that appears.

I am not a fan of the redirect on the "Changelog" page, I would prefer for the link to be the correct one, but I guess there is no simple way to do that with Hugo.

Johennes:
> I am not a fan of the redirect on the "Changelog" page, I would prefer for the link to be the correct one, but I guess there is no simple way to do that with Hugo.

Yeah, it's not pretty, especially because the index page briefly flashes before redirecting. Maybe we could render the latest version's changelog inline on the index page instead? That would effectively make `/changelog` a shortcut for `/changelog/$latest`.

layouts/shortcodes/changelog/changelogs.html
```diff
@@ -1,10 +1,8 @@
...
+   "content/changelogs" to the latest version's changelog page.
 */}}
 
-{{ with .Page.Resources.Match "*.md" }}
-{{ range ((sort . "Params.date" "desc")) }}
-{{ .RenderShortcodes }}
-{{ end }}
+{{ with index .Page.RegularPages.ByDate.Reverse 0 }}
+   <meta http-equiv="refresh" content="0; url={{ .Permalink }}" />
 {{ end }}
```
richvdh, line 8: we could add something like:

```yaml
manualLinkRelRef: unstable.md
```

to the front matter of `content/changelog/_index.html`, but it would need maintaining during releases. (We could hack the `generate-changelog.sh` script to update it?)

Or we could add more intelligence to `layouts/partials/sidebar-tree.html` to handle this case?

This is probably fine for now, and we could improve it in a future PR.

richvdh (approved):

richvdh:
I'm inclined to merge this if there are no pressing objections from other SCT members.

richvdh:
(I like breaking the changelog up into pages, irrespective of any benefits from the checklist)

clokep:
I'm a fan of the output, but haven't checked the implementation.

richvdh:
Let's ship it then

anoadragon453:
This appears to have generated broken links on the live site: https://github.com/matrix-org/matrix-spec/actions/runs/11070201829/job/30759445748#step:6:13

Johennes:
Hm, looks like these are from `link`s in the header:

```
<link rel="alternate" type="text/markdown" href="[/unstable/unstable/changelog/v1.5/checklist.md](view-source:file:///unstable/unstable/changelog/v1.5/checklist.md)">
```

Locally these look fine for me though:

```
<link rel="alternate" type="text/markdown" href="[//localhost:1313/changelog/v1.5/checklist.md](view-source:http://localhost:1313/changelog/v1.5/checklist.md)">
```

🤔

Johennes:
I've opened https://github.com/matrix-org/matrix-spec/pull/1954 to remove these links entirely.

zecakeh:
It's not the only URLs that are affected, the one from the `<meta http-equiv="refresh" …/>` is too, and is more problematic. I believe that changing `.Permalink` to `.RelPermalink` in `layout/shortcodes/changelog/changelogs.html` fixes it though.

zecakeh:
Basically, the problem seems to be the use of `canonifyURLs: true` with `--baseURL`. The former is deprecated so we should probably work on removing it (just removing it breaks some URLs right now).

Johennes:
> It's not the only URLs that are affected, the one from the `<meta http-equiv="refresh" …/>` is too, and is more problematic. I believe that changing `.Permalink` to `.RelPermalink` in `layout/shortcodes/changelog/changelogs.html` fixes it though.

Ah, I missed this. Thanks @zecakeh! Looks like `.RelPermalink` fixes this, indeed. I've opened https://github.com/matrix-org/matrix-spec/pull/1956.

PR #1938, opened 2024-08-26
MSC4151: Reporting rooms

Johennes:
As per https://github.com/matrix-org/matrix-spec-proposals/pull/4151.

The MSC proposed that clients should [hide reported rooms](https://github.com/matrix-org/matrix-spec-proposals/pull/4151/files#diff-0d9364264e36813883f313167d1b53b1bbaeb4ad6e687bfaba8aea6c16b2662dR55-R57). However, there isn't any API for ignoring rooms yet. So I left that part out of the spec as otherwise clients would have to invent a system for doing this across devices.

### Pull Request Checklist

Preview: https://pr1938--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1938.clarification
+++ b/changelogs/client_server/newsfragments/1938.clarification
@@ -1,1 +1,1 @@
-The `reason` parameter in `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}` can be omitted instead of left blank as per MSC2414.
+The `reason` parameter in `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}` can be omitted instead of left blank as per [MSC2414](https://github.com/matrix-org/matrix-spec-proposals/pull/2414).
--- a/changelogs/client_server/newsfragments/1938.clarification
+++ b/changelogs/client_server/newsfragments/1938.clarification
@@ -1,0 +1,1 @@
+The `reason` parameter in `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}` can be omitted instead of left blank as per MSC2414.
--- a/changelogs/client_server/newsfragments/1938.new
+++ b/changelogs/client_server/newsfragments/1938.new
@@ -1,1 +1,1 @@
-Add `POST /_matrix/client/v3/rooms/{roomId}/report` as per MSC4151.
+Add `POST /_matrix/client/v3/rooms/{roomId}/report` as per [MSC4151](https://github.com/matrix-org/matrix-spec-proposals/pull/4151).
--- a/changelogs/client_server/newsfragments/1938.new
+++ b/changelogs/client_server/newsfragments/1938.new
@@ -1,0 +1,1 @@
+Add `POST /_matrix/client/v3/rooms/{roomId}/report` as per MSC4151.
--- a/content/client-server-api/modules/report_content.md
+++ b/content/client-server-api/modules/report_content.md
@@ -5,9 +5,6 @@ Users may encounter content which they find inappropriate and should be
 able to report it to the server administrators or room moderators for
 review. This module defines a way for users to report content.
 
-Content is reported based upon a negative score, where -100 is "most
-offensive" and 0 is "inoffensive".
-
 #### Client behaviour
 
 {{% http-api spec="client-server" api="report_content" %}}
@@ -19,6 +16,16 @@ This may be a dedicated room to alert server administrators to the
 reported content or some other mechanism for notifying the appropriate
 people.
 
-{{< changed-in v="1.8" >}} The server MUST verify that the user 
-reporting the event is currently joined to the room the event is 
-in before accepting a report.
+Particularly during waves of a harmful content, users may report whole
+rooms instead of individual events. Server administrators and safety teams
+should, therefore, be cautious not to shut down rooms that might otherwise
+be legitimate.
+
+{{< changed-in v="1.8" >}} When processing event reports, servers MUST
+verify that the reporting user is currently joined to the room the event
+is in before accepting a report.
+
+Contrarily, servers MUST NOT restrict room reports based on whether or not the
+reporting user is joined to the room. This is because users can be exposed
+to harmful content without being joined to a room, for instance, through
+room directories.
--- a/content/client-server-api/modules/report_content.md
+++ b/content/client-server-api/modules/report_content.md
@@ -25,7 +25,7 @@ be legitimate.
 verify that the reporting user is currently joined to the room the event
 is in before accepting a report.
 
-Contrarily, servers MUST NOT restrict room reports based on whether or not the
-reporting user is joined to the room. This is because users can be exposed
-to harmful content without being joined to a room, for instance, through
-room directories.
+{{< added-in v="1.12" >}} Contrarily, servers MUST NOT restrict room reports
+based on whether or not the reporting user is joined to the room. This is
+because users can be exposed to harmful content without being joined to a
+room, for instance, through room directories.
--- a/content/client-server-api/modules/report_content.md
+++ b/content/client-server-api/modules/report_content.md
@@ -28,4 +28,4 @@ is in before accepting a report.
 {{< added-in v="1.12" >}} Contrarily, servers MUST NOT restrict room reports
 based on whether or not the reporting user is joined to the room. This is
 because users can be exposed to harmful content without being joined to a
-room, for instance, through room directories.
+room. For instance, through room directories or invites.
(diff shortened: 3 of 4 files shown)


content/client-server-api/modules/report_content.md
```diff
@@ -5,9 +5,6 @@ Users may encounter content which they find inappropriate and should be
 able to report it to the server administrators or room moderators for
 review. This module defines a way for users to report content.
 
-Content is reported based upon a negative score, where -100 is "most
-offensive" and 0 is "inoffensive".
-
```
Johennes, line 10: I removed this here because it only applies to event reporting and is repeated in the corresponding endpoint's documentation.

data/api/client-server/report_content.yaml
```diff
@@ -16,6 +16,69 @@ info:
...
+                  value: {
+                    "errcode": "M_NOT_FOUND",
+                    "error": "The room was not found."
+                  }
+        "429":
+          description: This request was rate-limited.
+          content:
+            application/json:
+              schema:
+                $ref: definitions/errors/rate_limited.yaml
```
Johennes, line 79: I'm not sure if this is technically correct since the MSC [strongly recommends](https://github.com/matrix-org/matrix-spec-proposals/pull/4151/files#diff-0d9364264e36813883f313167d1b53b1bbaeb4ad6e687bfaba8aea6c16b2662dR105) but doesn't require rate limiting.

data/api/client-server/report_content.yaml
```diff
@@ -29,7 +92,7 @@ paths:
         will require the homeserver to check whether a user is joined to
         the room. To combat this, homeserver implementations should add
         a random delay when generating a response.
-      operationId: reportContent
+      operationId: reportEvent
```
Johennes, line 95: I'm unsure what consequences changing this has but the old ID felt unsuitable now that there are two endpoints in the module.

data/api/client-server/report_content.yaml
```diff
@@ -62,7 +125,7 @@ paths:
                     and 0 is inoffensive.
                 reason:
                   type: string
-                  description: The reason the content is being reported. May be blank.
+                  description: The reason the content is being reported.
```
Johennes, line 128: I have removed this as proposed in [MSC2414](https://github.com/matrix-org/matrix-spec-proposals/pull/2414) and thought it aligns well enough with the rest of this pull request to sneak in. 😇

changelogs/client_server/newsfragments/1938.clarification
```diff
@@ -0,0 +1 @@
+The `reason` parameter in `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}` can be omitted instead of left blank as per MSC2414.
```
Johennes, line 1:
```suggestion
The `reason` parameter in `POST /_matrix/client/v3/rooms/{roomId}/report/{eventId}` can be omitted instead of left blank as per [MSC2414](https://github.com/matrix-org/matrix-spec-proposals/pull/2414).
```

changelogs/client_server/newsfragments/1938.new
```diff
@@ -0,0 +1 @@
+Add `POST /_matrix/client/v3/rooms/{roomId}/report` as per MSC4151.
```
Johennes, line 1:
```suggestion
Add `POST /_matrix/client/v3/rooms/{roomId}/report` as per [MSC4151](https://github.com/matrix-org/matrix-spec-proposals/pull/4151).
```

content/client-server-api/modules/report_content.md
```diff
@@ -19,6 +16,16 @@ This may be a dedicated room to alert server administrators to the
 reported content or some other mechanism for notifying the appropriate
 people.
 
-{{< changed-in v="1.8" >}} The server MUST verify that the user 
-reporting the event is currently joined to the room the event is 
-in before accepting a report.
+Particularly during waves of a harmful content, users may report whole
```
anoadragon453, line 19:
```suggestion
Particularly during waves of harmful content, users may report whole
```

content/client-server-api/modules/report_content.md
```diff
@@ -19,6 +16,16 @@ This may be a dedicated room to alert server administrators to the
...
+be legitimate.
+
+{{< changed-in v="1.8" >}} When processing event reports, servers MUST
+verify that the reporting user is currently joined to the room the event
+is in before accepting a report.
+
+{{< added-in v="1.12" >}} Contrarily, servers MUST NOT restrict room reports
+based on whether or not the reporting user is joined to the room. This is
+because users can be exposed to harmful content without being joined to a
+room, for instance, through room directories.
```
anoadragon453, line 31:
```suggestion
room. For instance, through room directories or invites.
```

data/api/client-server/report_content.yaml
anoadragon453: You wouldn't be able to rate-limit this endpoint without `429` being defined, so I believe that the MSC intended for such an error code to be spec'd. Adding the error code here does not require that the homeserver send it - just that the client must handle it should it come up. 

An MSC can't require implementations to add rate-limiting, only recommend they do.

data/api/client-server/report_content.yaml
anoadragon453: Looks like it's used to generate bindings, and as such we probably shouldn't change it...
https://github.com/matrix-org/matrix-spec/blob/2cbf6067a6d50ea2160e72d2a570f2cb8eeb0a20/meta/documentation_style.rst?plain=1#L113-L114

@turt2live curious if you happen to know more about this?

data/api/client-server/report_content.yaml
```diff
@@ -16,6 +16,69 @@ info:
...
+          application/json:
+            schema:
+              type: object
+              example: {
+                "reason": "this makes me sad"
+              }
+              properties:
+                reason:
+                  type: string
+                  description: The reason the room is being reported.
```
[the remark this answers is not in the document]
anoadragon453: Given the event reporting endpoint already has the parameters marked as optional, and MSC2414 was trying to match that endpoint, making them optional here too looks like the right thing to me.

data/api/client-server/report_content.yaml
```diff
@@ -16,6 +16,70 @@ info:
...
   version: 1.0.0
 paths:
+  "/rooms/{roomId}/report":
+    post:
+      x-addedInMatrixVersion: "1.12"
+      summary: Report a room as inappropriate.
+      description: |-
+        Reports a room as inappropriate to the server, which may then notify
+        the appropriate people. The caller is not required to be joined to the
+        room to report it.
```
anoadragon453, line 26: Just to make this point super clear:

```suggestion
        Reports a room as inappropriate to the server, which may then notify
        the appropriate people. How such information is delivered is left up to
        implementations. The caller is not required to be joined to the room to
        report it.
```

data/api/client-server/report_content.yaml
Johennes: Ah, thanks for explaining. That makes perfect sense.

data/api/client-server/report_content.yaml
Johennes: I suppose if we want to be safe keeping `operationId: reportContent` on the existing endpoint and `operationId: reportRoom` on the new one would not be the end of the world. So maybe reverting back would be best?

KitsuneRal: Indeed, `operationId` usually ends up as the method or class name in generated APIs, and therefore, may (actually will, in the vast majority of cases) break the generated code. However, given that even more fine-grained changes (e.g. adding a field in a request body) that are API-compatible in JSON may be breaking in the generated code (depending on the target language and/or the shape of the API) authors producing auto-generated code have to be careful with any upgrades of the API definitions. In particular, 1.12 has already introduced breaking changes elsewhere from the Quotient perspective. So I wouldn't be too concerned around it, just be cognisant that not everything back-compatible in JSON remains such in a native API.

Johennes: We concluded in #matrix-docs:matrix.org that changing the operation ID is not ideal but acceptable in this case as otherwise it'll be even harder to fix in future.

Johennes:
@anoadragon453 is there anything else you'd like me to add or change to help this land?

anoadragon453 (approved):
Ah, sorry. Note that if I'm not requested for review and there's no comments happening I won't get notified about a PR or see it in my review queue, so I'll forget about it :upside_down_face:

Thanks for the poke. This LGTM now!

data/api/client-server/report_content.yaml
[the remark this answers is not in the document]
turt2live: (for the historical record: MSC4151 deliberately went against the optionality of MSC2414 to reduce spam)

PR #1944, opened 2024-09-09
Add error codes from MSC4178

dbkr:
Specs https://github.com/matrix-org/matrix-spec-proposals/pull/4178

### Pull Request Checklist

Preview: https://pr1944--matrix-spec-previews.netlify.app


dbkr:
Hmm, good point. Element looks for that code so it looks like it probably ought to be on the list. I don't know whether this would need another MSC to add though, wdyt?

anoadragon453:
@dbkr Looks like this errcode comes from the [Identity Service API](https://spec.matrix.org/v1.12/identity-service-api/). So it is defined in Matrix, just not by the C-S spec. If we're being pedantic, this is technically a separate API and could have an MSC to add that errcode. But I suspect such an MSC would have no objection, as it's an obvious change.

So... to skip the time sink, I'd vote to skip the MSC stage and just add the errcode to the list of possible responses.

Issue #560, opened 2019-11-12
Clarify what query parameters mean in thirdparty endpoints


wfdewith:
The [`/_matrix/client/r0/thirdparty/location/{protocol}`](https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-thirdparty-location-protocol) endpoint specifies a `searchFields` query parameter with type `string`, but the description mentions it's: "one or more". What is the actual format for this query parameter?

The same is also true for [`/_matrix/client/r0/thirdparty/user/{protocol}`](https://matrix.org/docs/spec/client_server/r0.6.0#get-matrix-client-r0-thirdparty-user-protocol), the only difference being that the query parameter is called `fields...`. I found in [`api/openapi_extensions.md`](https://github.com/matrix-org/matrix-doc/blob/master/api/openapi_extensions.md) that it is supposed to mean "one or more", but no specification on the actual format.

deepbluev7:
The openapi spec should now be correct with https://github.com/matrix-org/matrix-spec/commit/14051847653383e432ebacae2cc430cc9ebce1e0 and https://github.com/matrix-org/matrix-spec/pull/1947. The text/rendering of the spec might still be unclear though.

For me it wasn't immediately obvious why that is the case though, so let me explain (for future me for example):
- The query parameters are now objects (of string: string)
- The query parameters use the default style: form
- In that case the explode option defaults to true, which expands the object into separate query parameters

References:
- https://swagger.io/docs/specification/describing-parameters/#query-parameters
- https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.0.3.md#fixed-fields-10

richvdh:
Since things have changed a bit since this issue was opened, I've opened a new one to summarise the situation: https://github.com/matrix-org/matrix-spec/issues/1993.

PR #1947, opened 2024-09-11
Clarify parameter definition for third-party location lookups

deepbluev7:
While 14051847653383e432ebacae2cc430cc9ebce1e0 redefined most fields to be the appropriate object type, it missed one of them.

### Pull Request Checklist

Preview: https://pr1947--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/1947.clarification
+++ b/changelogs/client_server/newsfragments/1947.clarification
@@ -1,1 +1,1 @@
-Clarify query parameters for third-party location lookups.
+Correct OpenAPI specification for query parameters to `GET /_matrix/client/v3/thirdparty/location/{protocol}` endpoint.
--- a/changelogs/client_server/newsfragments/1947.clarification
+++ b/changelogs/client_server/newsfragments/1947.clarification
@@ -1,0 +1,1 @@
+Clarify query parameters for third-party location lookups.
--- a/data/api/client-server/third_party_lookup.yaml
+++ b/data/api/client-server/third_party_lookup.yaml
@@ -98,12 +98,14 @@ paths:
           schema:
             type: string
         - in: query
-          name: searchFields
+          name: fields
           description: |-
             One or more custom fields to help identify the third-party
             location.
           schema:
-            type: string
+            type: object
+            additionalProperties:
+              type: string
       responses:
         "200":
           description: At least one portal room was found.


data/api/client-server/third_party_lookup.yaml
```diff
@@ -98,12 +98,14 @@ paths:
           schema:
             type: string
         - in: query
-          name: searchFields
+          name: fields
```
KitsuneRal, line 101: Why change the name? `searchFields` looks appropriate.

deepbluev7: I think it makes sense to use the same name everywhere:
- https://spec.matrix.org/v1.11/client-server-api/#get_matrixclientv3thirdpartyuserprotocol
- https://spec.matrix.org/v1.11/application-service-api/#get_matrixappv1thirdpartylocationprotocol
- https://spec.matrix.org/v1.11/application-service-api/#get_matrixappv1thirdpartyuserprotocol

For some reason the exact same API on the appservice endpoints uses `fields` and not `searchFields`. And the similar API to find users for a protocol also uses `fields` on both the CS and AS API. I think it makes sense to be consistent there, but I can revert that part, if you would prefer that.

richvdh:
Sorry for the delay in reviewing this.

It took me a while to figure out what is going on here; but with reference to [a previous discussion thread](https://github.com/matrix-org/matrix-spec/pull/1310#discussion_r1054701952) and the OpenAPI spec, I think I've figured it out. I'll write down my conclusions for the sake of future archaeologists (including myself).

As I understand it, this is replacing the `searchFields` query-parameter of the  [`GET /_matrix/client/v3/thirdparty/location/{protocol}`](https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientv3thirdpartylocationprotocol), which is a string, with a pseudo-parameter named `fields`, which is specified as an object.

The relevant part of the OpenAPI spec is the definition of the [Parameter Object](https://swagger.io/specification/#parameter-object); in particular the rules for serialization of the parameter. The relevant OpenAPI fields are `style` (which is specified to default to `form` for query parameters) and `explode` (which defaults to `true` when `style` is `form`). The [style examples](https://swagger.io/specification/#style-examples) table then gives an example serialization of `R=100&G=200&B=150` for an object containing keys `R`, `G` and `B`.

[I now see that @deepbluev7 has reached much the same conclusion at https://github.com/matrix-org/matrix-spec/issues/560#issuecomment-2344138695.]

It's worth noting that the *name* of the object parameter (`fields`) is never actually used in the API. AIUI, this is being changed from `searchFields` for consistency with the definitions of other, similar, APIs.

Looking at https://github.com/matrix-org/matrix-spec/issues/560, and https://github.com/matrix-org/matrix-spec/pull/1584, this appears to be a correct change.

changelogs/client_server/newsfragments/1947.clarification
```diff
@@ -0,0 +1 @@
+Clarify query parameters for third-party location lookups.
```
richvdh, line 1:
```suggestion
Correct OpenAPI specification for query parameters to `GET /_matrix/client/v3/thirdparty/location/{protocol}` endpoint.
```

richvdh (approved):

richvdh:
I've opened https://github.com/matrix-org/matrix-spec/issues/1993 to track the fact that the spec text makes a mess of this.

Issue #1962, opened 2024-10-05
Matrix Reference Implementation


Maniues:
**Link to problem area**:
Current Matrix Specification states that Matrix is fully open:
```
Fully open:

    Fully open federation - anyone should be able to participate in the global Matrix network
    Fully open standard - publicly documented standard with no IP or patent licensing encumbrances
    **Fully open source reference implementation - liberally-licensed example implementations with no IP or patent licensing encumbrances**
```
**Issue**
After Element's fork of Synapse, Matrix has **no reference implementation**, at least it is not liberally licensed.
If the Matrix Foundation does not maintain any reference implementation, the specification should not reflect this. 

Worse, if there is no implementation that could be fully compliant with the specification, stable and openly managed, I do not know if then it is possible to call Matrix fully open in the sense of the current specification and the Matrix Foundation position.

clokep:
> **Link to problem area**:
> Current Matrix Specification states that Matrix is fully open:

The link is https://spec.matrix.org/latest/#introduction-to-the-matrix-apis

> After Element's fork of Synapse, Matrix has **no reference implementation**, at least it is not liberally licensed.

Synapse has not considered itself a reference implementation for several years (long before the fork by Element). (This likely doesn't change your concern.)

richvdh:
Indeed; cf https://github.com/matrix-org/synapse/pull/10971.

We should just remove this mention of "reference implementation" from the spec.

richvdh:
I did, but I don't have a good answer for you. It just doesn't? Nobody has found it a worthwhile way to spend their time.

PR #1966, opened 2024-10-10
Remove note about reference implementations

Johennes:
Fixes: #1962

### Pull Request Checklist

Preview: https://pr1966--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/appendices/newsfragments/1966.clarification
+++ b/changelogs/appendices/newsfragments/1966.clarification
@@ -1,0 +1,1 @@
+Remove note about reference implementations.
--- a/content/_index.md
+++ b/content/_index.md
@@ -56,9 +56,6 @@ The principles that Matrix attempts to follow are:
         the global Matrix network
     -   Fully open standard - publicly documented standard with no IP or
         patent licensing encumbrances
-    -   Fully open source reference implementation - liberally-licensed
-        example implementations with no IP or patent licensing
-        encumbrances
 -   Empowering the end-user
     -   The user should be able to choose the server and clients they
         use


changelogs/appendices/newsfragments/1966.clarification
Johennes, line 1: Note sure if this is the right place but #1627 also put its changelog there.

content/_index.md
```diff
@@ -56,9 +56,6 @@ The principles that Matrix attempts to follow are:
         the global Matrix network
     -   Fully open standard - publicly documented standard with no IP or
         patent licensing encumbrances
-    -   Fully open source reference implementation - liberally-licensed
-        example implementations with no IP or patent licensing
-        encumbrances
```
Gnuxie, line 61: image

Johennes: Oh, where is this from?

Gnuxie: It's from the foundation's rules (Guiding Principles) https://matrix.org/media/2019-06-10%20-%20Matrix.org%20Foundation%20CIC%20Rules.pdf

Johennes: Hm, but then there are no reference implementations in existence today? I'm not sure if the foundation still considers this as one of its future goals. If it does, I guess we can also keep the part of the spec that mentions reference implementations.

richvdh: Regardless of the *aspirations* of the foundation, there are currently no reference implementations (of either clients or servers), so the spec should not claim that there are.

richvdh (approved):

content/_index.md
HarHarLinks: This is the list of
> The principles that Matrix attempts to follow

Are they not "aspirations"?

ara4n: the text says: "In particular, it is envisaged that the Foundation will undertake the following activities:" - i.e. these are aspirations as written in 2018. Other details there (e.g. referring to an "mx:// URI scheme") are also wrong. The bug here is that the Rules haven't been updated since first being conceived, and are overdue being updated to reflect how the Foundation's scope has evolved (by the guardians in consultation with the governing board).

HarHarLinks: I am referring to https://github.com/Johennes/matrix-spec/blob/083e6ef25da712eecc1b56a8e3a2fafaf48d130d/content/_index.md?plain=1#L48

It seems incorrectly phrased in the spec then and should also have been adjusted with this PR. Or perhaps this list should not be here at all in favor of a link to the document Gnuxie cited.

richvdh: Ugh, I hadn't realised that the heading of this list was "The principles that Matrix attempts to follow are:". I thought this bit of the spec was just a statement of fact, rather than a list of what the Matrix project is *trying* to achieve. I guess I should have read more carefully. (In my defence, I'd rather bias to action than agonise over every change.)

But it does make this PR less obviously correct. If time and money were unlimited, would the Matrix Foundation nowadays seek to maintain liberally-licensed reference implementations? Probably? It's not obvious (eg, W3C does not maintain reference implementations for HTML.) It's an open question, to be resolved by the Matrix guardians and governing board. 

I guess it feels like this whole section of the spec is a bit outdated and out-of-place. (AFAICT, it was written in 2014 and has been barely touched since.) I'd rather the spec were stating the current fact, and the hopes and dreams of the matrix project were written elsewhere.

Not entirely sure how best to proceed here, but I'm inclined to take time to think about it rather than rush anything.

PR #1967, opened 2024-10-10
Sort VoIP events semantically

Johennes:
This introduces an `x-weight` property that can be used to influence the order of events rendered with the `event-group` shortcode and uses it to put the VoIP events into semantic order.

Fixes: #937

### Pull Request Checklist

Preview: https://pr1967--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1967.clarification
+++ b/changelogs/internal/newsfragments/1967.clarification
@@ -1,0 +1,1 @@
+Add `x-weight` property for sorting events rendered with the `event-group` shortcode.
--- a/data/event-schemas/schema/m.call.answer.yaml
+++ b/data/event-schemas/schema/m.call.answer.yaml
@@ -1,6 +1,7 @@
 {
     "type": "object",
     "description": "This event is sent by the callee when they wish to answer the call.",
+    "x-weight": 40,
     "allOf": [{
         "$ref": "core-event-schema/room_event.yaml"
     }],
--- a/layouts/shortcodes/event-group.html
+++ b/layouts/shortcodes/event-group.html
@@ -14,20 +14,24 @@
 
 {{ $base_path := "event-schemas/schema" }}
 
-{{ $events := index .Site.Data "event-schemas" "schema" }}
 {{ $group_name := .Params.group_name }}
 
-{{ range $event_name, $event_data := $events }}
-
+{{/* Filter events and prepare them for sorting */}}
+{{ $events := slice }}
+{{ range $event_name, $event_data := index .Site.Data "event-schemas" "schema" }}
     {{ $prefix := substr $event_name 0 (len $group_name) }}
     {{ if eq $prefix $group_name }}
-
-        {{ $path := delimit (slice $base_path $event_name) "/" }}
-        {{ $event_data = partial "json-schema/resolve-refs" (dict "schema" $event_data "path" $path) }}
-        {{ $event_data := partial "json-schema/resolve-allof" $event_data }}
-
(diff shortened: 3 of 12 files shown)


data/event-schemas/schema/m.call.answer.yaml
Johennes, line 1: Unrelated: I'm not sure why some of these are named `*.yaml` when they seem to be JSON?

clokep: YAML is a superset of JSON so they're still valid. 🙃

Johennes: Oh, right. 🤦‍♂️

changelogs/internal/newsfragments/1967.clarification
```diff
@@ -0,0 +1 @@
+Add `x-weight` property for sorting events rendered with the `event-group` shortcode.
```
zecakeh, line 1: The custom property should be documented in `/openapi_extensions.md`.

Johennes: Thanks, I wasn't aware the custom keys are documented there. Have added `x-weight`.

layouts/shortcodes/event-group.html
```diff
@@ -14,20 +14,24 @@
...
     {{ if eq $prefix $group_name }}
+        {{ $events = $events | append (dict "event_name" $event_name "event_data" $event_data) }}
+    {{ end }}
+{{ end }}
 
-        {{ $path := delimit (slice $base_path $event_name) "/" }}
-        {{ $event_data = partial "json-schema/resolve-refs" (dict "schema" $event_data "path" $path) }}
-        {{ $event_data := partial "json-schema/resolve-allof" $event_data }}
+{{/* Render the events sorted by x-weight or otherwise their event name */}}
+{{ range sort $events "event_data.x-weight" "event_name" }}
```
anoadragon453, line 29: Does this work? It looks like the only thing you can put as the third argument to `sort` is `"asc"` or `"desc"`: https://gohugo.io/functions/collections/sort/

openapi_extensions.md
```diff
@@ -41,3 +41,8 @@ same use as `format`, but that applies to the pattern of the property. We also
 define custom values for formats with the `mx-` prefix in
 `data/custom-formats.yaml`. Those values are recognized in the rendered
 specification and link to the definition of the format.
+
+## Custom `x-weight` key
```
anoadragon453, line 45: Unless `x-weight` is a colloquial identifier, `x-sort-weight` may be slightly more self-descriptive?

layouts/shortcodes/event-group.html
Johennes: My bad and thanks for catching. I thought this worked because the output matched what I expected but it turns out this was just coincidence. The `sort` function silently [defaults](https://github.com/gohugoio/hugo/blob/4985be1a4af34dd835e7882605e429cd5ffb1912/tpl/collections/sort.go#L69) to `"asc"` if the second argument is anything other than `"desc"`.

I believe there isn't a strong use case for ties anyway so I simply made it sort by `x-weight` only and rebased the weights of call events to not cause ties.

openapi_extensions.md
Johennes: I don't feel strongly but I thought it might be good to match the [`weight`](https://gohugo.io/methods/page/weight/) frontmatter property on pages?

anoadragon453: Ah, OK, I can rally behind that! `x-weight` sounds good.

anoadragon453 (approved):
This LGTM, thank you!

PR #1970, opened 2024-10-13
Remove `boxes/added-in-paragraph` shortcode

### LINKED ISSUE #1204: Client-server spec gets processed for Markdown twice

zecakeh:
It was added to work around a bug that made the `added-in` and `changed-in` partials appear as separate blocks rather than inline. This issue does not happen anymore.

Links to the areas that used it, to check that there is no issue anymore:

- [CS-API - Content Repository](https://pr1970--matrix-spec-previews.netlify.app/client-server-api/#content-repository)
- [SS-API - Content Repository](https://pr1970--matrix-spec-previews.netlify.app/server-server-api/#content-repository)

Fixes #1204.

### Pull Request Checklist

Preview: https://pr1970--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Nice, thanks for cleaning this up.

PR #1971, opened 2024-10-13
Remove `withVersioning` parameter of `rver-fragment` shortcode.

zecakeh:
Since #1884, we detect automatically in the `added-in` and `changed-in` shortcodes whether we need to add text according to the page's room version so this parameter has become unnecessary.
Moreover, forgetting to set this param can remove the text generated by the shortcodes so it is safer to just get rid of it.

Case in point, two `[New in this version]` were missing:
- Room v3 - Handling redactions ([before](https://spec.matrix.org/v1.12/rooms/v3/#handling-redactions) / [after](https://pr1971--matrix-spec-previews.netlify.app/rooms/v3/#handling-redactions))
- Room v4 - Event IDs ([before](https://spec.matrix.org/unstable/rooms/v4/#event-ids) / [after](https://pr1971--matrix-spec-previews.netlify.app/rooms/v4/#event-ids))

### Pull Request Checklist

Preview: https://pr1971--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1971.clarification
+++ b/changelogs/internal/newsfragments/1971.clarification
@@ -1,0 +1,1 @@
+Remove `withVersioning` parameter of `rver-fragment` shortcode.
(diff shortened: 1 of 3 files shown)


anoadragon453 (approved):
Thanks!

PR #1974, opened 2024-10-14
Clarify that pusher URLs have a fixed path

Johennes:
Fixes: #809

### Pull Request Checklist

### DIFF
--- a/data/api/push-gateway/push_notifier.yaml
+++ b/data/api/push-gateway/push_notifier.yaml
@@ -35,9 +35,9 @@ paths:
         updating counts of unread notifications should be idempotent and
         therefore do not require duplicate suppression.
 
-        Notifications are sent to the URL configured when the pusher is created.
-        This means that the HTTP path may be different depending on the push
-        gateway.
+        Clients interested in receiving notifications via this endpoint MUST
+        configure its concrete URL when creating the associated pusher via
+        [`/_matrix/client/v3/pushers/set`](/client-server-api/#post_matrixclientv3pushersset).
       operationId: notify
       requestBody:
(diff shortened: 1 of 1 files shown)


data/api/push-gateway/push_notifier.yaml
```diff
@@ -35,9 +35,9 @@ paths:
         updating counts of unread notifications should be idempotent and
         therefore do not require duplicate suppression.
 
-        Notifications are sent to the URL configured when the pusher is created.
-        This means that the HTTP path may be different depending on the push
-        gateway.
+        Clients interested in receiving notifications via this endpoint MUST
+        configure its concrete URL when creating the associated pusher via
+        [`/_matrix/client/v3/pushers/set`](/client-server-api/#post_matrixclientv3pushersset).
```
anoadragon453, line 40: I'm not sure if this really tells me "the path must be `/_matrix/push/v1/notify`".

What do you mean by "concrete URL"?

Johennes: Sorry, "concrete" is probably a poor translation from German. I meant the URL passed to the `data.url` field in ` /_matrix/client/v3/pushers/set`. Maybe "MUST configure its _full_ URL" would be better?

Regarding the path restriction, nothing else in the push gateway spec other than the sentence I removed hints at the path actually being allowed to differ. The only ambiguous place is the `url` field in the linked CS endpoint which already contains the path requirement. Do you think this should be duplicated here?

Johennes: @anoadragon453 what do you think?

PR #1975, opened 2024-10-15
Always use `%` delimiter for `added-in` and `changed-in` shortcodes

zecakeh:
The `<>` delimiters are not necessary for the shortcode to be rendered inline, and in some cases they break some expectations: a shortcode that is separated from other text to be in its own paragraph is not actually wrapped by a `p` element, breaking the spacing between paragraphs.

Previously affected sections:
- `.m.rule.is_user_mention` ([before](https://spec.matrix.org/v1.12/client-server-api/#_m_rule_is_user_mention) / [after](https://pr1975--matrix-spec-previews.netlify.app/client-server-api/#_m_rule_is_user_mention))
- `.m.rule.is_room_mention` ([before](https://spec.matrix.org/v1.12/client-server-api/#_m_rule_is_room_mention) / [after](https://pr1975--matrix-spec-previews.netlify.app/client-server-api/#_m_rule_is_room_mention))
- Server-side aggregation of `m.replace` relationships ([before](https://spec.matrix.org/v1.12/client-server-api/#server-side-aggregation-of-mreplace-relationships) / [after](https://pr1975--matrix-spec-previews.netlify.app/client-server-api/#server-side-aggregation-of-mreplace-relationships))

### Pull Request Checklist

Preview: https://pr1975--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/1975.clarification
+++ b/changelogs/internal/newsfragments/1975.clarification
@@ -1,1 +1,1 @@
-Always use `%` delimiter for `added-in` and `changed-in` shortcodes.
+Fix formatting of `added-in` and `changed-in` shortcodes by using `%` delimiter.
--- a/changelogs/internal/newsfragments/1975.clarification
+++ b/changelogs/internal/newsfragments/1975.clarification
@@ -1,0 +1,1 @@
+Always use `%` delimiter for `added-in` and `changed-in` shortcodes.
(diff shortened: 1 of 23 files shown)


changelogs/internal/newsfragments/1975.clarification
```diff
@@ -0,0 +1 @@
+Always use `%` delimiter for `added-in` and `changed-in` shortcodes.
```
richvdh, line 1:
```suggestion
Fix formatting of `added-in` and `changed-in` shortcodes by using `%` delimiter.
```

PR #1977, opened 2024-10-15
Rename `custom-formats.yaml` to `string-formats.yaml`

zecakeh:
It does not contain only custom formats, so the name didn't fit anymore. Update also the docs to reflect that.

While editing this file, change the URL of RFC3986 because the previous URL was a redirect.

### Pull Request Checklist

Preview: https://pr1977--matrix-spec-previews.netlify.app


data/string-formats.yaml
```diff
@@ -12,15 +12,26 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# This file contains the list of custom formats supported for the `format` key
+# This file contains the list of string formats supported for the `format` key
 # and the `x-pattern-format` extension (see `openapi_extensions.md` for more
 # details).
```
richvdh, line 17: Some ideas for clarification while we're here.

(It took me embarassingly long to realise this was talking about `format` within jsonschema, rather than, say, within [PusherData](https://spec.matrix.org/v1.12/client-server-api/#post_matrixclientv3pushersset_request_pusherdata))

```suggestion
# This file contains the list of string formats supported for the `format` key
# and the `x-pattern-format` extension within JSON Schema objects in the
# OpenAPI documents within the Matrix Specification.  See 
# `openapi_extensions.md` [1] for more details.
#
# [1]: https://github.com/matrix-org/matrix-spec/blob/main/openapi_extensions.md#custom-x-pattern-format-key-and-custom-formats for
```

zecakeh: Applied (with fixes).

richvdh:
Thanks!

PR #1978, opened 2024-10-15
Clarify formats of string types for the `GET /.well-known/matrix/support` endpoint

zecakeh:
[Rendered definition](https://pr1978--matrix-spec-previews.netlify.app/client-server-api/#getwell-knownmatrixsupport)

### Pull Request Checklist

Preview: https://pr1978--matrix-spec-previews.netlify.app


data/custom-formats.yaml
```diff
@@ -50,4 +50,7 @@ mx-mxc-uri:
 uri:
   title: URI
   url: http://tools.ietf.org/html/rfc3986
-  # no regex
```
anoadragon453, line 53: Did you mean to remove the `# no regex` comment, and also not add one for `email`.

And if so, why?

data/custom-formats.yaml
```diff
@@ -50,4 +50,7 @@ mx-mxc-uri:
 uri:
   title: URI
   url: http://tools.ietf.org/html/rfc3986
```
anoadragon453, line 52: While we're here, could we switch this to an HTTPS url please?

data/custom-formats.yaml
zecakeh: Yes, because we updated the docs in #1977 to say that only custom formats need a regex. And the comment was removed there as well.

data/custom-formats.yaml
zecakeh: This URL was updated in #1977.

zecakeh:
I merged main to solve the conflict.

anoadragon453 (approved):
Looks good to me, thank you!

PR #1984, opened 2024-10-24
Fix relative URLs when serving the specification with a custom `baseURL`

### LINKED ISSUE #1957: Stop using canonifyURLs
As reported by @zecakeh in https://github.com/matrix-org/matrix-spec/pull/1937#issuecomment-2383089818, the `canonifyURLs: true` setting appears to be deprecated and causes issues when used together with `--baseURL`. We should probably work on removing `canonifyURLs`.

### LINKED ISSUE #1964: Historical spec artifacts starting with v1.11 have broken CSS
**Link to problem area**: spec-historical of https://github.com/matrix-org/matrix-spec/releases/tag/v1.11

**Issue**
image

compared to:
image

**Expected behaviour**
Implied above.

zecakeh:
~~Requires https://github.com/matrix-org/docsy/pull/4.~~

Using `canonifyURLs` and a `baseURL` that is not `/` at the same time makes `.RelPermalink` and `relref` return relative URLs where the first path segment is duplicated, e.g. `/v1.11/v1.11/rest/of/the/path`.

`canonifyURLs` is deprecated so let's get rid of it and allow Hugo to compute the relative URLs itself, because it keeps track of the resources and knows where they will be served.

This can be reviewed commit by commit.

Since the PR CI uses `/` as a `baseURL`, I guess the only way to test that this works before it is deployed is to generate the spec locally with a custom `baseURL` like `/v1.11` and make sure that there are no URLs where it is duplicated. 

Closes #1957.\
Fixes the causes behind #1964, but the artifacts that were already generated for v1.11 and v1.12 still need to be fixed somehow.

### Pull Request Checklist

Preview: https://pr1984--matrix-spec-previews.netlify.app


zecakeh:
Leaving as a draft for now, it turns out that it creates broken links in places (see [CI](https://github.com/matrix-org/matrix-spec/actions/runs/11497753686/job/32002168281?pr=1984)).

zecakeh:
Alright the issues have been fixed, but I am leaving this as a draft for now until the docsy PR is merged.

So the good news is that now some links that would have been broken in the OpenAPI definitions are also broken in the rendered spec, and we detect them in CI thanks to htmltest.

config.toml
```diff
@@ -138,6 +138,15 @@ sidebar_menu_compact = true
   [[module.imports]]
     path = "github.com/matrix-org/docsy"
     disable = false
+  # Mounting /assets and /static as assets directories helps Hugo to generate
+  # the correct URL.
+  [[module.mounts]]
+    source = 'assets'
+    target = 'assets'
```
anoadragon453, line 141: Is this needed to ensure that `assets` remains mounted to `assets` (the default, but defaults are not kept when defining any `module.mounts` entries)? If so, could you leave a comment noting that? Otherwise I fear this may appear redundant and be accidentally removed.

config.toml
```diff
@@ -138,6 +138,15 @@ sidebar_menu_compact = true
...
+  [[module.mounts]]
+    source = 'static'
+    target = 'assets'
```
anoadragon453, line 144: Since we're also mounting both `static` and `assets` to the same place, we should leave a note to say that any files named the same in both directories will conflict (I'm don't believe Hugo will error in an obvious way in that case). In that case, files in `static` will win, since it's mounted second.

.github/workflows/main.yml
```diff
@@ -1,4 +1,9 @@
 name: "Spec"
+
+env:
+  # Cannot build the spec with Hugo 0.125.0 and docsy < 0.10.0 because of https://github.com/google/docsy/issues/1930
+  HUGO_VERSION: 0.124.1
```
anoadragon453, line 5: Could we bump this up further now, since we're on docsy v0.11.0?

anoadragon453 (approved):
This LGTM, thank you for your work!

> Fixes the causes behind https://github.com/matrix-org/matrix-spec/issues/1964, but the artifacts that were already generated for v1.11 and v1.12 still need to be fixed somehow.

Would manual find/replace surgery be tractable, or do we need to figure out how to regenerate? We can also discuss this further on a new issue to avoid bumping a merged PR.

go.mod
```diff
@@ -2,4 +2,4 @@ module github.com/matrix-org/matrix-spec
 
 go 1.12
 
-require github.com/matrix-org/docsy v0.0.0-20231227121313-d55dbaf3aaaf // indirect
+require github.com/matrix-org/docsy v0.0.0-20241030165433-a14705b7e3ca // indirect
```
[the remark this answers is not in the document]
zecakeh: I planned to upgrade docsy in a separate PR because it brings visual changes that might need to be discussed.

zecakeh:
> Would manual find/replace surgery be tractable, or do we need to figure out how to regenerate? We can also discuss this further on a new issue to avoid bumping a merged PR.

There must be something like 30 invalid links in those artifacts so it is doable manually. Running htmltest detects all the issues.

But it should also be simple to regenerate, we just need to use Hugo 0.117.0, like the CI that generates the "non-historical" spec-artifact. We can discuss this in the original issue, I already left a few comments there: https://github.com/matrix-org/matrix-spec/issues/1964.

config.toml
zecakeh: Yes, if we override a mount, Hugo ignores the default.

anoadragon453: Could you leave a comment noting that? :)

config.toml
anoadragon453: Oop, indeed we are! Would carrying out that alternative solution mean we'd stop unintentionally serving those files? That sounds preferable?

anoadragon453 (commented):
Blocking on the threads above.

config.toml
zecakeh: I think so. Everything in `/static` is copied as-is in the rendered website. Files in `/asset` are only in the rendered website if they are in use.

zecakeh: I added a commit that does just that, and merged main to solve a conflict because of it.

config.toml
zecakeh: This is not a problem anymore with the latest commits.

anoadragon453 (approved):
This now LGTM. Thanks for your patience!

PR #1996, opened 2024-11-14
Upgrade docsy to 0.11.0 and hugo to 0.139.0

zecakeh:
Here are the visual changes I noticed in the rendered spec:

- Links are underlined, which is a [good practice for accessibility](https://webaim.org/techniques/hypertext/link_text#underlining)
- The background color of the inline code is lighter

Example, in the Transaction identifiers section:

[Before:](https://spec.matrix.org/v1.12/client-server-api/#transaction-identifiers)

image

[After:](https://pr1996--matrix-spec-previews.netlify.app/client-server-api/#transaction-identifiers)

image

Another change I noticed in the generated HTML but that is not visible is that some uses of shortcodes generate empty paragraphs now (`<p></p>`). Since it doesn't change the rendering, I guess that browsers are smart enough to ignore them.

### Pull Request Checklist

Preview: https://pr1996--matrix-spec-previews.netlify.app


MTRNord:
About the empty p tags: that may or may not (I am only marginaly into those Tools) be annoying to screenreaders https://stackoverflow.com/a/68292918

I guess a quick check with the Linux or Apple tooling might clear that up?

Though I am not sure if the spec Page is screen Reader friendly anyway or if it is a goal :)

zecakeh:
> I found a change that breaks the spec on the [proposals](https://pr1996--matrix-spec-previews.netlify.app/proposals/#work-in-progress) page:
> 
> _&lt;image&gt;_
>
> The links seem to be escaped instead of rendered.

If fixed it in #1997.

zecakeh:
I looked into the empty paragraphs and it doesn't seem like we are doing anything wrong. The only thing that seems to fix it is to use all the `cs-module` shortcodes on a single line, which feels wrong.

So I opened [a discussion on Hugo's forum](https://discourse.gohugo.io/t/hugo-generates-empty-paragraphs-at-beginning-of-or-between-shortcodes/52431).

zecakeh:
The problem of empty paragraphs has been [fixed upstream](https://github.com/gohugoio/hugo/pull/13053) so they won't be generated with the next version of Hugo.

zecakeh:
I also updated the partials that are modified versions of Docsy's to match the latest version.

Johennes:
Could you also update the version reference in the README? It currently claims that 117 still works which is untrue even before this PR.

https://github.com/matrix-org/matrix-spec/blob/main/README.md?plain=1#L64

config.toml
```diff
@@ -88,22 +85,28 @@ sidebar_menu_compact = true
...
+[[params.links.bottom]]
 	name = "YouTube"
 	url = "https://www.youtube.com/channel/UCVFkW-chclhuyYRbmmfwt6w"
 	icon = "fab fa-youtube"
   desc = "Matrix YouTube channel"
-[[params.links.developer]]
+[[params.links.bottom]]
 	name = "Twitter"
 	url = "https://twitter.com/matrixdotorg"
 	icon = "fab fa-twitter"
```
richvdh, line 112: please can you explain why all this is changing? The commit comment says "Update footer partial" which doesn't give many clues

zecakeh: There is a comment above the first `param.links.bottom` and other comments above that explain.

Basically the `footer` partial of Docsy has been split into 3 partials, one for each column (left, right, center). By default in Docsy the "user" links are shown in the left column and the "developer" links are shown in the right column. Until now we changed the whole partial to show "developer" links in the center, instead of on the right. Now, to minimize the changes compared to the upstream partials, I created a new type of links, "bottom", so we can show them in the center by changing only the center partial.

richvdh: Ok, I get it now. It would have been nice if that explanation had made it into a commit comment or something, to help with review/future reference.

richvdh (approved):

PR #1997, opened 2024-11-14
Fix author links in the proposals lists

zecakeh:
Links were broken since the upgrade of the hugo version in #1984.

The broken state can be viewed at [the preview from that PR](https://pr1984--matrix-spec-previews.netlify.app/proposals/#work-in-progress). ~~It cannot be viewed at https://spec.matrix.org/proposals yet because it hasn't been updated since the upgrade (although it is broken on a different way because of the upgrade too).~~ Now visible at https://spec.matrix.org/proposals/#work-in-progress

### Pull Request Checklist

Preview: https://pr1997--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/shortcodes/proposal-tables.html
+++ b/layouts/shortcodes/proposal-tables.html
@@ -60,7 +60,6 @@
             {{ end }}
             {{ $docs_links := delimit $docs_links_list ", " }}
 
-            {{ $authors_list := apply .authors "htmlEscape" "." }}
             {{ $authors_list := apply .authors "printf" "<a href=\"https://github.com/%s\">@%s</a>" "." "." }}
             {{ $authors := delimit $authors_list ", " }}
 
--- a/layouts/shortcodes/proposal-tables.html
+++ b/layouts/shortcodes/proposal-tables.html
@@ -60,6 +60,7 @@
             {{ end }}
             {{ $docs_links := delimit $docs_links_list ", " }}
 
+            {{ $authors_list := apply .authors "htmlEscape" "." }}
             {{ $authors_list := apply .authors "printf" "<a href=\"https://github.com/%s\">@%s</a>" "." "." }}
             {{ $authors := delimit $authors_list ", " }}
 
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/proposal-tables.html
```diff
@@ -60,6 +60,7 @@
             {{ end }}
             {{ $docs_links := delimit $docs_links_list ", " }}
 
+            {{ $authors_list := apply .authors "htmlEscape" "." }}
             {{ $authors_list := apply .authors "printf" "<a href=\"https://github.com/%s\">@%s</a>" "." "." }}
```
richvdh, line 64: I don't really understand this. Isn't the new value immediately overrwritten?

richvdh:
Could you explain how #1984 broke this?

layouts/shortcodes/proposal-tables.html
zecakeh: Oops, it's using the wrong variable here. The `htmlEscape` is a only precaution anyway in case there is a special HTML character in the username of an author, but I'm not even sure GitHub would allow that.

zecakeh:
> Could you explain how #1984 broke this?

I am not sure but there was a Hugo update, so I guess a fix on their part requires the use of `safeHTML` now, which makes sense in a place where HTML is expected and Hugo cannot be sure that the content is really HTML given that it's constructed from a string.

layouts/shortcodes/proposal-tables.html
zecakeh: I remember now why I did that, even if GitHub doesn't allow special characters in the username, it's because the data from the MSC description is not validated.

zecakeh: Ad now I realized that for the URL, we should not use `htmlEscape` but rather a function that converts to a URL-compatible format. Hugo has `urlize` but it also transforms the string instead of just converting special characters.

Maybe we should consider sanitizing to be the job of the script that fetches the proposals?

zecakeh: I removed the call to `htmlEscape`, maybe it can be done separately, to see if we want to sanitize the input of the proposals.

richvdh (approved):
thanks!

PR #1999, opened 2024-11-16
Improve the quality of the rendered diagrams

zecakeh:
The first commit allows to lazy-load the diagrams, which should improve the loading time of the CS API on mobile. In the process it also improves the alt text of the images.

The second commit serves the diagrams as high-resolution WebPs. Encoding a high resolution diagram as WebP gives a file of approximately the same size as the lower resolution PNG. For maximum compatibility we also serve them as a lower resolution WebP and a fallback PNG. WebP was chosen because it is one of the export formats of draw.io/diagrams.net, and it is widely available in modern browsers.

### Pull Request Checklist

Preview: https://pr1999--matrix-spec-previews.netlify.app


richvdh (approved):
LGTM

PR #2000, opened 2024-11-16
Update Inter font and set `display: swap`

zecakeh:
The main goal here was to set `display: swap` for Inter. It means that the browser will render the text of the spec even before Inter is loaded, making the website appear to load faster on mobile.

I noticed that the Google Fonts API has a setting for that, and that it serves the variable font by default, which means that a single file is used for all the font weights, so I updated the script to load the fonts and updated the files. Variable fonts are supported widely in modern browsers.

### Pull Request Checklist

Preview: https://pr2000--matrix-spec-previews.netlify.app


assets/css/fonts/Inter.css
```diff
@@ -2,167 +2,62 @@
...
   font-style: normal;
-  font-weight: 300;
-  src: url(../../fonts/Inter-latin-ext-normal-300.woff2);
-  unicode-range: U+0100-024F, U+0259, U+1E00-1EFF, U+2020, U+20A0-20AB, U+20AD-20CF, U+2113, U+2C60-2C7F, U+A720-A7FF;
+  font-weight: 100 900;
+  font-display: swap;
+  src: local('Inter'), url(../../fonts/Inter-latin-ext-normal.woff2) format('woff2');
+  unicode-range: U+0100-02BA, U+02BD-02C5, U+02C7-02CC, U+02CE-02D7, U+02DD-02FF, U+0304, U+0308, U+0329, U+1D00-1DBF, U+1E00-1E9F, U+1EF2-1EFF, U+2020, U+20A0-20AB, U+20AD-20C0, U+2113, U+2C60-2C7F, U+A720-A7FF;
 }
 /* latin */
```
zecakeh, line 55: Another improvement that we could add to this PR is that we really only need the "latin" character set given that the spec is written in english, so we could serve only this font file and render the CSS inline since it would now be a single block. The only catch is that if we ever start to include characters out of that set, we need to remember to add extra font files, but I don't know if that's likely to happen.

richvdh (approved):
LGTM. Thanks!

PR #1969, opened 2024-10-12
Enforce consistent vertical spacing between paragraphs in endpoint definitions

zecakeh:
Use `p` elements to separate paragraphs instead of `br` and enforce single paragraphs to be wrapped in `p` for consistency. Finally, remove the bottom margin of the last paragraph in a table cell.

Examples of previously affected areas:

- `GET /_matrix/client/versions`:

  Unnecessary space above "Changed in"

  [Before:](https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientversions)

  Capture d’écran du 2024-10-12 18-12-49

  [After:](https://pr1969--matrix-spec-previews.netlify.app/client-server-api/#get_matrixclientversions)

  image

- `GET /_matrix/client/v3/rooms/{roomId}/messages`'s query parameters:

  Inconsistent space above "Changed in": when the description is made of a single paragraph, the "Changed in" looks like a second paragraph with a regular spacing, but in this case, when the description is made of several paragraphs, the spacing is twice as big.
  
  [Before:](https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientv3roomsroomidmessages)

  Capture d’écran du 2024-10-12 18-19-01

  [After:](https://pr1969--matrix-spec-previews.netlify.app/client-server-api/#get_matrixclientv3roomsroomidmessages)

  image

- `GET /_matrix/client/v3/rooms/{roomId}/messages`'s 200 response:

  "Required:" is not inline: when the description is made of a single paragraph, "Required:" looks like it is part of the paragraph, but in this case, when the description is made of several paragraphs, it is in its own paragraph.
  
  [Before:](https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientv3roomsroomidmessages)

  Capture d’écran du 2024-10-12 18-24-47

  [After:](https://pr1969--matrix-spec-previews.netlify.app/client-server-api/#get_matrixclientv3roomsroomidmessages)

  image

- `GET /_matrix/client/v3/rooms/{roomId}/messages`'s 200 response:

  Inconsistent spacing at the bottom of the table cells. When the description is made of a single paragraph, the padding looks the same at the bottom as at the top, but in this case, when the description is made of several paragraphs, the padding at the bottom is twice as big. This also affects cells with "Added in" or "Changed in" paragraphs.
  
  [Before:](https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientv3roomsroomidmessages)

  Capture d’écran du 2024-10-12 18-30-30

[...]

PR #2005, opened 2024-11-19
Fix regression introduced when inlining "Required" with description

zecakeh:
This was a regression in the rendered spec introduced in #1969.

If the description is not set in the object definition, Hugo generates a weird string after "Required": `%!s(<nil>)`.

To avoid that, we default the description to an empty string when it is not set.

Example, in the `m.text` definition:

[Before](https://spec.matrix.org/unstable/client-server-api/#mtext):

image

### Pull Request Checklist


richvdh (approved):
thanks

PR #2006, opened 2024-11-19
Fix `changed-in` partial when used with multiple paragraphs

zecakeh:
When used with a text that includes multiple paragraphs, the partial created invalid HTML by nesting `<p>` elements. It also changed the rendering by making "Changed in vX.XX:" a separate paragraph, when it is inline with a single paragraph.

To change that we do as with "Required" and prepend "Changed in vX.XX:" to the text before it is rendered, making it inline with the first paragraph.

Example: in the `POST /_matrix/client/v3/join/{roomIdOrAlias}`'s request definition:

[Before](https://spec.matrix.org/unstable/client-server-api/#post_matrixclientv3joinroomidoralias):

image

[After](https://pr2006--matrix-spec-previews.netlify.app/client-server-api/#post_matrixclientv3joinroomidoralias):

image

### Pull Request Checklist

Preview: https://pr2006--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2006.clarification
+++ b/changelogs/internal/newsfragments/2006.clarification
@@ -1,0 +1,1 @@
+Fix `changed-in` partial when used with multiple paragraphs.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
lgtm, thanks!

PR #2007, opened 2024-11-20
Restore heading self links

zecakeh:
It was a change I did not notice when upgrading Docsy from 0.8.0 to 0.11.0 in #1996. Docsy changed the way heading self links are generated: they used to be rendered with JS and now they use a Hugo render hook.

This means two things:
- We need to enable them explicitly for headings rendered by Hugo by overriding the `_default/_markup/render-heading.html` template.
- We need to add the self heading ourselves to headings that are not rendered by Hugo, i.e. HTML headings that we create ourselves.

Note that now the symbol for the self links is `#` by default instead of a chain link icon. We can change it if preferred.

image

### Pull Request Checklist

Preview: https://pr2007--matrix-spec-previews.netlify.app


layouts/_default/_markup/render-heading.html
```diff
@@ -0,0 +1 @@
+{{ template "_default/_markup/td-render-heading.html" . }}
```
richvdh, line 1: could you give this a comment explaining what it does? AIUI this is a hugo rendering hook?

Links to https://gohugo.io/render-hooks/headings/ and https://www.docsy.dev/docs/adding-content/navigation/#heading-self-links might be useful.

layouts/_default/_markup/td-heading-self-link.html
```diff
@@ -0,0 +1,11 @@
+{{- /*
+
+    A copy of the private "_default/_markup/_td-heading-self-link.html" template
+    in Docsy's "_default/_markup/_td-render-heading.html" partial to be able to
```
richvdh, line 4:
```suggestion
    in Docsy's "_default/_markup/td-render-heading.html" template, so that we can
```

I don't think it's a partial per se?

layouts/_default/_markup/td-heading-self-link.html
richvdh, line 1: is it deliberate that this is called `td-heading-self-link.html` rather than `_td-heading-self-link.html` with the underscore?

layouts/_default/_markup/td-heading-self-link.html
```diff
@@ -0,0 +1,11 @@
...
+    reuse it when the title has custom markup.
```
richvdh, line 5:
```suggestion
    reuse it when the heading has custom markup.
```

richvdh (changes requested):

layouts/_default/_markup/td-heading-self-link.html
zecakeh: Yes, the behavior when two templates have the same name is not documented (but [Hugo docs say](https://gohugo.io/templates/partial/#inline-partials) that an inline partial must not have the same name as another partial) so I preferred to avoid any problems.

zecakeh: Ah I just saw in Docsy's docs that we can just reuse the inline template. The thing though is that ours is simplified: theirs takes a dict with an `Anchor` key, ours just takes a string.

layouts/_default/_markup/td-heading-self-link.html
zecakeh: Changed

layouts/_default/_markup/td-heading-self-link.html
zecakeh: Indeed, applied.

layouts/_default/_markup/render-heading.html
zecakeh: Comment added.

PR #2008, opened 2024-11-20
Optimize generated CSS by removing unused selectors

zecakeh:
This was done thanks to [this Hugo guide](https://gohugo.io/hugo-pipes/postprocess/#css-purging-with-postcss).

The way this works is that Hugo generates stats about the HTML elements, IDs and classes that can be found in the website, and we post-process the generated CSS with [PurgeCSS](https://github.com/FullHuman/purgecss) that uses those stats to remove unused selectors.

The generated file's size goes down from 403.8 Ko to 38.1 Ko (less than 10%!). Gzipped it goes down from 64.8 Ko to 8.7 Ko (less than 15%!). This huge gain is probably explained because Docsy includes the whole CSS from Bootstrap instead of only the parts that we need.

More than the gain in the download time of the file, we should gain performance in the parsing of the CSS.

### Pull Request Checklist

Preview: https://pr2008--matrix-spec-previews.netlify.app


layouts/partials/head-css.html
```diff
@@ -0,0 +1,37 @@
+{{- /*
+
+    A modified version of the head-css.html partial of Docsy, that allows
+    post-processing of the generated CSS to remove unused selectors.
```
richvdh, line 4:
```suggestion
    A modified version of the head-css.html partial of Docsy, that adds a call to
    `resources.PostProcess`, allowing post-processing of the generated CSS to
    remove unused selectors.
```

richvdh (approved):
LGTM

PR #2010, opened 2024-11-20
Clarify that lazy_load_members in timeline isn't respected in /sync

bradtgmurray:
This one caught me out today, clarify that sync only uses `lazy_load_members` from `state`.

https://github.com/element-hq/synapse/blob/develop/synapse/api/filtering.py#L240-L241

### Pull Request Checklist

Signed-off-by: Brad Murray <brad@beeper.com>

Preview: https://pr2010--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/room_event_filter.yaml
+++ b/data/api/client-server/definitions/room_event_filter.yaml
@@ -28,7 +28,8 @@ allOf:
       description: |-
         If `true`, enables lazy-loading of membership events. See
         [Lazy-loading room members](/client-server-api/#lazy-loading-room-members)
-        for more information. Defaults to `false`.
+        for more information. Defaults to `false`. Only applies to the `/messages`
+        endpoint, for `/sync` this should be specified in `StateFilter` instead.
     include_redundant_members:
       type: boolean
       description: |-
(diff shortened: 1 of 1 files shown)


data/api/client-server/definitions/room_event_filter.yaml
```diff
@@ -28,15 +28,17 @@ allOf:
       description: |-
         If `true`, enables lazy-loading of membership events. See
         [Lazy-loading room members](/client-server-api/#lazy-loading-room-members)
-        for more information. Defaults to `false`.
+        for more information. Defaults to `false`. Only applies to the `/messages`
```
zecakeh, line 31: It should also apply to the `/context` endpoint, as listed in the "lazy-loading room members" section. I believe it would be better to only list exceptions here.

turt2live:
Are we sure this isn't a synapse bug?

richvdh (changes requested):
> Are we sure this isn't a synapse bug?

It seems to be part of the spec: https://spec.matrix.org/v1.12/client-server-api/#get_matrixclientv3sync says:

> Lazy-loading members is only supported on a StateFilter for this endpoint

That said, I'm not convinced this is the clearest way to fix that confusion. For one thing, I don't know why we treat `StateFilter` and `RoomEventFilter` as different types, when they have *exactly the same fields*: I have opened https://github.com/matrix-org/matrix-spec/pull/2015 to fix this.

I think probably we want words like: 

> Note that `/sync` only supports lazy-loading members on the `state` part of a `RoomFilter`.

Also, please use hyperlinks when referring to endpoints, other type definitions, etc.

richvdh:
@bradtgmurray any interest in continuing work on this? The effort to clarify the spec is much appreciated but I don't think it's quite ready to land as it is.

richvdh:
Rather than keep this open indefinitely, I'm going to go ahead and close it. @bradtgmurray: if you have any interest in picking this up again, let us know and we'll reopen it!

Issue #270, opened 2018-03-05
Appservices should be receiving EDUs (read receipts, typing notifications, etc)


turt2live:
I thought there was already an issue for this, but after reading the entire issue list I couldn't find it.

richvdh:
https://github.com/matrix-org/matrix-spec-proposals/pull/2409 is a proposal to fix this.

PR #2018, opened 2024-11-28
Allow sending ephemeral data to application services

zecakeh:
As per MSC2409.

Fixes: #270 

### Pull Request Checklist

Preview: https://pr2018--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/application_service/newsfragments/2018.feature
+++ b/changelogs/application_service/newsfragments/2018.feature
@@ -1,0 +1,1 @@
+Allow to send EDUs to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
--- a/changelogs/application_service/newsfragments/2018.feature
+++ b/changelogs/application_service/newsfragments/2018.feature
@@ -1,1 +1,1 @@
-Allow to send EDUs to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
+Allow to send ephemeral data to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
--- a/changelogs/application_service/newsfragments/2018.feature
+++ b/changelogs/application_service/newsfragments/2018.feature
@@ -1,1 +1,1 @@
-Allow to send ephemeral data to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
+Allow sending ephemeral data to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -207,6 +207,33 @@ processed the events.
 
 {{% http-api spec="application-service" api="transactions" %}}
 
+##### Pushing EDUs
+
+{{% added-in v="1.13" %}}
+
+If the `receive_ephemeral` settings is enabled in the [registration](#registration)
+file, homeservers MUST send all EDUs that are relevant to the application
+service via the transaction API, using the `ephemeral` property of the request's
+body. This property is an array that is effectively a combination of the
+`presence` and `ephemeral` sections of the client-server [`/sync`](/client-server-api/#get_matrixclientv3sync)
+API. It means that there are only three event types that can currently occur:
+[`m.presence`](/client-server-api/#mpresence), [`m.typing`](/client-server-api/#mtyping),
+and [`m.receipt`](/client-server-api/#mreceipt).
+
+EDUs that can be associated to a particular room (i.e. `m.typing` or `m.receipt`)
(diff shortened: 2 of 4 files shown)


content/application-service-api.md
```diff
@@ -207,6 +207,33 @@ processed the events.
 
 {{% http-api spec="application-service" api="transactions" %}}
 
+##### Pushing EDUs
```
richvdh, line 210: I know that the MSC used the term "EDU", but I think it's somewhat misleading. In particular: we do *not* forward `m.direct_to_device` EDUs to ASes (not least because such EDUs contain multiple to-device messages, possibly for other users).

Also: presence/receipts/typing from users on the same server don't arrive via EDUs, so the current wording would imply that the AS would only receive updates from users on other servers, which I don't believe was the intention of the MSC.

Also also: the event format used by the C-S API (and hence, now, by the A-S API) is quite different from that used by EDUs in the S-S API.

Could we, instead of saying "EDU", say "ephemeral data" ?

content/application-service-api.md
```diff
@@ -207,6 +207,33 @@ processed the events.
...
+
+{{% added-in v="1.13" %}}
+
+If the `receive_ephemeral` settings is enabled in the [registration](#registration)
+file, homeservers MUST send all EDUs that are relevant to the application
```
richvdh, line 215: not all EDUs; device list updates and to-device messages are excluded.

```suggestion
file, homeservers MUST send ephemeral data that is relevant to the application
```

content/application-service-api.md
```diff
@@ -207,6 +207,33 @@ processed the events.
...
+service via the transaction API, using the `ephemeral` property of the request's
+body. This property is an array that is effectively a combination of the
+`presence` and `ephemeral` sections of the client-server [`/sync`](/client-server-api/#get_matrixclientv3sync)
+API. It means that there are only three event types that can currently occur:
+[`m.presence`](/client-server-api/#mpresence), [`m.typing`](/client-server-api/#mtyping),
+and [`m.receipt`](/client-server-api/#mreceipt).
```
richvdh, line 221:
```suggestion
API. 

There are currently three event types that can be delivered to an application service:
[`m.presence`](/client-server-api/#mpresence), [`m.typing`](/client-server-api/#mtyping),
and [`m.receipt`](/client-server-api/#mreceipt).
```

content/application-service-api.md
```diff
@@ -207,6 +207,33 @@ processed the events.
...
+
+EDUs that can be associated to a particular room (i.e. `m.typing` or `m.receipt`)
+MUST also include a `room_id` property to identify the room that they were sent
```
richvdh, line 224:
```suggestion
MUST use the same format as the client-server API, with the addition of a `room_id` property at the top level to identify the room that they were sent
```

content/application-service-api.md
```diff
@@ -207,6 +207,33 @@ processed the events.
...
+regular events, meaning that the application service must have registered
+interest in the room itself, or in a user that is in the room.
+[Private read receipts](/client-server-api/#private-read-receipts) should only
+be sent for users matching one of the application service's namespaces. Normal
+read receipts and threaded read receipts are always sent.
+
+EDUs which are not associated with a particular room (i.e. `m.presence`), should
+be sent to the application service if the EDU would apply contextually. For
+example, a presence update for a user an application service shares a room with,
+or matching one of the application service's namespaces.
```
richvdh, line 235: It might be easier to describe each of the three event types separately, with a bullet list, rather than trying to describe the behaviour via generic rules.

richvdh (changes requested):

changelogs/application_service/newsfragments/2018.feature
```diff
@@ -0,0 +1 @@
+Allow to send ephemeral data to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
```
richvdh, line 1:
```suggestion
Allow sending ephemeral data to application services, as per [MSC2409](https://github.com/matrix-org/matrix-spec-proposals/pull/2409).
```

data/api/application-service/transactions.yaml
```diff
@@ -61,6 +70,21 @@ paths:
...
                   items:
                     $ref: ../client-server/definitions/client_event.yaml
+                ephemeral:
+                  type: array
+                  x-addedInMatrixVersion: "1.13"
+                  description: |-
+                    A list of ephemeral data, if the `receive_ephemeral` setting was enabled in the
+                    [registration](/application-service-api/#registration) file.
+
+                    There are only three event types that can currently occur: `m.presence`,
```
richvdh, line 80:
```suggestion
                    There are only three event types that can currently occur in this list: `m.presence`,
```

richvdh (approved):
LGTM, thanks!

PR #2016, opened 2024-11-26
Add hyperlinks to endpoints

richvdh:
... because there's nothing more infuriating than someone talking about an endpoint, then having to search through 150 pages of spec for it.

Preview: https://pr2016--matrix-spec-previews.netlify.app

PR #2014, opened 2024-11-26
Specify account suspension

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3823

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

~~### 🚨 DO NOT MERGE UNTIL MSC3823 FINISHES FCP SUCCESSFULLY~~

~~As of writing, the MSC only [just entered FCP](https://github.com/matrix-org/matrix-spec-proposals/pull/3823#issuecomment-2501482778), which must conclude successfully before this can be merged. Early review is appreciated to facilitate the fastest possible process though.~~

It landed

----

Preview: https://pr2014--matrix-spec-previews.netlify.app

PR #2026, opened 2024-12-09
Move standard error schema to appendices and share between all APIs

olivia-fl:
I asked about this on matrix, and tulir [confirmed][1] that the error schema was intended to be shared between all APIs. I would like the shared schema to be explicit in the spec text so that ruma can move to using the same error type for client-server and server-server endpoints. This would simplify error handling in server implementations.

I'm unsure what the correct thing to do with the error codes that were named in the identity service API but not the client-server API. `M_MISSING_PARAMS` (distinct from `M_MISSING_PARAM` from C2S) is particularly weird. I checked that these are actually used in sydent, and not used in synapse, so I just left them in the identity service API spec.

[1]: https://matrix.to/#/#matrix-spec-discussion:neko.dev/$Y2yTCeR_AeW6g_4jViFbx4gTE_AwF0RN7yrHJ25F5Q8

### Pull Request Checklist

Preview: https://pr2026--matrix-spec-previews.netlify.app


richvdh:
@olivia-fl apologies, this has conflicted with #2016. Could you merge latest main into it?

content/appendices.md
```diff
@@ -4,6 +4,179 @@ weight: 70
...
+hosting environment may reach a resource limit if it starts using too
+much memory or disk space. The error MUST have an `admin_contact` field
+to provide the user receiving the error a place to reach out to.
+Typically, this error will appear on routes which attempt to modify
+state (e.g.: sending messages, account data, etc) and not routes which
+only read state (e.g.: `/sync`, get account data, etc).
+
+`M_CANNOT_LEAVE_SERVER_NOTICE_ROOM`
+The user is unable to reject an invite to join the server notices room.
+See the [Server Notices](#server-notices) module for more information.
```
richvdh, line 178: It's far from obvious to me that all of these error codes belong outside the C-S API spec. Some of them do seem generic, but anything talking about access tokens is probably  irrelevant outside the C-S API.

I'll also note that MSC3823 (which is about to land, via https://github.com/matrix-org/matrix-spec/pull/2014) explicitly specifies a new error code which can only be returned by the C-S API.

I'm happy for the schema definition to move, but I think we should be more conservative before we specify that any S-S API can return `M_USER_LOCKED` etc.

olivia-fl: That seems reasonable. I can go through this list and try to determine which ones are C-S-specific, and then move them to the C-S docs (like what I did with the identity service error codes).

I think that at least stuff like `M_UNKNOWN` should be shared.

olivia-fl: New version has all the error codes that are only used in C2S in the C2S docs. As mentioned in the commit message, I was only able to determine this by reading the synapse source code rather than by looking at the existing spec text, because most endpoints are missing error codes. Let me know if you think the resulting partition is reasonable.

olivia-fl:
Rebased on `main`, but found a duplicated error code introduced in #1944. I've removed that in af6e8a2f24af0fe2e28a1d3b081d0363ce15424b, before the change to move the error codes to the appendix. I also decided to move the rate limiting section to the appendix, because it is referenced by the common `M_LIMIT_EXCEEDED` error, and because we have "Rate-Limited: [Yes/No]" on non-C2S endpoints.

content/appendices.md
```diff
@@ -4,6 +4,136 @@ weight: 70
...
+[`/client/user/{userId}/account_data/{type}`](/client-server-api/#get_matrixclientv3useruseridaccount_datatype), etc).
+
+`M_UNSUPPORTED_ROOM_VERSION`
+The client's request to create a room used a room version that the
+server does not support.
+
+`M_INCOMPATIBLE_ROOM_VERSION`
+The client attempted to join a room that has a version the server does
+not support. Inspect the `room_version` property of the error response
+for the room's version.
```
richvdh, line 112: I don't think this section has any place here. If it is true that they are specific to certain endpoints, then, as the TODO says, they should be documented at those endpoints, rather than as part of the "Standard Error response" list. I'd suggest moving most of them back to the C-S API.

(I'm aware that a couple of them are used in the S-S API, such as `M_UNSUPPORTED_ROOM_VERSION`, but let's solve that another time.)

There's a lot of overlap between these codes and the common codes. (eg, what's the difference between `M_UNAUTHORIZED` and `M_FORBIDDEN`? Or `M_INVALID_PARAM` and `M_BAD_JSON`?)  Again, let's solve it another time, but moving it to the appendices (and thus somehow declaring them valid for all 5 APIs) seems a retrograde step to me.

content/push-gateway-api.md
```diff
@@ -39,6 +39,11 @@ notification provider (e.g. APNS, GCM).
 
 ## API standards
 
+### Standard error response
+
+All push gateway API endpoints MUST return error responses conforming to the
+[standard error response](/appendices#standard-error-response) schema.
```
richvdh, line 45: I'm unconvinced this is true; Synapse does not require the push gateway to return standard error responses, and I don't think Sygnal does so.

If we want to change this, I think it requires an MSC.

content/application-service-api.md
```diff
@@ -12,6 +12,16 @@ Application Service API (AS API) defines a standard API to allow such
 extensible functionality to be implemented irrespective of the
 underlying homeserver implementation.
 
+## API standards
+
+### Standard error response
+
+All homeserver -> application service API endpoints MUST return error responses
+conforming to the [standard error response](/appendices#standard-error-response)
+schema. Similarly, all application service client-server API extension
```
richvdh, line 21: As with the push gateway API, I think we need to think harder about whether it applies to the AS API. Are application services really required to return standard error responses? I'm far from convinced that is something we can just mandate today without an MSC.

richvdh (changes requested):
Thanks for this. Unfortunately it's one of those things that has years of cruft, and cleaning it up isn't as easy as one might hope. A few comments here.

content/appendices.md
olivia-fl: Sure, I can move this back to C-S. I don't think documenting these _only_ on the affected endpoints would necessarily be an improvement, because the definition would then be duplicated in many places (and likely go out of sync over time).

tulir:
Sorry about the miscommunication here. The error schema is definitely meant to be shared and moving that to appendices doesn't require an MSC. It's already shared in practice: there are copies of the schema in the S-S, AS and IS specs. Even the push gateway spec references `M_UNRECOGNIZED` without actually saying how to return it 🙈

However, mandating all endpoints return standard errors does require an MSC, because the current AS and push gateway specs don't have such requirements. The other error codes section shouldn't really exist at all, so probably best to leave it alone in the C-S spec until someone gets around to cleaning it up properly.

olivia-fl:
The whole synapse pro thing has really killed my motivation to work on most matrix stuff at this point. I'm gonna close this and maybe I'll take another shot at it later if things are looking better.

ara4n:
(I'd have thought Synapse Pro would drive alternative FOSS homeserver implementions, tbh!)

Issue #2030, opened 2024-12-13
Account data section needs cleanup


richvdh:
**Link to problem area**:

https://spec.matrix.org/v1.12/client-server-api/#client-config

**Issue**

* For some reason, it is called "Client config" in the spec, when literally nobody calls it that
* It refers to account data as "events", when they really really aren't that. (They might appear as event-like things when sent down `/sync` or `/events`, but that's something else.) Related: https://github.com/matrix-org/matrix-spec/issues/897
* Lots of bits of the spec refer to "account data" but don't hyperlink back to the spec section on account data
* [Room tags](https://spec.matrix.org/v1.12/client-server-api/#room-tagging) are built on top of account data, but appear before account data in the spec, so that feels backwards.
* The "client config" (i.e. account data) section has specific information about `m.tag` ("`m.tag` events appearing in `/events` will have a `room_id` with the room the tags are for"), which seems like it should be common to all room-based account data.

richvdh:
I've labelled this as "good first issue", but if you'd like to pick this up as the first issue you work on, I suggest starting by fixing *one* of the things in the above list, rather than trying to solve them all at once.

PR #2032, opened 2024-12-14
Rename Client Config to Account Data

yoyodragon:
Part of issue #2030.

### Pull Request Checklist

Preview: https://pr2032--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2032.clarification
+++ b/changelogs/client_server/newsfragments/2032.clarification
@@ -1,0 +1,1 @@
+Rename Client Config to Account Data for issue #2030
--- a/changelogs/client_server/newsfragments/2032.clarification
+++ b/changelogs/client_server/newsfragments/2032.clarification
@@ -1,1 +1,1 @@
-Rename Client Config to Account Data for issue #2030
+Rename "Client Config" section to "Account Data".
--- a/changelogs/legacy/client_server.rst
+++ b/changelogs/legacy/client_server.rst
@@ -557,7 +557,7 @@ Since the draft stage, the following major changes have been made:
   - Invites based on third-party identifiers
   - Room tagging
   - Guest access
-  - Client config
+  - Account Data
 - The following APIs were added:
   - ``/sync``
   - ``/publicRooms``
--- a/changelogs/legacy/client_server.rst
+++ b/changelogs/legacy/client_server.rst
@@ -557,7 +557,7 @@ Since the draft stage, the following major changes have been made:
   - Invites based on third-party identifiers
   - Room tagging
   - Guest access
-  - Account Data
+  - Client config
 - The following APIs were added:
   - ``/sync``
   - ``/publicRooms``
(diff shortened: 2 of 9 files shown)


richvdh (changes requested):
I assume this is related to https://github.com/matrix-org/matrix-spec/issues/2030. 

Thank you for the contribution. It's a fair start, but - even though I said "don't try to solve everything at once" -  I think this is the opposite extreme. Just changing the list of modules, without actually changing the name of the section itself, is going to cause confusion. Perhaps you could update your PR so that you're also changing the name of the section itself.

A couple of other points to note:
 * It's helpful to give a bit of description in the PR title about what the PR is trying to achieve. "Update _index.md" doesn't really tell anyone anything.
 * You've ticked "Pull request includes a sign off", but your pull request does not, in fact, include a sign off.

yoyodragon:
Thank you for the suggestions, I have tried implementing them to the best of my knowledge.
I am new to open source and am not completely aware of the rules and protocols followed, I'm sorry for the inconvenience.Please let me know if I have to make a changelog entry too, I couldn't figure out where to do it.
Thank you.

changelogs/legacy/client_server.rst
```diff
@@ -557,7 +557,7 @@ Since the draft stage, the following major changes have been made:
   - Invites based on third-party identifiers
   - Room tagging
   - Guest access
-  - Client config
+  - Account Data
```
zecakeh, line 560: You should not edit an old changelog entry.

data/api/client-server/keys.yaml
```diff
@@ -15,7 +15,7 @@
 # limitations under the License.
 openapi: 3.1.0
 info:
-  title: Matrix Client-Server Client Config API
+  title: Matrix Client-Server Account Data API
```
zecakeh, line 18: I don't think this is quite right (but it was already wrong before). Looking at the spec it looks like it should be "Key Management API" for this file.

content/client-server-api/_index.md
```diff
@@ -2927,7 +2927,7 @@ that profile.
 | [Typing Notifications](#typing-notifications)              | Required  | Required | Required | Required | Optional |
 | [User and Room Mentions](#user-and-room-mentions)          | Required  | Required | Required | Optional | Optional |
 | [Voice over IP](#voice-over-ip)                            | Required  | Required | Required | Optional | Optional |
-| [Client Config](#client-config)                            | Optional  | Optional | Optional | Optional | Optional |
+| [Account Data](#client-config)                            | Optional  | Optional | Optional | Optional | Optional |
```
zecakeh, line 2930: You also need to apply changes to the URL fragment. `#client-config` should have been changed to `#account-data`, here and maybe in other parts of the spec.

zecakeh (commented):
> Please let me know if I have to make a changelog entry too, I couldn't figure out where to do it.

Yes, every PR must have a changelog entry. Are the instructions not clear in CONTRIBUTING.rst?

Also this qualifies as "other changes" as far as I can tell.

changelogs/client_server/newsfragments/2032.clarification
```diff
@@ -0,0 +1 @@
+Rename Client Config to Account Data for issue #2030
```
richvdh, line 1:
```suggestion
Rename "Client Config" section to "Account Data".
```

richvdh: (We don't need the issue number in the changelog.)

PR #2033, opened 2024-12-15
Generate changelog release info with Hugo rather than when generating changelog

zecakeh:
Since all the data is available in the frontmatter, we can just use Hugo to generate the `release-info` table, it allows to have more control over it, notably to:

- Show a link to the checklist on the unstable changelog.
- Generate only markdown in the checklist output format.

The second commit helps to detect the new expected date format in the `generate-changelog.sh` script.

### Pull Request Checklist

Preview: https://pr2033--matrix-spec-previews.netlify.app


layouts/docs/changelog.html
```diff
@@ -0,0 +1,46 @@
+{{- /*
+
+    Template to render a page with a `changelog` layout or the `changelog`
+    section page. This conflation seems to be a limitation of Hugo currently, it
```
zecakeh, line 4: Note: this conflation could be avoided by using a different name for the layout, but on the other end I don't think it hurts that much and avoids to have a shortcode just for redirecting.

scripts/generate-changelog.sh
```diff
@@ -1,31 +1,47 @@
...
     TITLE="Changes since last release"
     LINKTITLE="Unstable"
     FILENAME="unstable.md"
+    DATE=$(date -Iseconds)
 else
     TITLE="$VERSION Changelog"
     LINKTITLE="$VERSION"
     FILENAME="$VERSION.md"
+
+    if [ -z "$DATE" ]; then
```
zecakeh, line 26: Note: We could actually always get the date from the system instead of providing it as a parameter of the script, if we assume that the changelog will always be created on the date of the release.

Even if that's not the case, the date could be edited manually in the changelog page after it is generated.

layouts/docs/changelog.html
```diff
@@ -0,0 +1,46 @@
...
+        {{ if ne $version "unstable" -}}
+            {{ $commitLink := printf "https://github.com/matrix-org/matrix-spec/tree/%s" $version -}}
+        <tr><th>Git commit</th><td><a href="{{ $commitLink }}">{{ $commitLink }}</a></td>
+        <tr><th>Release date</th><td>{{ .Date | time.Format ":date_long" }}</td>
+        {{ end -}}
+        {{ $checklist := .OutputFormats.Get "checklist" -}}
+        <tr><th>Checklist</th><td><a href="{{ $checklist.RelPermalink }}">checklist.md</a></td>
+    </table>
+
+	{{ .Content }}
```
anoadragon453, line 43: A couple tabs seem to have slipped in here.

scripts/generate-changelog.sh
anoadragon453: Speaking as a maintainer of Synapse, [our release script](https://github.com/element-hq/synapse/blob/develop/scripts-dev/release.py) uses the current date and it's perfectly serviceable. It'd be inconvenient (and prone to human error) to have to type it in manually each time.

So +1 to automating that.

anoadragon453 (changes requested):
A few small things, but otherwise LGTM. Nice set of cleanups!

scripts/generate-changelog.sh
zecakeh: Done

layouts/docs/changelog.checklist.md
```diff
@@ -0,0 +1,34 @@
+{{- /*
+
+    Template to render a page with a `changelog` layout as a markdown checklist.
+
+    This transforms the markdown source of the changelog to change list items to
+    checklist items.
+
+    If it is not the unstable changelog, this also adds a table at the top of
+    the page with information about the Matrix spec release:
```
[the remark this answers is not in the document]
zecakeh: Applied

layouts/docs/changelog.html
zecakeh: Removed

anoadragon453 (approved):
Thanks for the making the changes. This now LGTM!

PR #2041, opened 2024-12-19
Update path to config in release steps

turt2live:
Preview: https://pr2041--matrix-spec-previews.netlify.app


anoadragon453 (approved):
FYI The config file rename was made in https://github.com/matrix-org/matrix-spec/pull/2008.

PR #2042, opened 2024-12-22
Remove unused `release_date` from Hugo config

zecakeh:
It seems unnecessary because it is not used anywhere. Also the date format used in the changelog would change in #2033 so it's unclear if the formats should still match if we keep this.

The second commit removes an erroneous sentence since we are in the area. The version is actually updated manually during the release (as can be seen in `/meta/releasing.md`), not by CI.

### Pull Request Checklist

Preview: https://pr2042--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2042.clarification
+++ b/changelogs/internal/newsfragments/2042.clarification
@@ -1,0 +1,1 @@
+Remove unused `release_date` from Hugo config.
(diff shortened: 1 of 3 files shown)


anoadragon453 (approved):
This looks good on the surface, but I'd like @turt2live to double-check as he's more familiar with the release process.

turt2live (approved):
Good to know I was populating an unused field 😅 

Thanks for cleaning this up!

PR #2045, opened 2025-01-06
Clarify that v1.0 of Matrix was a symbolic release

### LINKED ISSUE #12: v1.0 support detection is ambiguous
**Link to problem area**: https://spec.matrix.org/v1.2/#legacy-versioning

**Issue**
The above clarifies a support matrix for backwards definition linking as "v1.0 support", however, it does little to define how to actually detect such v1.0 support.

https://github.com/ruma/ruma/issues/852 raises some questions about this;
- Should v1.0 be determined to be supported if any of the above revisions appear in `/versions`?
  - Should it only be if *all* are detected?
  - Should it be only on `r0.6.1` or `r0.6.0`?
  - Should it be on *any* `r*` version?
- Should it (additionally) also be determined if a server sends a `v1.0` string in there?

zecakeh:
Explain that it was a release prior to the current global versioning system and that it is not used in the APIs by clients/servers.

Closes #12.

### Pull Request Checklist

Preview: https://pr2045--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2045.clarification
+++ b/changelogs/internal/newsfragments/2045.clarification
@@ -1,1 +1,1 @@
-Clarify that v1.0 of Matrix was a symbolic release prior to the current global versionins system.
+Clarify that v1.0 of Matrix was a symbolic release prior to the current global versioning system.
--- a/changelogs/internal/newsfragments/2045.clarification
+++ b/changelogs/internal/newsfragments/2045.clarification
@@ -1,0 +1,1 @@
+Clarify that v1.0 of Matrix was a symbolic release prior to the current global versionins system.
--- a/changelogs/internal/newsfragments/2045.clarification
+++ b/changelogs/internal/newsfragments/2045.clarification
@@ -1,1 +1,1 @@
-Clarify that v1.0 of Matrix was a symbolic release prior to the current global versioning system.
+Clarify that v1.0 of Matrix was a release prior to the current global versioning system.
--- a/content/_index.md
+++ b/content/_index.md
@@ -504,8 +504,8 @@ For historical reference, the APIs were versioned as `rX.Y.Z` where `X`
 roughly represents a breaking change, `Y` a backwards-compatible change, and
 `Z` a patch or insignificant alteration to the API.
 
-`v1.0` of Matrix was released on June 10th, 2019 with the following API
-versions:
+`v1.0` of Matrix was a symbolic version released on June 10th, 2019 with the
+following minimum API versions:
 
 | API/Specification       | Version |
 |-------------------------|---------|
@@ -516,6 +516,9 @@ versions:
 | Push Gateway API        | r0.1.0  |
 | Room Version            | v5      |
 
+It is not used in this specification and should not be returned by servers in
+the [`GET /_matrix/client/versions`](/client-server-api/#get_matrixclientversions)
+endpoint. The first version using the current global versioning system was `v1.1`.
 
 ## License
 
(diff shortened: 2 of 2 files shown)


changelogs/internal/newsfragments/2045.clarification
```diff
@@ -0,0 +1 @@
+Clarify that v1.0 of Matrix was a symbolic release prior to the current global versionins system.
```
zecakeh, line 1: I am not sure where to add the changelog, it doesn't seem that there is a section for the homepage of the spec.

content/_index.md
```diff
@@ -504,8 +504,8 @@ For historical reference, the APIs were versioned as `rX.Y.Z` where `X`
 roughly represents a breaking change, `Y` a backwards-compatible change, and
 `Z` a patch or insignificant alteration to the API.
 
-`v1.0` of Matrix was released on June 10th, 2019 with the following API
-versions:
+`v1.0` of Matrix was a symbolic version released on June 10th, 2019 with the
```
zecakeh, line 507: Should we link to the matrix.org blog post of the release?

content/_index.md
```diff
@@ -504,8 +504,8 @@ For historical reference, the APIs were versioned as `rX.Y.Z` where `X`
...
+following minimum API versions:
```
richvdh, line 508: I don't really think that "symbolic release" is going to mean much to the unfamiliar reader. Maybe:

```suggestion
The current global versioning system was introduced with `v1.1`. 
[Matrix 1.0][https://matrix.org/blog/2019/06/11/introducing-matrix-1-0-and-the-matrix-org-foundation/]
did not correspond directly to a specification version; instead, it was based on the following
versions for the individual APIs:
```

content/_index.md
```diff
@@ -516,6 +516,9 @@ versions:
 | Push Gateway API        | r0.1.0  |
 | Room Version            | v5      |
 
+It is not used in this specification and should not be returned by servers in
+the [`GET /_matrix/client/versions`](/client-server-api/#get_matrixclientversions)
+endpoint. The first version using the current global versioning system was `v1.1`.
```
richvdh, line 521:
```suggestion
`v1.0` should **not** be returned by servers in
the [`GET /_matrix/client/versions`](/client-server-api/#get_matrixclientversions)
response.
```

content/_index.md
```diff
@@ -516,6 +516,9 @@ versions:
 | Push Gateway API        | r0.1.0  |
 | Room Version            | v5      |
```
richvdh, line 517: Separate problem, but I don't really think that Matrix 1.0 corresponded to room v5: if you look at, say, https://github.com/matrix-org/matrix-spec-proposals/blob/f44483a8cf0f174376647ac5a777ad6145b5f195/specification/index.rst#specification-versions, it's clear that *all* room versions from v1 to v5 were part of Matrix 1.0.

So, maybe while we're here:

```suggestion
| Room Versions           | v1-v5   |
```

cc @turt2live, as the original author of this table

turt2live: at a spec level room versions aren't linear, so we'd define this as `1, 2, 3, 4, 5` (yes, v5 is also wrong - it's just `5`).

the backing history here is that it felt weird to list out the room versions like that, so a slight trick was applied: because all the room versions inherit from each other, even though they don't have to, we can simply point to the "latest" which drags the remainder into scope as well. The deeper rationale is that by listing fewer numbers, we made the table less scary. Since we're well past the 1.0 launch window though, we can probably reintroduce a lot of that complexity.

richvdh:
```suggestion
| Room Versions           | 1, 2, 3, 4, 5 |
```

then

content/_index.md
zecakeh: Applied

content/_index.md
zecakeh: Applied

content/_index.md
zecakeh: Applied

richvdh (approved):

PR #2046, opened 2025-01-06
Clarify formats of string types in account sections

zecakeh:
This should cover sections 4.7 through 4.9 of the CS API.

### Pull Request Checklist

Preview: https://pr2046--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/administrative_contact.yaml
+++ b/data/api/client-server/administrative_contact.yaml
@@ -120,7 +120,6 @@ paths:
                       description: The client secret used in the session with the identity server.
                     id_server:
                       type: string
-                      format: mx-server-name
                       description: The identity server to use.
                     id_access_token:
                       type: string
@@ -280,7 +279,6 @@ paths:
                   description: The client secret used in the session with the identity server.
                 id_server:
                   type: string
-                  format: mx-server-name
                   description: The identity server to use.
                 id_access_token:
                   type: string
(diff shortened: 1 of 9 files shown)


data/api/client-server/administrative_contact.yaml
```diff
@@ -120,6 +120,7 @@ paths:
                       description: The client secret used in the session with the identity server.
                     id_server:
                       type: string
+                      format: mx-server-name
```
richvdh, line 123: are you sure that `id_server` is supposed to be a server name?

AIUI, synapse accepts things with trailing paths here, which is obviously not allowed by a server name.

See also https://github.com/matrix-org/synapse/pull/9802#discussion_r613172343 which has a bit of historical debate about this. It's certainly questionable whether Synapse *should* allow trailing paths here, but changing it probably needs an MSC.

zecakeh: I am not sure of anything, this is what I understood from descriptions like:

> The hostname of the identity server to communicate with. May optionally include a port.

But it is indeed not clear everywhere what this should look like, and looking at the PR you linked it is clear that at least Synapse allows to set a path too.

So I will revert the changes about `id_server`. Should we open an issue that this needs to be clarified?

zecakeh: Alright I reverted the `id_server` formats and opened https://github.com/matrix-org/matrix-spec/issues/2056.

PR #2049, opened 2025-01-06
Fix syntax highlighting and click-to-copy buttons for code blocks

zecakeh:
It turns out that we were purging too much CSS since #2008 which broke their style.

We use a simple but lazy solution that parses the JS scripts and keeps the matching selectors. This increases the CSS file size to 59KB.

Before:

image

After:

image

### Pull Request Checklist

Preview: https://pr2049--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Improvement is clear - thanks for catching and fixing!

PR #2057, opened 2025-01-15
Clarify that new concerns will cancel FCP

HarHarLinks:
idk this is a wild guess about what seems reasonable

### Pull Request Checklist

Signed-off-by: HarHarLinks <2803622+HarHarLinks@users.noreply.github.com>

Preview: https://pr2057--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2057.clarification
+++ b/changelogs/internal/newsfragments/2057.clarification
@@ -1,0 +1,1 @@
+Clarify that new concerns will cancel FCP. Contributed by @HarHarLinks.
--- a/content/proposals.md
+++ b/content/proposals.md
@@ -243,7 +243,8 @@ is as follows:
 -   A concern can be raised by a Spec Core Team member at any time,
     which will block an FCP from beginning. An FCP will only begin when
     75% of the members of the Spec Core Team agree on its outcome, and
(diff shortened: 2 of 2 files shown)


HarHarLinks:
Alternatively, it would seem reasonable that after resolving the concern, FCP can be proposed again.
Seeing as the SCT already agreed on FCP, restarting FCP automatically would be the lighter approach.

richvdh:
I don't think this is correct. IIRC, a concern stops the FCP completing, but doesn't stop the 5-day timer. In other words, if the 5-day timer elapses while the concern is active, the FCP will complete as soon as the concern is resolved.

tulir:
The next bullet point already says concerns block the FCP completing, but don't reset the timer, so I don't think any changes are needed?

turt2live (changes requested):
as others have said, this isn't what is done in practice. Concerns are intended to prevent the FCP from *starting* or *finishing* depending on where it's at, but it doesn't *cancel* the FCP (cancelling means we have to re-acquire *all* checkboxes).

richvdh:
I'm going to close this, as it seems to be incorrect, and the existing doc is clear.

PR #2064, opened 2025-01-24
Clarify which rooms are returned from /hierarchy

### LINKED ISSUE #1184: spec does not say which rooms are returned from `/hierarchy`
According to [MSC2946](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2946-spaces-summary.md#client-server-api), "Any room that is ... potentially joinable (per MSC3173) is returned". This doesn't seem to be reflected at https://spec.matrix.org/v1.3/client-server-api/#get_matrixclientv1roomsroomidhierarchy.

clokep:
Clarifies which rooms are returned from `/hierarchy` to fix #1184.


data/api/client-server/space_hierarchy.yaml
```diff
@@ -88,7 +88,16 @@ paths:
...
+                    description: |-
+                       The rooms for the current page, with the current filters.
+
+                       Rooms which the requesting user can request the [stripped state](#) of are returned, this includes:
+
+                       * Any rooms where the requesting user's [room membership](#room-membership) is `join` or `invite`
+                       * Any rooms the user can join:
+                         * Rooms with [join rules](#mroomjoin_rules) set to `join`, `knock`, or `knock_restricted`
+                         * Rooms with a join rule of [`restricted`](#restricted-rooms) if the user meets one of the specified conditions
+                       * Any rooms with [`m.room.history_visibility`](#room-history-visibility) set to `world_readable`.
```
richvdh, line 100: I'm not actually sure you can request stripped state per se: it tends to be *pushed* via /sync. Further, in some of those states, you get full state, not stripped state. So: I'm not sure this is the clearest way to describe it. 

How about this?

```suggestion
                       The server should return any rooms where at least one of the following conditions is true:
                       
                       * The requesting user is currently a member (their [room membership](#room-membership) is `join`).
                       * The requesting user already has permission to join, i.e. one of the following:
                         * The user's room membership is `invite`.
                         * The room's [join rules](#mroomjoin_rules) are set to `public`.
                         * The room's join rules are set to [`restricted`](#restricted-rooms), provided the user meets one of the specified conditions.
                       * The room is "knockable" (the room's join rules are set to `knock`, or `knock_restricted`, in a room version that supports those settings).
                      * The room's [`m.room.history_visibility`](#room-history-visibility) is set to `world_readable`.
```

data/api/client-server/space_hierarchy.yaml
```diff
@@ -88,7 +88,16 @@ paths:
...
                     type: array
-                    description: The rooms for the current page, with the current filters.
+                    description: |-
+                       The rooms for the current page, with the current filters.
+
+                       Rooms which the requesting user can request the [stripped state](#) of are returned, this includes:
+
+                       * Any rooms where the requesting user's [room membership](#room-membership) is `join` or `invite`
+                       * Any rooms the user can join:
+                         * Rooms with [join rules](#mroomjoin_rules) set to `join`, `knock`, or `knock_restricted`
```
richvdh, line 98: (note join rules cannot be "join")

clokep:
Thanks @richvdh, probably good now?

richvdh (approved):
thanks!

Issue #2062, opened 2025-01-24
Document that incoming invite events will reach a homeserver twice - once over /invite and again over /send


anoadragon453:
**Original title:** Calling `PUT /_matrix/federation/v1/invite/{roomId}/{eventId}` on a remote homeserver is redundant if they're already in the room

---

[`PUT /_matrix/federation/v1/invite/{roomId}/{eventId}`](https://spec.matrix.org/v1.13/server-server-api/#put_matrixfederationv1inviteroomideventid) is useful, and its own endpoint, as homeservers that aren't yet participating in a room need to be directly notified when one of their users has been invited.

However, if a homeserver is already participating in a room, when one of their user is invited they'll receive both:

* The `m.room.member` `invite` event over the [`PUT /_matrix/federation/v1/send/{txnId}`](https://spec.matrix.org/v1.13/server-server-api/#put_matrixfederationv1sendtxnid) endpoint.
* A request to `PUT /_matrix/federation/v1/invite/{roomId}/{eventId}`.

Both of these tell the homeserver the same thing - and in practice developers tend to only implement event handling based on one of these. Typically `/invite`.

This also created subtle implementation bugs such as https://github.com/element-hq/synapse/pull/18075 in Synapse - where Synapse notifies clients that they've been invited based on `/invite` - but doesn't actually consider the user *invited* until the corresponding `m.room.member` event comes in. If the user tries to join the room in this gap - they'll get an error that they haven't been invited.

It's also just unnecessary work - we don't need to notify a remote homeserver about an invite twice.

(credit to @MadLittleMods for originally pointing this out in the Synapse PR linked about.)

richvdh:
IIRC, invite events are supposed to be signed by the *invited* homeserver, before they are sent out over `PUT /_matrix/federation/v1/send/{txnId}`.  `/v1/invite` allows the target homeserver to return that signature.

You could argue that sending it again over `PUT /_matrix/federation/v1/send/{txnId}` to the target HS is redundant, but special-casing it is *more* work.

richvdh:
A clarification sounds very sensible to me

anoadragon453:
I've created a PR with one here: https://github.com/matrix-org/matrix-spec/pull/2067

PR #2067, opened 2025-01-28
Add a note to the federation invite endpoints that invites can be sent twice

anoadragon453:
As this may be non-obvious when implementing behaviour that is triggered by an incoming invite event.

See https://github.com/matrix-org/matrix-spec/issues/2062 for more context. Closes https://github.com/matrix-org/matrix-spec/issues/2062.

Preview: https://pr2067--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/2067.clarification
+++ b/changelogs/server_server/newsfragments/2067.clarification
@@ -1,0 +1,1 @@
+Add a note to the invite endpoints that invites to local users may be received twice over federation if the homeserver is already in the room.
--- a/data/api/server-server/invites-v1.yaml
+++ b/data/api/server-server/invites-v1.yaml
@@ -35,7 +35,7 @@ paths:
 
         Also note that if the remote homeserver is already in the room, it will receive the
         invite event twice; once through this endpoint, and again through a [federation
-        transaction](https://spec.matrix.org/v1.13/server-server-api/#transactions).
+        transaction](/server-server-api/#transactions).
       operationId: sendInviteV1
       security:
         - signedRequest: []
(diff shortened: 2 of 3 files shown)


zecakeh (commented):
Regarding the discussion in #2062, I actually saw this endpoint only as a request from the inviting homeserver to the invited homeserver to sign the event, not that it should have other side effects. But it's true that the description is vague and doesn't actually say what should happen on the invitee's side.

As a side note, it is also not clear in which order the signatures are applied here. Should the event already have the signature from the inviting homeserver? It would make sense that the invited homeserver checks the signature before signing it itself. The text doesn't say anything about it, and the definitions don't include a `signatures` field in the request (or the response for that matter), but the data has an example of a request with the `signatures` field (which appears in #2076).

richvdh:
> Should the event already have the signature from the inviting homeserver?

Synapse's implementation of `PUT /_matrix/federation/v2/invite/{roomId}/{eventId}` seems to require a valid signature on the incoming event, so I think the answer to this is "yes". That doesn't seem to be specced though, and it could do with being.

richvdh:
> I actually saw this endpoint only as a request from the inviting homeserver to the invited homeserver to sign the event, not that it should have other side effects.

It *has* to have a side-effect, at least in the case where the invited homeserver is not already in the room: the server must tell the user they have been invited and allow them to join the room if they wish.

zecakeh:
> > I actually saw this endpoint only as a request from the inviting homeserver to the invited homeserver to sign the event, not that it should have other side effects.
> 
> It _has_ to have a side-effect, at least in the case where the invited homeserver is not already in the room: the server must tell the user they have been invited and allow them to join the room if they wish.

Ah right, sorry for the noise then.

richvdh (changes requested):
@anoadragon453 could you make the requested changes please

richvdh (approved):

PR #2068, opened 2025-01-31
Describe behaviour when the `topic` key is falsey in a `m.room.topic` event.

Half-Shot:
We seem to have [updated this for m.room.name](https://github.com/matrix-org/matrix-spec/pull/1639) some years back but omitted it for topic.

Signed-off-by: Half-Shot <signoff@half-shot.uk>

### Pull Request Checklist

Preview: https://pr2068--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/schema/m.room.topic.yaml
+++ b/data/event-schemas/schema/m.room.topic.yaml
@@ -7,8 +7,8 @@ description: |-
     be suitable for the room name.
     The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
     
-    If a room has an `m.room.topic` event with an absent, null, or empty `topic`
-    field, it should be treated the same as a room with no `m.room.topic` event.
+    If the `topic` property is absent, null, or empty then the topic is unset. In other words,
+    an empty `topic` property effectively resets the room to having no topic.
 properties:
   content:
     properties:
--- a/data/event-schemas/schema/m.room.topic.yaml
+++ b/data/event-schemas/schema/m.room.topic.yaml
@@ -1,7 +1,14 @@
 ---
 allOf:
   - $ref: core-event-schema/state_event.yaml
-description: 'A topic is a short message detailing what is currently being discussed in the room.  It can also be used as a way to display extra information about the room, which may not be suitable for the room name. The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+description: |- 
+    A topic is a short message detailing what is currently being discussed in the room.
+    It can also be used as a way to display extra information about the room, which may not
+    be suitable for the room name.
+    The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+    
+    If a room has an `m.room.topic` event with an absent, null, or empty `topic`
+    field, it should be treated the same as a room with no `m.room.topic` event.
 properties:
   content:
     properties:
--- a/changelogs/client_server/newsfragments/2068.clarification
+++ b/changelogs/client_server/newsfragments/2068.clarification
@@ -1,0 +1,1 @@
+Clarify behaviour when the `topic` key of a `m.room.topic` event is absent, null, or empty.


data/event-schemas/schema/m.room.topic.yaml
```diff
@@ -1,7 +1,14 @@
...
   - $ref: core-event-schema/state_event.yaml
-description: 'A topic is a short message detailing what is currently being discussed in the room.  It can also be used as a way to display extra information about the room, which may not be suitable for the room name. The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+description: |- 
+    A topic is a short message detailing what is currently being discussed in the room.
+    It can also be used as a way to display extra information about the room, which may not
+    be suitable for the room name.
+    The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+    
+    If a room has an `m.room.topic` event with an absent, null, or empty `topic`
+    field, it should be treated the same as a room with no `m.room.topic` event.
```
KitsuneRal, line 11: While it's a great clarification in itself, it would be even greater to further clarify what happens when an event with an absent/null/empty `topic` comes upon a previous event with non-falsey `topic`. I guess the room topic should be reset to the as-if-no-topic state?

Half-Shot: Hm, I suppose I thought that was the implication as with the `m.room.name` description => clients should use the latest state to render the room, and this description tells clients to assume the topic isn't set.

I'm struggling to find words that doesn't sound a bit repetitive?

data/event-schemas/schema/m.room.topic.yaml
```diff
@@ -1,7 +1,14 @@
...
 allOf:
   - $ref: core-event-schema/state_event.yaml
-description: 'A topic is a short message detailing what is currently being discussed in the room.  It can also be used as a way to display extra information about the room, which may not be suitable for the room name. The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+description: |- 
+    A topic is a short message detailing what is currently being discussed in the room.
+    It can also be used as a way to display extra information about the room, which may not
+    be suitable for the room name.
+    The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+    
+    If a room has an `m.room.topic` event with an absent, null, or empty `topic`
```
Half-Shot, line 10: So it should be noted, again as with `m.room.name`, that [Synapse doesn't allow a topic](https://github.com/element-hq/synapse/blob/ebad618bf0f4a7cd8adb5c65d6025d320387b492/synapse/events/validator.py#L213-L215) without a string topic key.

So this present behavior doesn't exist there, but I need to survey the other homeservers.

data/event-schemas/schema/m.room.topic.yaml
Half-Shot: better?

richvdh (approved):
LGTM

data/event-schemas/schema/m.room.topic.yaml
```diff
@@ -1,7 +1,14 @@
...
   - $ref: core-event-schema/state_event.yaml
-description: 'A topic is a short message detailing what is currently being discussed in the room.  It can also be used as a way to display extra information about the room, which may not be suitable for the room name. The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+description: |- 
+    A topic is a short message detailing what is currently being discussed in the room.
+    It can also be used as a way to display extra information about the room, which may not
+    be suitable for the room name.
+    The room topic can also be set when creating a room using `/createRoom` with the `topic` key.'
+    
+    If the `topic` property is absent, null, or empty then the topic is unset. In other words,
+    an empty `topic` property effectively resets the room to having no topic.
```
Johennes, line 11: I'm slightly confused about the absent / null part of this. `topic` is labeled as required. So I think it could at most be an empty string?

richvdh: It could have been redacted. Or received over federation from a server that doesn't enforce the presence of the "required" properties.

Johennes: Oh right. I hadn't thought about redactions.

PR #2070, opened 2025-02-07
Refactor PDU definitions to reduce duplication

zecakeh:
First of all, all PDU definitions were based on the v1 PDU definition with a few fields "overwritten" with `allOf` to change the format with needed. While that works when rendering the spec, this is semantically incorrect because it means that the objects must match both schemas, which is impossible.

So now we make a base with only the common fields, and we add the others as needed by the room version. Note that there is no more "unsigned PDU" definition since it is not used directly, and hashes and signatures are the same across all versions.

### Pull Request Checklist

Preview: https://pr2070--matrix-spec-previews.netlify.app


richvdh (approved):
LGTM. Much clearer. Thanks!

PR #2072, opened 2025-02-20
Clients can choose which history visibility options they offer to users when creating rooms

Johennes:
The current wording is ambiguous because it's not clear whether it means for clients to offer:

1. all history visibility options that the spec defines
2. only those that the client chose to support during room creation

Option 1 doesn't seem reasonable to me. Evidently Element X doesn't allow users to choose between all 4 visibility values when creating rooms. It appears perfectly reasonable for clients to decide that they don't want to support creating rooms with a certain history visibility.

Option 2 seems reasonable to me but I don't think it requires a notice in the spec. Clients are generally free to choose what subset of configuration options they expose in their UI (see e.g. push rules).

Therefore, I think we should drop this paragraph.

### Pull Request Checklist

Preview: https://pr2072--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2072.clarification
+++ b/changelogs/client_server/newsfragments/2072.clarification
@@ -1,0 +1,1 @@
+Clients can choose which history visibility options they offer to users when creating rooms.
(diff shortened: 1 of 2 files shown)


richvdh (approved):
seems fair to me

PR #2076, opened 2025-02-21
Look for examples in all possible locations

### LINKED ISSUE #1157: It is not possible to specify multiple examples of a response payload
**Link to problem area**: https://github.com/matrix-org/matrix-spec/blob/783624d2cd11a90c5ea76c36d165a078a957dc0f/layouts/partials/openapi/render-responses.html#L83-L97

**Issue**
The quoted bit of go template language seems to only look for an example defined under the key `application/json`. This means we can provide at most one example for a response in the text of the spec. Sometimes this is an obstacle, e.g. if the same HTTP status code is used to indicate two different error conditions and we wish to give an example of both.

### LINKED ISSUE #1811: Empty objects and null in registration arrays

zecakeh:
We used to only look for examples in a few (sometimes arbitrary) places, and we didn't support showing several examples in most cases. This is intended to fix this. In the process we try to deduplicate code to make sure that we use the same logic everywhere.

This can be reviewed commit by commit:

1. Deduplicate code to render examples. Since there is some logic necessary for "JSONification", notably decoding `<` and `>` inside examples, it's better to keep it in a single place. We also centralize the support for non-JSON examples by using the `no-highlight` value.
2. Change the `json-schema/resolve-example` partial to return a slice of examples, and to look also at the `examples` key, which is the correct key for examples in JSON Schema.
3. Deduplicate code between requests and responses by identifying what the OpenAPI spec calls "Media Type Objects". In the process we also look in more places, like the `examples` key once again, given that the `example` key is deprecated. And we unify the order of the locations where we look, by using the full examples before we try to generate one from the examples of the fields. This produces only a few changes in the rendered spec (see below).
4. Fix a few SS API examples that are now picked up to be rendered in the spec, that contain a `room_id` field although it is not in the corresponding definition.
5. Only show empty examples if the object is empty. Fixes #1811.

Closes #1157.

### Changes in the rendered spec

The only changes are located in the SS API. Other than a few minor changes in the values of fields (like using `example.org` instead of `matrix.org`) in several examples, the 2 following examples for the `/invite` endpoints have significant changes:

**`PUT /_matrix/federation/v1/invite/{roomId}/{eventId}`'s request body example:**

[Before](https://spec.matrix.org/v1.13/server-server-api/#put_matrixfederationv1inviteroomideventid):

```json
{
  "content": {
    "membership": "invite"
  },
  "origin": "matrix.org",
  "origin_server_ts": 1234567890,
  "sender": "@someone:example.org",
  "state_key": "@joe:elsewhere.com",
  "type": "m.room.member",
  "unsigned": {
    "invite_room_state": [
      {
        "content": {
          "name": "Example Room"
        },
        "sender": "@bob:example.org",
        "state_key": "",
        "type": "m.room.name"
      },
      {
        "content": {
          "join_rule": "invite"
        },
        "sender": "@bob:example.org",
        "state_key": "",
        "type": "m.room.join_rules"
      }
    ]
  }
}
```

[After](https://pr2076--matrix-spec-previews.netlify.app/server-server-api/#put_matrixfederationv1inviteroomideventid):

```json
{
  "content": {
    "membership": "invite"
  },
  "origin": "example.org",
  "origin_server_ts": 1549041175876,
  "sender": "@someone:example.org",
  "signatures": {
    "example.com": {
      "ed25519:key_version": "SomeSignatureHere"
    }
  },
  "state_key": "@joe:elsewhere.com",
  "type": "m.room.member"
}
```

**`PUT /_matrix/federation/v2/invite/{roomId}/{eventId}`'s request body example:**

[Before](https://spec.matrix.org/v1.13/server-server-api/#put_matrixfederationv2inviteroomideventid):

```json
{
  "event": {
    "content": {
      "membership": "invite"
    },
    "origin": "matrix.org",
    "origin_server_ts": 1234567890,
    "sender": "@someone:example.org",
    "state_key": "@joe:elsewhere.com",
    "type": "m.room.member"
  },
  "invite_room_state": [
    {
      "content": {
        "name": "Example Room"
      },
      "sender": "@bob:example.org",
      "state_key": "",
      "type": "m.room.name"
    },
    {
      "content": {
        "join_rule": "invite"
      },
      "sender": "@bob:example.org",
      "state_key": "",
      "type": "m.room.join_rules"
    }
  ],
  "room_version": "2"
}
```

[After](https://pr2076--matrix-spec-previews.netlify.app/server-server-api/#put_matrixfederationv2inviteroomideventid):

```json
{
  "event": {
    "content": {
      "membership": "invite"
    },
    "origin": "example.org",
    "origin_server_ts": 1549041175876,
    "sender": "@someone:example.org",
    "signatures": {
      "example.com": {
        "ed25519:key_version": "SomeSignatureHere"
      }
    },
    "state_key": "@joe:elsewhere.com",
    "type": "m.room.member"
  },
  "room_version": "2"
}
```

In both cases, the `invite_room_state` is removed, and `signatures` is added. I believe that the old example is better so we could just remove those examples.

### Pull Request Checklist

Preview: https://pr2076--matrix-spec-previews.netlify.app


layouts/partials/json-schema/resolve-examples.html
```diff
@@ -0,0 +1,66 @@
...
+            {{/*
+                It would be too complex to handle several nested examples,
+                just use the first one.
+            */}}
+            {{ with index $this_property_examples 0 }}
+                {{ $example = merge (dict $key .) $example }}
+            {{ end }}
+        {{ end }}
+
+        {{ if (or $example (not $this_object.properties)) }}
```
richvdh, line 40: It's hard to understand what this is doing, without a comment.

```suggestion
        {{/*
            Add the assembled example to the list if either (a) the example is non-empty, or (b) the object
            itself is meant to be empty (so an empty example is correct).
        */}}
        {{ if (or $example (not $this_object.properties)) }}
```

richvdh (changes requested):
lgtm otherwise

layouts/partials/openapi/render-media-type-objects.html
```diff
@@ -0,0 +1,98 @@
...
+            {{ $type_of = $schema.items.title }}
+        {{ end }}
+<p>Array of <code>{{ $type_of }}</code>.</p>
+    {{ end }}
+
+    {{/*
+        Render object tables for any objects referenced in the response.
+        
+        This will be a no-op for response types which aren't objects or arrays.
+    */}}
```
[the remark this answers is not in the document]
zecakeh: Applied

layouts/partials/json-schema/resolve-examples.html
zecakeh: Applied

richvdh (approved):
thanks

Issue #1700, opened 2023-12-26
Should we require disclosure of potential Conflict of Interest on MSCs?


ara4n:
**Suggestion**

A recurring theme is confusion/concern from the community over the motives of MSCs, and which ones gets prioritised - especially for MSCs from folks on the Matrix Core Team who are employed to work on Matrix by Element.  For that matter, it's also useful for the SCT and community to have more context on a given MSC (e.g. is this driven by Gematik's TI-Messenger work? is this MTRNord proposing something as an independent contributor, or for a Nordeck project?) etc.

So, shall we put a "Disclosures" section at the bottom of the MSC template and ask folks to self-identify the hats they're wearing when proposing an MSC?  For instance, for me, it would be something like:

#### Disclosures

Disclosure: the author is a Guardian of Matrix, a member of the SCT, is Project Lead of Matrix, and employed by Element. However, this MSC is [one of]:

 * Written as an individual FOSS contributor
 * Written as a Matrix spec core team member (funded to work on core Matrix work by Element)
 * Written as a Matrix core team member (funded to work on core Matrix work by Element)
 * Written as an Element employee (funded by commercial Element work)
 * Written as an Element employee (funded by the Ruritanian Government) (if the end customer is willing to be disclosed)

tcpipuk:
I'm all for this as an ***option***, but not as a mandatory disclosure. People are welcome to transparency, but some of us are developing personally and do not want to disclose other "potential" interests that have nothing to do with our interactions here.

turt2live:
I do agree that the hat someone is wearing while writing the MSC should be disclosed, though I don't think it needs to be a formal, normative, section of the MSC.

The sign-off email address alone is usually pretty telling where an MSC came from - maybe we should elevate its presence. Likely putting a few words into the description is fine. Conflicts are pretty irrelevant after the MSC is merged, so no need to take up reading time. I have similar thoughts about the Dependencies section already in the template, fwiw.

I strongly disagree with the notion that being employed by a Matrix vendor (ie: Element) means you're required to disclose that employment in the MSC, however. If the MSC is written in such a capacity then it should absolutely be disclosed (and the employer should feel compelled to make the disclosure a thing). 

ftr, I do think it should be mandatory for SCT members to both list their employer (if a Matrix vendor or using Matrix) and which hat they've used to write the MSC (if different). This only seems fair, given the position SCT members fill.

[...]

PR #2079, opened 2025-02-21
Make disclosure of conflicts of interest mandatory for spec change proposals

Johennes:
Fixes: #1700

### Pull Request Checklist

Preview: https://pr2079--matrix-spec-previews.netlify.app

### DIFF
--- a/content/proposals.md
+++ b/content/proposals.md
@@ -79,6 +79,19 @@ Please [see
 MSC1779](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1779-open-governance.md)
 for full details of the project's Guiding Principles.
 
+For the purposes of transparency and to underpin the "greater benefit"
+principle, proposals MUST disclose possible conflicts of interest where
+they might reasonably exist. Authors MAY do this by filling out the
+following checklist and publishing it together with their proposal.
+
+```
+- [ ] I am being paid for working on this proposal by <ORGANIZATION>
+- [ ] I am an employee or member of <ORGANIZATION> which is involved in the Matrix ecosystem
+- [ ] I have another relationship that might appear to be a conflict of interest: <DETAILS>
+- [ ] I have some other type of financial interest in Matrix or this proposal: <DETAILS>
+- [ ] None of the above
+```
+
 ## Technical notes
 
 Proposals **must** develop Matrix as a layered protocol: with new
@@ -193,7 +206,8 @@ is as follows:
         filename that follows the format `1234-my-new-proposal.md` where
         `1234` is the MSC ID.
     -   Your PR description must include a link to the rendered Markdown
-        document and a summary of the proposal.
+        document and a disclosure of potential conflicts of interest where
+        they might reasonably apply.
     -   It is often very helpful to link any related MSCs or [matrix-spec
         issues](https://github.com/matrix-org/matrix-spec/issues) to give
         context for the proposal.
--- a/changelogs/internal/newsfragments/2079.new
+++ b/changelogs/internal/newsfragments/2079.new
@@ -1,0 +1,1 @@
+Make disclosure of conflicts of interest mandatory for spec change proposals.


content/proposals.md
```diff
@@ -193,7 +206,8 @@ is as follows:
         filename that follows the format `1234-my-new-proposal.md` where
         `1234` is the MSC ID.
     -   Your PR description must include a link to the rendered Markdown
-        document and a summary of the proposal.
+        document and a disclosure of potential conflicts of interest where
+        they might reasonably apply.
```
Johennes, line 210: Empirically, proposal summaries in the PR description don't appear to happen at all in practice. Therefore, I've dropped this part when adding the CoI clause.

content/proposals.md
```diff
@@ -79,6 +79,19 @@ Please [see
 MSC1779](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1779-open-governance.md)
 for full details of the project's Guiding Principles.
 
+For the purposes of transparency and to underpin the "greater benefit"
+principle, proposals MUST disclose possible conflicts of interest where
+they might reasonably exist. Authors MAY do this by filling out the
+following checklist and publishing it together with their proposal.
```
Johennes, line 85: It is debatable whether RFC2119 language makes sense here. I found it helpful for differentiating between hard and optional requirements. I might be partial to this language by having worked with it in the spec for too long, however.

richvdh (approved):
LGTM, but will leave open for a bit to gather other opinions

content/proposals.md
```diff
@@ -79,6 +79,19 @@ Please [see
...
+
+```
+- [ ] I am being paid for working on this proposal by <ORGANIZATION>
+- [ ] I am an employee or member of <ORGANIZATION> which is involved in the Matrix ecosystem
+- [ ] I have another relationship that might appear to be a conflict of interest: <DETAILS>
+- [ ] I have some other type of financial interest in Matrix or this proposal: <DETAILS>
+- [ ] None of the above
```
Gnuxie, line 92: This doesn't include a situation where the employer is being paid by another organisation to develop the proposal

Johennes: I guess for the author, the first checkbox would still be applicable. It's a good question whether companies sponsoring other companies should also be disclosed. This might actually be tricky because those relationships could be subject to contractual privacy agreements.

Gnuxie: > I guess for the author, the first checkbox would still be applicable

I feel like it would be a good idea to make that explicit

> This might actually be tricky because those relationships could be subject to contractual privacy agreements.

Yeah, however, nothing stops individuals also making these sorts of contractual agreements. And the first line would still apply there i think? So it makes sense for there to be consistency between individuals and employers. I don't know how either would disclose that they have to keep the customer secret though...

Johennes: >> I guess for the author, the first checkbox would still be applicable
>
> I feel like it would be a good idea to make that explicit

How about something like this?

> I am being paid for working on this proposal by \<ORGANIZATION\> (which might itself be funded or sponsored by other organizations)

KitsuneRal: While I appreciate an attempt to identify a conflict of interest that may hide anywhere down the chain of funding, I don't think it's prudent (let alone feasible) to unwrap the chain all the way, particularly in the checklist which is meant to streamline the declaration. The onus is on the author to declare the CoI wherever it sits - no kind of language can cover all specific cases. If the employer is funded by a private equity firm that has vested interest in X (pun intended), declaring that the employer is funded by that specific private equity firm doesn't expose CoI anyway.

content/proposals.md
```diff
@@ -79,6 +79,19 @@ Please [see
 MSC1779](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/1779-open-governance.md)
 for full details of the project's Guiding Principles.
 
+For the purposes of transparency and to underpin the "greater benefit"
+principle, proposals MUST disclose possible conflicts of interest where
+they might reasonably exist. Authors MAY do this by filling out the
+following checklist and publishing it together with their proposal.
+
+```
+- [ ] I am being paid for working on this proposal by <ORGANIZATION>
```
KitsuneRal, line 88:
```suggestion
- [ ] I am being paid (directly or indirectly) for working on this proposal by <ORGANIZATION>
```
@Gnuxie would that be better though?

Gnuxie: yeah that makes sense :+1:

Johennes: What I was trying to get at in the other thread is that only direct relationships should have to be disclosed. I think a lot of times, people writing MSCs will probably not even know what business relationships their employer has with other companies. Empirically, the joint ventures in Matrix projects can be quite complex. This formulation now makes me worry about having to name every other organisation that my employer gets income from.

Gnuxie: Well it does mean that the disclosure process is quite weak and doesn't put any accountability onto matrix vendors. All the disclosure is for individuals and allows vendors to hide behind their employees who will obviously only be doing work for their employer... which we likely already knew. I'm not expecting anyone to come up with a magic solution to fix that but it kinda sucks a bit now that i realise it.

Gnuxie: It is a double standard though.. really if your employer wants to contribute to matrix then they're gonna have to be upfront and make a disclosure statement on the MSC... if that's too much or legally grey for them... then why's the same not apply to individuals?

Gnuxie: To illustrate why this is a big deal: It allows the foundation to claim that the spec process has a disclosure process which on the surface sounds good. And does apply to individuals. But it does not apply to the interests of any matrix vendor. Because their disclosure is as simple as "I am a vendor employee", and there is no disclosure.

Gnuxie: Whereas as an individual i would likely need to declare all of the following information:
- That i have an option contact to buy shares as part of my previous employment at vccorp.example.com
- I am receiving funds from a grant from nice-people.example.com to complete a feature that is relevant to the MSC. <link to grant details>
- I am carrying out contractual work for small-company.example.com to complete xyz feature

This is a lot of relevant information that i should disclose. But the moment i create a legal company then i don't need to disclose anything.

Johennes: Yeah, this is definitely an issue. 😕

I'm not sure how to fix this. In a way, the entire disclosure builds on good faith and we cannot systematically verify it. The only thing we can do is, if at some point it comes out that you have misrepresented your conflicts of interest,  "you then get shunned as a bad faith actor" (as Rich put it on [#1700](https://github.com/matrix-org/matrix-spec/issues/1700#issuecomment-2674076034)).

Therefore, I was trying to phrase the checklist in a way that doesn't make the good faith, normal people obsess about their disclosures given that we have limited control over the bad faith people anyway.

Johennes:
Based on the discussion in https://github.com/matrix-org/matrix-spec/issues/1700, I think we can abandon this.

richvdh:
Thanks for making the effort :/

PR #2081, opened 2025-02-22
Adjust margins in rendered endpoints

zecakeh:
It felt that there was a lot of wasted space in those because of the spacing between tables. There are 2 commits here:

1. Use consistent spacing between tables and code blocks:

    Tables used `4 rem` between 2 tables which is more than the margin above a `h2` title, which seems excessive. We change it to `2 rem` which matches the default margin of code blocks and info boxes. To make sure that this stays consistent we apply it to both tables and code blocks.

    _I feel like we could probably reduce it further to `1.5 rem`. That would require to change the default margins for code blocks and info boxes for consistency._

    We also remove the margin if it's the last item of a `.rendered-data` block because this is just wasted space.

    |Before|After|
    |---|---|
    |image|image|
2. Adjust spacing around `h2` and `h3` titles:

    Given that `h2` titles are always preceded by an horizontal rule, we don't need a lot of space to separate them from the previous section.

    We also reduce the spacing when a `h2` title is followed directly by a `h3` title.

    Finally, we add a little spacing below both so that tables are less close to the title.

    |Before|After|
    |---|---|
    |image|image|

### Pull Request Checklist

Preview: https://pr2081--matrix-spec-previews.netlify.app


richvdh (approved):
LGTM

PR #2088, opened 2025-02-25
Replace Hugo shortcodes in OpenAPI output

### LINKED ISSUE #1873: Handlebars spills through into API viewer
**Link to problem area**:
https://github.com/matrix-org/matrix-spec/blob/7773716d189b16c1d5fa1dd2bfb8a1685bfb1108/data/api/client-server/content-repo.yaml#L226 and elsewhere in the same file.

**Issue**
API description files can be used by code generation tools, and our own [API viewer](https://playground.matrix.org) consumes them. See, for example [`GET /_matrix/media/v3/download/{serverName}/{mediaId}`](https://playground.matrix.org/#get-/_matrix/media/v3/download/-serverName-/-mediaId-). While it's still sort of readable through the handlebars, I'd rather stick to Markdown in API files.

zecakeh:
I went with simple prefix replacements, but for the `boxes` shortcodes, it might not be obvious where the note/rationale/warning ends if it spans several paragraphs. We might want to use blockquotes too.

Fixes #1873.

### Pull Request Checklist

Preview: https://pr2088--matrix-spec-previews.netlify.app

### DIFF
--- a/scripts/dump-openapi.py
+++ b/scripts/dump-openapi.py
@@ -37,10 +37,10 @@ api_dir = os.path.join(os.path.dirname(scripts_dir), "data", "api")
 # A shortcode is defined as (newlines and whitespaces for presentation purpose):
 #
 # {{%
-#     <one or more whitespaces>
+#     <zero or more whitespaces>
 #     <name of shortcode>
-#     <one or more whitespaces>
-#     (optional <list of parameters><one or more whitespaces>)
+#     (optional <one or more whitespaces><list of parameters>)
+#     <zero or more whitespaces>
 # %}}
 #
 # With:
@@ -48,7 +48,12 @@ api_dir = os.path.join(os.path.dirname(scripts_dir), "data", "api")
 # * <name of shortcode>: any word character and `-` and `/`.
 # * <list of parameters>: any character except `}`, must not start or end with a
 #   whitespace.
-shortcode_regex = re.compile(r"\{\{\%\s+(?P<name>[\w\/-]+)\s+(?:(?P<params>[^\s\}][^\}]+[^\s\}])\s+)?\%\}\}", re.ASCII)
+shortcode_regex = re.compile(r"""\{\{\%                                   # {{%
+                                 \s*                                      # zero or more whitespaces
+                                 (?P<name>[\w/-]+)                        # name of shortcode
+                                 (?:\s+(?P<params>[^\s\}][^\}]+[^\s\}]))? # optional list of parameters
+                                 \s*                                      # zero or more whitespaces
+                                 \%\}\}                                   # %}}""", re.ASCII | re.VERBOSE)
 
 # Parses the parameters of a Hugo shortcode.
 #
--- a/scripts/dump-openapi.py
+++ b/scripts/dump-openapi.py
@@ -32,6 +32,29 @@ import yaml
 scripts_dir = os.path.dirname(os.path.abspath(__file__))
 api_dir = os.path.join(os.path.dirname(scripts_dir), "data", "api")
 
+# Finds a Hugo shortcode in a string.
+#
+# A shortcode is defined as (newlines and whitespaces for presentation purpose):
+#
+# {{%
+#     <one or more whitespaces>
+#     <name of shortcode>
+#     <one or more whitespaces>
+#     (optional <list of parameters><one or more whitespaces>)
+# %}}
+#
+# With:
+#
+# * <name of shortcode>: any word character and `-` and `/`.
+# * <list of parameters>: any character except `}`, must not start or end with a
+#   whitespace.
+shortcode_regex = re.compile(r"\{\{\%\s+(?P<name>[\w\/-]+)\s+(?:(?P<params>[^\s\}][^\}]+[^\s\}])\s+)?\%\}\}", re.ASCII)
+
+# Parses the parameters of a Hugo shortcode. 
+#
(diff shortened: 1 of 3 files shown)


zecakeh:
This also updates the Python version used in CI because `match` was added in Python 3.10.

scripts/dump-openapi.py
```diff
@@ -44,17 +67,95 @@ def prefix_absolute_path_references(text, base_url):
...
+        if match['params']:
+            for param in shortcode_params_regex.finditer(match['params']):
+                if param['key']:
+                    params[param['key']] = param['value']
+        
+        shortcode = {
+            'name': match['name'],
+            'params': params,
+            'match': match,
+        }
```
richvdh, line 135: Not a fan of this "`dict` as struct" usage. It makes Python into 1997-style Javascript and is generally hard to follow.

We've only got three properties here: just pass them separately into `replace_shortcode`? In fact, `name` is arguably redundant. 

Alternatively, we already have a dependency on [`attrs`](https://www.attrs.org/en/stable/) for these scripts, so define a class for them:

```
import attrs

@attrs.define
class ShortCodeInstance:
    name: str
    params: Dict[str, str]
    match: Match
```

scripts/dump-openapi.py
```diff
@@ -44,17 +67,95 @@ def prefix_absolute_path_references(text, base_url):
     """
     return text.replace("](/", "]({}/".format(base_url))
 
-def edit_links(node, base_url):
-    """Finds description nodes and makes any links in them absolute."""
+def replace_match(text, match, replacement):
+    """Replaces the regex match by the replacement in the text."""
+    return text[:match.start()] + replacement + text[match.end():]
```
richvdh, line 72: it might be worth noting that `text` is also accessible as `match.string` here, so you could avoid threading `text` down the call stack. I'm not entirely sure if that makes things clearer though!

scripts/dump-openapi.py
```diff
@@ -44,17 +67,95 @@ def prefix_absolute_path_references(text, base_url):
...
+                    params[param['key']] = param['value']
+        
+        shortcode = {
+            'name': match['name'],
+            'params': params,
+            'match': match,
+        }
+        text = replace_shortcode(text, shortcode)
+
+    return text
```
richvdh, line 138: Won't this leave the closing tag (`{{% /boxes/warning %}}` etc) in place?

Maybe you need to do two passes: one for `added-in` and `changed-in`, which have no body, and then one for matching pairs of `{{% boxes/foo %}} ... {{% /boxes/foo %}}`

scripts/dump-openapi.py
```diff
@@ -32,6 +32,29 @@
...
+#     <one or more whitespaces>
+#     (optional <list of parameters><one or more whitespaces>)
+# %}}
+#
+# With:
+#
+# * <name of shortcode>: any word character and `-` and `/`.
+# * <list of parameters>: any character except `}`, must not start or end with a
+#   whitespace.
+shortcode_regex = re.compile(r"\{\{\%\s+(?P<name>[\w\/-]+)\s+(?:(?P<params>[^\s\}][^\}]+[^\s\}])\s+)?\%\}\}", re.ASCII)
```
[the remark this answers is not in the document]
zecakeh: > Also, what's with the re.ASCII flag?

It makes the regex only match ASCII characters, which makes sense IMO given that we are matching code.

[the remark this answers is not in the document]
richvdh: couldn't a shortcode param contain utf-8?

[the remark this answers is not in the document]
zecakeh: Sorry, I should have been clearer, in our case it makes `\w` match ASCII words (for the shortcode name, and the params keys) and `\s` match ASCII whitespaces. Any param value is accepted.

scripts/dump-openapi.py
zecakeh: In `replace_shortcode` we remove all closing tags, without caring if they are associated to an opening tag.

scripts/dump-openapi.py
```diff
@@ -44,17 +67,95 @@ def prefix_absolute_path_references(text, base_url):
...
+
+    return text
+
+def edit_descriptions(node, base_url):
+    """Finds description nodes and apply fixes to them.
+    
+    The fixes that are applied are:
+
+    * Make links absolute
+    * Replace shortcodes
```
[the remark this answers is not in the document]
zecakeh: Applied

scripts/dump-openapi.py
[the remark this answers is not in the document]
zecakeh: > It might be a good idea to use [re.VERBOSE](https://docs.python.org/3/library/re.html#re.VERBOSE) here, and add some line breaks.

Done, let me know if you would like to clarify the regex even more.

> In [\w\/-]: I don't think the / needs quoting?

Indeed, it's not necessary in Python. I used a website to build the regex that required it.

> Is whitespace within the delimiters required? is {{%foo%}} not valid?

Indeed, it is valid, I changed the regex.

scripts/dump-openapi.py
zecakeh: I removed it and passed just the match instead to `replace_shortcode`.

scripts/dump-openapi.py
zecakeh: Indeed, I changed the arguments to only take the `match` and `replacement`.

PR #2093, opened 2025-03-13
MSC4260: Reporting users (Client-Server API)

Johennes:
Spec for https://github.com/matrix-org/matrix-spec-proposals/pull/4260

### Pull Request Checklist

### DIFF
--- a/content/client-server-api/modules/report_content.md
+++ b/content/client-server-api/modules/report_content.md
@@ -29,3 +29,20 @@ is in before accepting a report.
 based on whether or not the reporting user is joined to the room. This is
 because users can be exposed to harmful content without being joined to a
 room. For instance, through room directories or invites.
+
+{{% added-in v="1.14" %}} Similarly, servers MUST NOT restrict user reports
+based on whether or not the reporting user is joined to any rooms that the
+reported user is joined to. This is because users can be exposed to harmful
+content without being joined to a room. For instance, through user
+directories or invites.
+
+Clients can infer whether a reported event, room or user exists based on the
+404 responses of the reporting endpoints. Homeservers that wish to conceal
+this information MAY return 200 responses regardless of the existence of the
+reported subject.
+
+Furthermore, it might be possible for clients to deduce whether a reported
+event, room or user exists by timing the response. This is because only a
+report for an existing subject will require the homeserver to do further
+processing. To combat this, homeserver implementations MAY add a random
+delay when generating a response.
--- a/data/api/client-server/report_content.yaml
+++ b/data/api/client-server/report_content.yaml
@@ -88,12 +88,6 @@ paths:
         Reports an event as inappropriate to the server, which may then notify
         the appropriate people. The caller must be joined to the room to report
         it.
-        
-        It might be possible for clients to deduce whether an event exists by
-        timing the response, as only a report for an event that does exist
-        will require the homeserver to check whether a user is joined to
-        the room. To combat this, homeserver implementations should add
-        a random delay when generating a response.
       operationId: reportEvent
       parameters:
         - in: path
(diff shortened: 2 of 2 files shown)


data/api/client-server/report_content.yaml
```diff
@@ -164,6 +158,73 @@ paths:
...
+          application/json:
+            schema:
+              type: object
+              example: {
+                "reason": "this makes me sad"
+              }
+              properties:
+                reason:
+                  type: string
+                  description: The reason the room is being reported.
```
Johennes, line 192: Contrary to the MSC I made this field optional due to https://github.com/matrix-org/matrix-spec/pull/1938#discussion_r1730832640.

content/client-server-api/modules/report_content.md
```diff
@@ -29,3 +29,20 @@ is in before accepting a report.
...
+Clients can infer whether a reported event, room or user exists based on the
+404 responses of the reporting endpoints. Homeservers that wish to conceal
+this information MAY return 200 responses regardless of the existence of the
+reported subject.
+
+Furthermore, it might be possible for clients to deduce whether a reported
+event, room or user exists by timing the response. This is because only a
+report for an existing subject will require the homeserver to do further
+processing. To combat this, homeserver implementations MAY add a random
+delay when generating a response.
```
Johennes, line 48: This has previously not been stated for the room-reporting endpoint. However, I cannot think of a reason why it wouldn't apply there, too. So I'd consider adding it to be a clarification that I'm combining into this spec changes.

content/client-server-api/modules/report_content.md
```diff
@@ -29,3 +29,20 @@ is in before accepting a report.
...
+{{% added-in v="1.14" %}} Similarly, servers MUST NOT restrict user reports
+based on whether or not the reporting user is joined to any rooms that the
+reported user is joined to. This is because users can be exposed to harmful
+content without being joined to a room. For instance, through user
+directories or invites.
+
+Clients can infer whether a reported event, room or user exists based on the
+404 responses of the reporting endpoints. Homeservers that wish to conceal
+this information MAY return 200 responses regardless of the existence of the
+reported subject.
```
Johennes, line 42: This has previously not been stated explicitly for the room- and event-reporting endpoints. I would consider it implied though, especially since the event-reporting endpoint covers the more sophisticated timing attack.

data/api/client-server/report_content.yaml
```diff
@@ -164,6 +158,73 @@ paths:
...
+        Clients may wish to [ignore](#ignoring-users) users after reporting them.
+      operationId: reportUser
+      parameters:
+        - in: path
+          name: userId
+          description: The user being reported.
+          required: true
+          example: "@someguy:example.com"
+          schema:
+            type: string
```
zecakeh, line 180: This will identify the string as a user ID in the spec:

```suggestion
            type: string
            format: mx-user-id
            pattern: "^@"
```

Johennes: Oh, nice. Thank you.

content/client-server-api/modules/report_content.md
turt2live: This would require an MSC to change how the report room endpoint works, because the MSC was quite clear that it's a 404.

content/client-server-api/modules/report_content.md
turt2live: This would be part of the MSC for changing the report room endpoint.

data/api/client-server/report_content.yaml
turt2live: err, I missed this in the report room stuff. It was intentional that `reason` is required, and the implementation also requires it. The string can be empty, but *must* be provided.

turt2live (changes requested):
otherwise lgtm - thanks for writing it up!

Johennes:
I have opened https://github.com/matrix-org/matrix-spec-proposals/pull/4277 for future alignment of the three endpoints.

turt2live (approved):
thanks, and sorry for the thrashing on `required` - future MSCs will hopefully document these differences better.

Issue #633, opened 2020-06-05
what is a "public room"


richvdh:

bkil:
Somewhat related: #620 #830

Johennes:
I think there are two things we could do here:

1. We clean up the naming of endpoints and identifiers to not use "public" with different semantics. For instance, `visibility: public, private` could become `visibility: visible, hidden`, `/publicRooms` could become `/directory/list/rooms` and so on.
2. We unpack the term "public" in the spec text to its concrete meaning. Let's consider this passage, for instance:
   > Homeservers MUST at a minimum allow profile look-up for [...] users that reside in public rooms known to the homeserver".

   What this (probably) means to say is that profile look-up should be possible for users that the requester would be able to see anyway due to room access. So "public" is meant in the sense of join rules only and we could rephrase it as:

   > Homeservers MUST at a minimum allow profile look-up for [...] users that reside in <del>public</del> rooms _with `public` join rules_ known to the homeserver".

1 sounds like a sizeable undertaking in terms of migrating the ecosystem. It would also not save us from doing 2. It would just be an optional prior step.

Personally, I find 2 more important than 1. The semantics of "public" in the spec text are currently rather ambiguous. Things like `join_rule: public` or `visibility: public`, on the other hand, bind "public" to a certain dimension which feels less confusing.

Curious what others think.

Johennes:
I have taken a stab at 2. in the series of PRs linked since my last comment.

anoadragon453:
@Johennes Synapse [considers](https://github.com/element-hq/synapse/blob/7cee1a190f3b0bb20d5796b238fe59381c9c8aff/synapse/storage/databases/main/room.py#L524-L526) rooms with `knock` and `knock_restricted` join rules part of its `/publicRooms` response. This stems from MSC2403, which [specifically added](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/2403-knock.md#changes-regarding-the-public-rooms-directory) rooms with the `knock` join rule to `/publicRooms`, so that users could find rooms to knock on.

Does this change the definition of a "public room" to one that has the `join_rule` `public`, `knock` or `knock_restricted`?

Johennes:
I've moved this to https://github.com/matrix-org/matrix-spec/pull/2104#discussion_r2003394076 where there's already been a related discussion.

Johennes:
[...]

PR #2104, opened 2025-03-19
Clarify the meaning of "public rooms" in the room directory

Johennes:
Relates to: #633

### Pull Request Checklist

Preview: https://pr2104--matrix-spec-previews.netlify.app


data/api/server-server/public_rooms.yaml
```diff
@@ -62,21 +66,18 @@ paths:
...
+      summary: Lists the server's published room directory with an optional filter
       description: |-
-        Lists the public rooms on the server, with optional filter.
+        Lists the server's published room directory with an optional filter.
 
         This API returns paginated responses. The rooms are ordered by the number
         of joined members, with the largest rooms first.
-
-        Note that this endpoint receives and returns the same format that is seen
-        in the Client-Server API's `POST /publicRooms` endpoint.
```
Johennes, line 79: This felt slightly confusing because the CS endpoint has a `server` parameter that is missing here. I think the identity of the request and response bodies is clear enough without this note.

data/api/server-server/public_rooms.yaml
```diff
@@ -147,69 +148,11 @@ paths:
...
         "200":
-          description: A list of the rooms on the server.
+          description: A filtered list of the published rooms on the server.
           content:
             application/json:
               schema:
-                type: object
-                description: A list of the rooms on the server.
-                required:
-                  - chunk
```
Johennes, line 157: AFAICT this is the only difference to `public_rooms_response.yaml`. I cannot imagine that `chunk` was supposed to be required on only one of the four endpoints so I have made this reuse the shared schema.

data/api/client-server/definitions/public_rooms_response.yaml
```diff
@@ -13,28 +13,15 @@
...
-          title: PublicRoomsChunk
-          properties:
-            # Override description of join_rule
-            join_rule:
-              type: string
-              description: |-
-                The room's join rule. When not present, the room is assumed to
-                be `public`. Note that rooms with `invite` join rules are not
-                expected here, but rooms with `knock` rules are given their
-                near-public nature.
```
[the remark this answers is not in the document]
zecakeh: I believe that it's more a clarification about which join rules clients can expect to find here.

I agree that this should be next to the meaning of visibility.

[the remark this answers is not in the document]
Johennes: Have added a phrase about knockable rooms in the warning box.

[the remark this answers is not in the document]
Johennes: Thanks for the context @anoadragon453! 🙏

So when responding to `/publicRooms`, Synapse first retrieves all rooms with `public` visibility but then applies an additional filter to only include these cases:

```
join_rules = '{JoinRules.PUBLIC}'
OR join_rules = '{JoinRules.KNOCK}'
OR join_rules = '{JoinRules.KNOCK_RESTRICTED}'
OR history_visibility = 'world_readable'
```

I'm struggling a bit with understanding how the room directory is _meant_ to work. Other than the slightly unclear statement "rooms with `knock` rules are given their near-public nature", the spec doesn't mention any restrictions on what rooms can be included in `/publicRooms`. Therefore, my assumption was that rooms with a `public` visibility would show up regardless of their join rules or history visibility.

In fact, [MSC2403](https://github.com/matrix-org/matrix-spec-proposals/issues/2403) seems to only define `join_rule` as a new field in the response of `/publicRooms`. It doesn't appear to specify any filtering and instead says:

> The spec does not prevent us from adding rooms with 'knock' join_rules to the public rooms directory.

😵‍💫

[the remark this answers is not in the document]
anoadragon453: Correct, the room is only considered if `visibility` is `public` when modifying the public rooms list with [`PUT /_matrix/client/v3/directory/list/room/{roomId}`](https://spec.matrix.org/v1.13/client-server-api/#put_matrixclientv3directorylistroomroomid).

To speak to Synapse's implementation, the `join_rule` restriction within the query was only added during a [performance improvement change](https://github.com/matrix-org/synapse/pull/6019/files#diff-19a2cd8fb2449c768a87f070a690453cd96635efbb77839fce4ae823fd291adaR107). I could see this requirement being useful, in that if a room is set to `private`, it would immediately no longer show in the public rooms list (as it is not publicly-joinable). While the spec may not require homeservers to manage their lists in that way, Synapse can choose to do so.

Regardless, it's clear that Synapse's filtering of its public rooms list here should not influence the spec's definition of a public room. So I think we can ignore this and consider a `join_rule` of `public` the only variant that constitutes a room as public.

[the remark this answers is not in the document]
Johennes: > To speak to Synapse's implementation, the join_rule restriction within the query was only added during a [performance improvement change](https://github.com/matrix-org/synapse/pull/6019/files#diff-19a2cd8fb2449c768a87f070a690453cd96635efbb77839fce4ae823fd291adaR107). I could see this requirement being useful, in that if a room is set to private, it would immediately no longer show in the public rooms list (as it is not publicly-joinable). While the spec may not require homeservers to manage their lists in that way, Synapse can choose to do so.

Oh, interesting. This sounds sensible. Maybe we could add this additional filtering as a MAY hint into the spec?

> So I think we can ignore this and consider a `join_rule` of `public` the only variant that constitutes a room as public.

You mean `visibility` of `public`?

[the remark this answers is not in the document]
anoadragon453: > Oh, interesting. This sounds sensible. Maybe we could add this additional filtering as a MAY hint into the spec?

Sounds reasonable to me.

> You mean _`visibility`_ of `public`?

Hah, yes sorry. `join_rules` aren't mandated affect the filtering. Terminology!

data/api/server-server/public_rooms.yaml
```diff
@@ -13,16 +13,20 @@
...
       description: |-
-        Gets all the public rooms for the homeserver. This should not return
-        rooms that are listed on another homeserver's directory, just those
-        listed on the receiving homeserver's directory.
+        Lists the server's published room directory.
+
+        This API returns paginated responses. The rooms are ordered by the number
+        of joined members, with the largest rooms first.
+
+        This should not return rooms that are listed on another homeserver's directory,
```
HarHarLinks, line 28:
```suggestion
        This SHOULD not return rooms that are listed on another homeserver's directory,
```

data/api/client-server/definitions/public_rooms_response.yaml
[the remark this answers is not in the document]
HarHarLinks: From my understanding of the spec, indeed this endpoint's response SHOULD be entirely independent of what rooms' `join_rules` are, even though it does seem reasonable for an implementation to filter out `join_rule: private` rooms as discovering a room this way is mostly useless.

content/client-server-api/_index.md
```diff
@@ -2831,7 +2831,38 @@ re-invited.
...
+### Room directory
+
+Homeservers MAY publish a room directory to allow users to discover rooms. A room
+can have one of two visibility settings in the directory:
+
+`public`
+The room will be shown in the published room directory.
+
+`private`
+The room will be hidden from the published room directory.
```
anoadragon453, line 2843:
```suggestion
-   `public`: The room will be shown in the published room directory.
-   `private`: The room will be hidden from the published room directory.
```

nit: since this appeared to fundamentally be a list.

content/client-server-api/_index.md
```diff
@@ -2831,7 +2831,38 @@ re-invited.
...
+[join rule](#mroomjoin_rules) or [history visibility](#room-history-visibility).
+
+In particular, a visibility setting of `public` should not be confused with a `public`
+join rule. Rooms with a join rule of `knock`, for instance, could reasonably be published
+in the directory, too.
+
+Similarly, a visibility setting of `public` does not necessarily imply a `world_readable`
+history visibility.
+
+To increase performance, servers MAY apply additional filters when listing the
```
anoadragon453, line 2862:
```suggestion
To increase performance or by preference, servers MAY apply additional filters when listing the
```

To keep things flexible.

anoadragon453 (dismissed):
Just some minor notes, otherwise this looks good to me. Thank you!

I'd also like to explicitly highlight the transition from "public room directory" to "published room directory" as being one the ecosystem will need to adapt to. And if you could just audit the diff to make sure we're using "published room directory" vs. "room directory" in all the right places.

content/client-server-api/_index.md
Johennes: Good point 👍

Johennes:
> I'd also like to explicitly highlight the transition from "public room directory" to "published room directory" as being one the ecosystem will need to adapt to.

I'm slightly unsure how exactly to reflect this in the spec text. Or did you just mean calling it out explicitly in the changelog?

data/api/client-server/create_room.yaml
```diff
@@ -87,12 +87,9 @@ paths:
...
                     - private
                   description: |-
-                    A `public` visibility indicates that the room will be shown
-                    in the published room list. A `private` visibility will hide
-                    the room from the published room list. Rooms default to
-                    `private` visibility if this key is not included. NB: This
-                    should not be confused with `join_rules` which also uses the
-                    word `public`.
+                    The room's visibility in the server's
+                    [published room directory](#published-room-directory).
```
richvdh, line 91: This needs more anchoring, so that it works in https://playground.matrix.org/:

```suggestion
                    [published room directory](/client-server-api#published-room-directory).
```

richvdh (approved):
LGTM otherwise

data/api/client-server/create_room.yaml
Johennes: Sorry, forgot about this again. 🤦‍♂️

Issue #2116, opened 2025-03-27
Historical spec builds don't say they're historical


turt2live:
**Link to problem area**: https://spec.matrix.org/v1.13/

**Issue**

The deployed version of the v1.13 spec is the `spec-historical` artifact, but the spec doesn't say it's old. This issue was introduced in v1.13 and is a problem in v1.14 too (through local/manual verification of the artifact).

v1.12 is unaffected: https://spec.matrix.org/v1.12/

turt2live:
I've manually fixed this for v1.13:

![Image](https://github.com/user-attachments/assets/69e1287b-a5b9-4eaa-9204-47a4fbbaaaaa)

```
<!-- MANUALLY ADDED - TravisR -->
            <div style="margin-top:0;margin-right:0;margin-left:0;border:0;border-left:solid 5px #0098d4;background-color:#fbfbfb;max-width:80%;font-weight:500;color:inherit;border-radius:0;margin-bottom:2rem;padding:1.5rem;padding-bottom:.5rem;">
  <p>You're looking at an old version of this specification.<p>
  <p><a href="https://spec.matrix.org/latest">Switch to the current stable release</a>.</p>
</div>
<!-- /MANUALLY ADDED -->
```

it's an ugly copy/paste of the v1.12 element with inline styles because I guess unused classes get compiled out.

PR #2123, opened 2025-03-29
Fix generation of historical spec

zecakeh:
With the move of the config file, the command in CI did not work as expected anymore.
I am unsure why Hugo didn't complain about the missing config file from the command…

To avoid this problem in the future and simplify the job, we use the default config and add an environment variable for the status which will always take precedence over the config.

This should take care of #2116 for future versions.

~~I enabled the historical job for every workflow temporarily to check that it builds as expected.~~ To see the fixed historical build, you need to get the artifact from one of the 2 first commits, not from the last one.

### Pull Request Checklist

Preview: https://pr2123--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -265,7 +265,7 @@ jobs:
     name: "📖 Build the historical backup spec"
     runs-on: ubuntu-latest
     needs: [build-openapi]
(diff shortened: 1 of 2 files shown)


richvdh (approved):
much better, thanks!

PR #2132, opened 2025-04-09
[schema] Application Service Registration meta schema

V02460:
Sets the meta schema for the Application Service Registration to `https://json-schema.org/draft/2020-12/schema`. Specifying the meta schema clarifies the semantics of the schema files and allows tools to consume it correctly.

It seems to me that the `registration.yaml` and `namespace_list.yaml` are completely disjunct from the other OpenAPI schemas in the `data/api/application-service` folder. Because the other files follow OpenAPI v3.1.0, I still chose to use JSON Schema version 2020-12, which is the version that [OpenAPI v3.1.0 data types and schema objects are based on](https://spec.openapis.org/oas/v3.1.0.html#data-types).

Relevant parts of the OpenAPI v3.1.0 spec that justify the inclusion of the `$schema` keyword:

> [The Schema Object] is a superset of the JSON Schema Specification Draft 2020-12.

> Unless stated otherwise, the property definitions follow those of JSON Schema and do not add any additional semantics.

From https://spec.openapis.org/oas/v3.1.0.html#schema-object

### Pull Request Checklist

Preview: https://pr2132--matrix-spec-previews.netlify.app


V02460:
@tulir: I don’t think that is necessary. These two files are special in that they

a) are useful as JSON Schema files independently of OpenAPI definitions. While OpenAPI has well-defined implicit meta schemas, that’s not the case for JSON Schemas.
b) are not actually used by any of the OpenAPI files – I think they were simply snuck in there for lack of a better location.

KitsuneRal:
I agree that OpenAPI files don't necessarily need that because they have their own schema version designator already. However, we also have event schemas written as JSON Schema files (despite them having `.yaml` extension); point a) certainly applies to them as well. Point b) is rather irrelevant: the rationale of adding `schema` to JSON Schema files (clearer semantics and tooling-friendliness) is the same regardless of where they are located.

KitsuneRal:
Both; but `event-schemas` are most interesting.

V02460:
Also I found out that schemas are already checked against `https://json-schema.org/draft/2020-12/schema`:

https://github.com/matrix-org/matrix-spec/blob/fb4a0d8f66fb6263786ab7c4de3ef21517fcc64c/scripts/check-json-schemas.py#L139

https://github.com/matrix-org/matrix-spec/blob/fb4a0d8f66fb6263786ab7c4de3ef21517fcc64c/scripts/check-json-schemas.py#L175-L182.

V02460:
Do I need to add a second newsfragment somewhere for the event schema changes?

anoadragon453:
@V02460 event definitions are currently considered to be part of the C-S API. A newsfile under `changelogs/client_server/newsfragments` would make sense to me.

KitsuneRal (approved):

anoadragon453 (approved):

PR #2133, opened 2025-04-15
Clarify wording around public room directory

HarHarLinks:
This PR suggest a switch to clearer wording in the API descriptions of what is commonly known as "(public) room directory", which contains "published rooms" (namely, published to the directory), to differentiate it more clearly from what is commonly known as "public rooms" (a room with the join rule = public).

Sadly, the name "publicRooms" is baked into the api; ideally it would be named similar to the user directory, however a migration path to that would be rough and certainly take an MSC.

### Pull Request Checklist

Preview: https://pr2133--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2133.clarification
+++ b/changelogs/client_server/newsfragments/2133.clarification
@@ -1,0 +1,1 @@
+Clarify wording around public room directory. Contributed by @HarHarLinks.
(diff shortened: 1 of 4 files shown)


HarHarLinks:
Beyond the initial scope of this PR, the C2S API also calls the room directory "Listing Rooms" and does not include an introductory paragraph to the topic, while the S2S API does have such a paragraph that both uses the term "room directory" and refers to the C2S API - a reference that given the lack of mention of the same term in a heading is not easy to follow.

What is the SCT's opinion on this whole matter? Can/should I try to fix this in this/another PR?

HarHarLinks:
...welp, now that I've done the work, I see there is substantial overlap with #2104.

PR #2141, opened 2025-05-14
Split account registration and management section and definitions

zecakeh:
Since account locking and suspension are authentication API agnostic, this is a pre-requisite to adding the new OAuth 2.0-based API.

This also splits the endpoints that where all included in the registration OpenAPI data, to separate them cleanly in the spec, and avoid having deactivation show before registration.

Fixes #1073.

### Pull Request Checklist

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

PR #2147, opened 2025-05-24
Add the OAuth 2.0 server metadata discovery endpoint

zecakeh:
Based on #2141.

As per MSC2965.

### Pull Request Checklist

Preview: https://pr2147--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/oauth_server_metadata.yaml
+++ b/data/api/client-server/oauth_server_metadata.yaml
@@ -1,0 +1,176 @@
+# Copyright 2025 The Matrix.org Foundation C.I.C.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+openapi: 3.1.0
+info:
+  title: Matrix Client-Server OAuth 2.0 Server Metadata Discovery API
+  version: 1.0.0
+paths:
+  "/auth_metadata":
+    get:
+      summary: Get the OAuth 2.0 authorization server metadata.
+      description: |-
+        Gets the OAuth 2.0 authorization server metadata, as defined in
+        [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414), including the
+        endpoint URLs and the supported parameters that can be used by the
+        clients.
+
+        The definition of this endpoint includes only the fields that are
+        meaningful in the context of the Matrix specification. The full list of
+        possible fields is available in the [OAuth Authorization Server Metadata
+        registry](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#authorization-server-metadata),
+        and normative definitions of them are available in their respective
+        RFCs.
+
+        {{% boxes/note %}}
+        The authorization server metadata is relatively large and may change
+        over time. Clients should:
+
+        - Cache the metadata appropriately based on HTTP caching headers
+        - Refetch the metadata if it is stale
+        {{% /boxes/note %}}
+      operationId: getAuthMetadata
+      responses:
+        "200":
+          description: The OAuth 2.0 authorization server metadata.
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  issuer:
+                    type: string
+                    format: uri
+                    description: |-
+                      The authorization server's issuer identifier, which is a URL that uses the
+                      `https` scheme and has no query or fragment components.
+
+                      This is not used in the context of the Matrix specification, but is required
+                      by [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414).
+                  authorization_endpoint:
+                    type: string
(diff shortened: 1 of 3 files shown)


data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -0,0 +1,176 @@
...
+                  "token_endpoint": "https://account.example.com/oauth2/token",
+                  "registration_endpoint": "https://account.example.com/oauth2/clients/register",
+                  "revocation_endpoint": "https://account.example.com/oauth2/revoke",
+                  "response_types_supported": ["code"],
+                  "grant_types_supported": ["authorization_code", "refresh_token"],
+                  "response_modes_supported": ["query", "fragment"],
+                  "code_challenge_methods_supported": ["S256"],
+                }
+      tags:
+        - Session management
```
richvdh, line 164: I wonder if this is the best place. Maybe we should start a new section for the OAuth APIs?

data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -0,0 +1,176 @@
...
+  "/auth_metadata":
+    get:
+      summary: Get the OAuth 2.0 authorization server metadata.
+      description: |-
+        Gets the OAuth 2.0 authorization server metadata, as defined in
+        [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414), including the
+        endpoint URLs and the supported parameters that can be used by the
+        clients.
+
+        The definition of this endpoint includes only the fields that are
```
richvdh:
```suggestion
        This endpoint definition includes only the fields that are
```

... otherwise it sounds like "this endpoint is defined to only return these fields"

data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -0,0 +1,176 @@
...
+                    items:
+                      type: string
+                      description: A response mode that the server supports.
+                  code_challenge_methods_supported:
+                    type: array
+                    description: |-
+                      List of OAuth 2.0 Proof Key for Code Exchange (PKCE) code challenge methods
+                      that the server supports at the authorization endpoint.
+
+                      This array MUST contain at least the `s256` value, for improved security in
```
richvdh:
```suggestion
                      This array MUST contain at least the `S256` value, for improved security in
```

data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -0,0 +1,176 @@
...
+                      List of OpenID Connect prompt values that the server supports at the
+                      authorization endpoint.
+
+                      Only the `create` value defined in [Initiating User Registration via OpenID
+                      Connect](https://openid.net/specs/openid-connect-prompt-create-1_0.html) is
+                      supported, for a client to signal to the server that the user desires to
+                      register a new account.
+                    items:
+                      type: string
+                      description: A prompt value that the server supports.
```
richvdh, line 141: For the historical record: this appears to have come from MSC2964.

data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -0,0 +1,176 @@
...
+                  prompt_values_supported:
+                    type: array
+                    description: |-
+                      List of OpenID Connect prompt values that the server supports at the
+                      authorization endpoint.
+
+                      Only the `create` value defined in [Initiating User Registration via OpenID
+                      Connect](https://openid.net/specs/openid-connect-prompt-create-1_0.html) is
+                      supported, for a client to signal to the server that the user desires to
+                      register a new account.
```
richvdh, line 138: Given that this isn't defined in RFC8414, maybe we should link to https://openid.net/specs/openid-connect-prompt-create-1_0.html#name-discovery-metadata as (I think?) the authoritative definition?

data/api/client-server/oauth_server_metadata.yaml
zecakeh: I have no strong opinion. I guess that we could rename the existing tag "Legacy Session Management" in this case and create a new one for the OAuth APIs ?

data/api/client-server/oauth_server_metadata.yaml
zecakeh: I am not sure what you mean by this comment, it looks like you are saying that we are lacking a link to the specification but there is already a link to this spec in this description. Or maybe you would prefer the link to be exactly the one you provided?

data/api/client-server/oauth_server_metadata.yaml
zecakeh: Fixed

data/api/client-server/oauth_server_metadata.yaml
zecakeh: Changed

data/api/client-server/oauth_server_metadata.yaml
richvdh: Sure, that sounds good. I think otherwise it's all going to be a bit cluttered on playground.matrix.org.

data/api/client-server/oauth_server_metadata.yaml
richvdh: I have no idea what I mean either. Sorry.

data/api/client-server/oauth_server_metadata.yaml
richvdh: that said: there aren't going to be many OAuth APIs, are there?

Let's leave it for now, and maybe come back to it in future.

richvdh (approved):
LGTM, but there is a conflict now

richvdh:
Thanks!

PR #2169, opened 2025-06-26
Clarify that the `latest_event` field is serialized in the same form as the event itself

bnjbvr:
This belongs to the category of "other changes", as this is mostly a clarification.

Preview: https://pr2169--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2169.clarification
+++ b/changelogs/client_server/newsfragments/2169.clarification
@@ -1,1 +1,1 @@
-The `latest_event` in a bundled thread information uses the same format as the event itself.
+The `latest_event` in an aggregated set of thread events uses the same format as the event itself.


changelogs/client_server/newsfragments/2169.clarification
```diff
@@ -0,0 +1 @@
+The `latest_event` in a bundled thread information uses the same format as the event itself.
```
anoadragon453:
```suggestion
The `latest_event` in an aggregated set of thread events uses the same format as the event itself.
```

anoadragon453 (approved):
A small wording change to the newsfile. But otherwise this LGTM. Thanks for reducing the ambiguity in the spec!

PR #2182, opened 2025-07-25
Fix new redocly lints

zecakeh:
Redocly had a new major release and it seems that a new lint was triggered in CI.

The definitions or `Protocol` triggered a new lint because the `instances` field is not defined in `protocol_base.yaml`, so it cannot be in the `required` array in that file. That condition must be moved to where this field is defined.

The other changes were required because the definitions provoked a crash in the new version of Redocly (https://github.com/Redocly/redocly-cli/issues/2224). While the crash should not happen in the first place, we can avoid it in our case by cleaning up the definitions:
  - For the room summary endpoint, the `required` array was duplicated at each level of nesting. Keeping it only on the base definition gets rid of the crash.
  - For the `SyncStateEvent` definition, we use `allOf` only for `$ref` and add other fields at the top level. It actually makes more sense to declare the added field in a schema that is also an item of `allOf`, which is what this does.

This doesn't change anything in the rendered spec.

### Pull Request Checklist

### DIFF
--- a/data/api/application-service/definitions/protocol.yaml
+++ b/data/api/application-service/definitions/protocol.yaml
@@ -23,3 +23,4 @@ allOf:
       type: array
       items:
         $ref: protocol_instance.yaml
+  required: ['instances']
--- a/data/api/application-service/definitions/protocol_base.yaml
+++ b/data/api/application-service/definitions/protocol_base.yaml
@@ -77,4 +77,4 @@ properties:
         "placeholder": "#foobar"
       }
     }
-required: ['user_fields', 'location_fields', 'icon', 'field_types', 'instances']
+required: ['user_fields', 'location_fields', 'icon', 'field_types']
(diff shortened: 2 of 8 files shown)


changelogs/application_service/newsfragments/2182.clarification
```diff
@@ -0,0 +1 @@
+Fix JSON schemas
```
richvdh:
```suggestion
Minor fixes to JSON schemas.
```

changelogs/client_server/newsfragments/2182.clarification
```diff
@@ -0,0 +1 @@
+Fix JSON schemas
```
richvdh:
```suggestion
Minor fixes to JSON schemas.

richvdh (approved):
lgtm!

PR #2188, opened 2025-08-03
Clarify terminology for keys in cross-signing module

codedust:
- do not use the term 'cross-signing keys' anymore: Previously, the term 'cross-signing keys' was used to refer to the master, user-signing and self-signing keys. This is not ideal since the master key is used for cross-signing but may also be used to sign the backup key, for example. In these contexts, the master key is not used for cross-signing. The term 'cross-signing keys' has therefor been replaced by 'keys used for cross-signing' or, more explicitely, by 'master, user-signing and self-signing key'.
- the naming of the master key has been harmonised (no more 'master cross-signing key' or 'master signing keys'). Also the abbr. 'MSK' has been replaced by 'MK'.
- in the QR code example, the term 'cross-signing key' has been replaced by 'master key' since in mode 0x00, the current user's own master key and what the device thinks the other user's master key is used.
- it has been made more explicit that private keys used for cross-signing can be stored on the server are stored as described in the secrets module (as opposed to store them in unencrypted form)

### Pull Request Checklist


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1202,15 +1205,14 @@ The binary segment MUST be of the following form:
...
   - if `0x02`, then the current device's Ed25519 signing key
 - the second key, as 32 bytes.  The key to use depends on the mode field:
   - if `0x00`, then what the device thinks the other user's master
-    cross-signing public key is
+    public key is
   - if `0x01`, then what the device thinks the other device's Ed25519 signing
     public key is
-  - if `0x02`, then what the device thinks the user's master cross-signing public
-    key is
+  - if `0x02`, then what the device thinks the user's master public key is
```
HarHarLinks: which user is meant here?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -93,7 +93,7 @@ Example:
 ```
 
 `ed25519` and `curve25519` keys are used for [device keys](#device-keys).
-Additionally, `ed25519` keys are used for [cross-signing keys](#cross-signing).
+Additionally, `ed25519` keys are keys used for [cross-signing](#cross-signing).
```
```suggestion
Additionally, `ed25519` keys are used for [cross-signing](#cross-signing).
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -675,7 +675,7 @@ The process between Alice and Bob verifying each other would be:
 15. Assuming they match, Alice and Bob's devices each calculate Message
     Authentication Codes (MACs) for:
     * Each of the keys that they wish the other user to verify (usually their
-      device ed25519 key and their master cross-signing key).
+      device ed25519 key and their master key, see below).
```
richvdh: hrm, I am not sure I love the term "master key". In some respects, the secret storage key is more of a "master key" since it allows access to all the other keys.

Maybe we should stick with a "master signing key"?

Alternatively, we could call it "master identity key" (see also: https://github.com/matrix-org/matrix-spec-proposals/pull/4161).

content/client-server-api/modules/end_to_end_encryption.md
richvdh: In mode `0x02`, we are verifying another device belonging to the same user, so there is only one user. I think this is clear enough.

content/client-server-api/modules/end_to_end_encryption.md
Johennes: I feel similarly. If I understand the motivation correctly, the issue is that the MSK is not _only_ used for cross signing. It's still always(?) used for signing, however, so the term "master signing key" seems somewhat fitting to me. Given that "MSK" is already an established term, we might actually cause more harm than good by rebranding it "MK".

codedust: In API endpoints, the key is already named `master_key`. Hence, the current situation is super confusing (`master_key` in the API endpoints, `master key` in textual explainations, `MSK` as its abbreviation).

codedust (commented):
docs: grammar fix

content/client-server-api/modules/end_to_end_encryption.md
Johennes: Hm, that's true and changing API identifiers feels quite unfeasible. 🫤

richvdh: > In API endpoints, the key is already named `master_key`.

True, but as @Johennes says, changing that is hard. The text of the spec does not need to be constrained by the poor choices of the API.

If necessary, we can use words of the form "master signing key (sometimes known, for historical reasons, simply as `master_key`)", but to be honest I don't think that's necessary.

codedust:
Renamed to 'master key' back to 'master signing key' (commit message of the original commit needs an update now, will need to squash)

content/client-server-api/modules/end_to_end_encryption.md
codedust: I changed the terminology to back to:

> master signing key (MSK, for historical reasons sometimes known as `master_key`)

KitsuneRal:
Please don't. Renaming of master to main doesn't work here, it is _not_ the main key to sign things with.

PR #2190, opened 2025-08-06
Clarify that clients should replace events with the most recent replacement by origin_server_ts

andybalaam:
It seems clear that if the server aggregates in this order, then the client should do the same, so I argue that this change clarifies something that was already implied in the existing spec.

### Pull Request Checklist

Preview: https://pr2190--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/event_replacements.md
+++ b/content/client-server-api/modules/event_replacements.md
@@ -268,7 +268,7 @@ Client authors are reminded to take note of the requirements for [Validity of
 replacement events](#validity-of-replacement-events), and to ignore any
 invalid replacement events that are received.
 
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/event_replacements.md
```diff
@@ -268,6 +268,11 @@ Client authors are reminded to take note of the requirements for [Validity of
 replacement events](#validity-of-replacement-events), and to ignore any
 invalid replacement events that are received.
 
+Clients should render the content of the **most recent** replacement event. The
```
richvdh:
```suggestion
Clients should render the content of the **most recent** valid replacement event. The
```

richvdh (approved):
thanks!

PR #2191, opened 2025-08-06
s2s/keys: clarify minimum_valid_until_ts query

famfo:
### Pull Request Checklist

Preview: https://pr2191--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/2191.clarification
+++ b/changelogs/server_server/newsfragments/2191.clarification
@@ -1,0 +1,1 @@
+Clarify what the `minimum_valid_until_ts` field means when it is set in key queries.
--- a/data/api/server-server/keys_query.yaml
+++ b/data/api/server-server/keys_query.yaml
@@ -37,7 +37,7 @@ paths:
             A millisecond POSIX timestamp. The returned keys SHOULD be valid
             until at least this timestamp.
 
-            If not supplied, the notary server SHOULD use the current time.
+            If not supplied, the notary server MUST use the current time.
           required: false
           example: 1234567890
           schema:
@@ -101,7 +101,7 @@ paths:
                             A millisecond POSIX timestamp. The returned keys
                             SHOULD be valid until at least this timestamp.
 
-                            If not supplied, the notary server SHOULD use the
+                            If not supplied, the notary server MUST use the
                             current time.
                           example: 1234567890
               required:


data/api/server-server/keys_query.yaml
```diff
@@ -34,8 +34,8 @@ paths:
         - in: query
           name: minimum_valid_until_ts
           description: |-
-            A millisecond POSIX timestamp in milliseconds indicating when the returned
-            certificates will need to be valid until to be useful to the requesting server.
+            A millisecond POSIX timestamp indicating until when the returned
+            keys MUST at least be valid.
```
uhoreg: TBH, I personally find the existing wording easier to follow and the proposed wording sounds awkward to me.  Aside from the redundancy regarding milliseconds, and calling it certificates instead of keys, I don't see any problem with the existing wording.  But if you think it's unclear, maybe something like:

> A millisecond POSIX timestamp.  The returned keys must be valid at least until this time to be useful to the requesting server.

famfo: That proposal is also more understandable, it would be nice to actually use RFC2119 keywords after they have been introduced at the beginning of the spec ^^

Not sure if the "to be useful to the requesting server" is adding anything useful though, it is kinda unnecessary and already implied in my opinion.

> A millisecond POSIX timestamp. The returned keys MUST at least be valid until this timestamp.

uhoreg: I believe the intention here is not to say that the returned key MUST be valid until this timestamp in the RFC2119 sense, but rather that the parameter expresses the minimum time that the requesting server wants the keys to be valid for.

That is, the intention here is to explain the meaning of the parameter, rather than to set a requirement for the response.  The current wording doesn't set a requirement on the response, so it would be a spec change to make it into a requirement.

famfo: The interpretation of the existing parameter could go either way here, exactly because defined language is not used (which is unfortunately all over in this spec).

Not having a requirement on the response for query parameters is... definitely unexpected reading a definition for said query parameters. Something along the lines of "The returned keys MAY be valid until at least this timestamp." would be more in line with this interpretation.

In my opinion, having such a parameter is kind of pointless if it does not impose any requirements (which, I agree, removing it would be a spec change. I don't see how requiring the parameter to have an effect would require a spec change though).

changelogs/server_server/newsfragments/2191.clarification
```diff
@@ -0,0 +1 @@
+Clarify what the minimum_valid_until_ts field means when it is set in key queries.
```
anoadragon453:
```suggestion
Clarify what the `minimum_valid_until_ts` field means when it is set in key queries.
```

data/api/server-server/keys_query.yaml
anoadragon453: I think @famfo's updated wording is much clearer, but @uhoreg is right in that this technically is a change to the spec.

> The interpretation of the existing parameter could go either way here, exactly because defined language is not used

This is the crux of the issue. The language of the spec *could* be interpreted as "The returned keys MAY be valid until at least this timestamp", and thus a "MUST" is a change.

To move forwards here, I suggest:

* This PR change the wording to be MAY instead of MUST, so that this becomes a clarification.
* A (very small) MSC be opened to update the wording to MUST.

...as I agree that I don't see a need for homeservers to ask the remote for keys that must be valid after a certain date, only for the remote to return keys that aren't.

uhoreg: Yes, the current wording is kind of strange.  It implies that something is expected from the responding server, but doesn't actually spell it out as a requirement, and doesn't say what should happen if the responding server is unable to meet that requirement.

Looking at the [synapse code for responding to this request](https://github.com/element-hq/synapse/blob/develop/synapse/rest/key/v2/remote_key_resource.py), it looks like synapse is using this parameter as a hint for whether it can use a cached value, or whether it should re-fetch the keys: if the cached value is new enough to satisfy the `minimum_valid_until_ts`, then it will just send the cached value, no matter how old it is.  Otherwise, it will try to re-fetch the keys from the remote server.  But unless I'm missing something, I don't see any attempt to actually filter out values that are too old.  If the only value that the server can return is too old, AFAICT, it will still return it.

anoadragon453: In that case, I suggest we clarify this as "The returned keys MAY be valid until at least this timestamp" and file a separate issue to discuss whether this should be changed to a MUST.

Today, someone implementing the spec may misinterpret the vague wording and expect the wrong thing from i.e. Synapse.

uhoreg: I'm not sure that "The returned keys MAY be valid until at least this timestamp" is helpful wording.  If I'm reading the spec, I want to know, as a requester, what value I should put in that parameter and why; or as a responder, what I should be doing when I see a value in that parameter.  The current wording tries to communicate the former (you stick the timestamp in there that you want the key to be valid until, because older keys aren't useful to you), but doesn't to a great job of the latter.  "The returned keys MAY be valid until at least this timestamp" doesn't really communicate either thing.

So if we're taking synapse's implementation as an example what it's supposed to be doing, I might use something similar to the current wording, but add something along the lines of "The responding server can use this and a hint to refresh any keys that it has cached, if they are older than the requested timestamp."

uhoreg: FWIW, I don't think there's much value in making an MSC to change the spec so that it's a MUST.  I agree that it's pointless for the responding server to send keys that are useless to the requesting server.  But it's also easy enough to filter out the useless values on the requesting server.  And given that the existing behaviour seems to be to include the useless keys, the requesting servers would have to handle those values anyways, or we'd need to stick a new version on the endpoint.  Which all seems like a lot of work for not much value.

I could, on the other hand, see a case for making it a SHOULD, and I could also see that an argument could be made that making it a SHOULD wouldn't require an MSC.  But the wording should still answer the questions of "why would the requester set the parameter to any given value", and "what should the responder do with the value" in some way or other.

famfo: I am not sure what the aim of the spec is at this point. Either an implementable and understandable spec (which would require some MSC to either remove this endpoint or change it to return up to date keys or an error if there are none in my opinion) or a spec with a bunch of loosely defined endpoints (which is unfortunately the status quo, especially with some of the older endpoints which haven't been touched and presumably implemented in a while).

In my opinion the spec should be clear and the two options are:
1. an MSC to change this endpoint to MUST return valid keys
2. an MSC to remove this endpoint because other servers can't rely on a notary server to return up to date information anyways

Fwiw, Synapse is also not the reference implementation anymore, it should follow the spec, not the other way around.

uhoreg: Synapse is not the reference implementation, but this endpoint comes from a time when Synapse was the reference implementation.  For such cases, Synapse's implementation is one of the things that we consider when deciding what to do if the spec and Synapse differ.

anoadragon453: I internally discussed the practice of evaluating ambiguous spec that was conceived before the MSC process existed with the SCT. The conclusion is that it comes down to a case-by-case basis. Namely, what is the ecosystem currently doing, and does the spec reflect that?

@uhoreg mentioned above what Synapse currently does (use it as a cache threshold, return keys even if they are old), tuwunel [doesn't appear to implement the endpoint](https://github.com/matrix-construct/tuwunel/blob/af570c481e81aad205e9b84dadda0191962fa67e/src/api/server/key.rs).

Dendrite [will return the keys from the cache](https://github.com/element-hq/dendrite/blob/a042861df511c8f81e2cbb2bace09eb0c3e94d20/federationapi/internal/query.go#L80-L89) if they satisfy `minimum_valid_until_ts`, otherwise it will fetch updated keys directly. It won't filter it out if the key is too old though. Thus, the same as what Synapse is doing.

When making the request, tuwunel will set `minimum_valid_until_ts` to [1 hour from now](https://github.com/matrix-construct/tuwunel/blob/af570c481e81aad205e9b84dadda0191962fa67e/src/service/server_keys/mod.rs#L185-L186).

Synapse will set `minimum_valid_until_ts` to an event's `origin_server_ts` when [validating said event](https://github.com/element-hq/synapse/blob/develop/synapse/federation/federation_base.py#L288), and the current time when [validating a remote federation request](https://github.com/element-hq/synapse/blob/41938d6fd2a2690a6ca5c884a787f819907d1255/synapse/federation/transport/server/_base.py#L135).

---

Given all this, I think it's fine to forgo the MSC in this instance and clearly document what homeservers are currently doing:

1. Remote homeservers SHOULD return a key whose expiry time is the same as or after the given `minimum_valid_until_ts`, yet MAY return a key with an expiry time that is older.
2. Requesting homeservers SHOULD set `minimum_valid_until_ts` to an event's `origin_server_ts` when validating an event, or the current time when validating an incoming, signed request from another homeserver over federation.

@famfo 

> or an error if there are none in my opinion

This would be clearer, and if a homeserver really wanted expired keys, they could set this value to `0`. However, no homeserver does that today. So I believe an endpoint bump (and an MSC) would be required to make this change.

famfo: > Requesting homeservers SHOULD set minimum_valid_until_ts to an event's origin_server_ts when validating an event

I think _verification_ of events is a bit out of scope for something trying to clarify how long the returned keys by notary servers should be valid for, but I will note the rest to the observed behavior.

data/api/server-server/keys_query.yaml
```diff
@@ -34,10 +34,10 @@ paths:
         - in: query
           name: minimum_valid_until_ts
           description: |-
-            A millisecond POSIX timestamp in milliseconds indicating when the returned
-            certificates will need to be valid until to be useful to the requesting server.
+            A millisecond POSIX timestamp. The returned keys SHOULD be valid
+            until at least this timestamp.
 
-            If not supplied, the current time as determined by the notary server is used.
+            If not supplied, the notary server SHOULD use the current time.
```
uhoreg: I'm not sure why you changed this line (and the occurrence below).  It's pretty standard wording in the spec for optional parameters to say something along the lines of "If not supplied, [some value] is used."  Changing this to a "SHOULD" actually weakens the assertion, because it currently states what the default value is to be (effectively, as "MUST"), rather than being just a suggestion.

If you don't like the current wording, maybe something like "Defaults to the current time as determined by the notary server" would read better, without changing the meaning.

famfo: I took this from @anoadragon453's comment

> Requesting homeservers SHOULD set minimum_valid_until_ts to an event's origin_server_ts when validating an event, or the current time when validating an incoming, signed request from another homeserver over federation.

I'd be happy to change this to a MUST though to also use standardized language.

anoadragon453: The above was making a suggestion for what the requesting (non-notary) homeserver should set `minimum_valid_until_ts` to.

> or the current time when validating an incoming, signed request from another homeserver over federation.

This referred to the requesting homeserver validating a request from yet another, unrelated homeserver.

> If not supplied, the current time as determined by the notary server is used.

...talks about what the notary server will default to if the parameter isn't given, which is separate.

I agree that it would be "MUST" when converting this to RFC2119 keywords; which I'm happy to see more of.

PR #2207, opened 2025-09-08
Specification for MSC4311: Create event availability in stripped state

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4311

Reviewable(ish) commit-by-commit.

**Note**: In line with existing spec, this change reminds readers *constantly* about the event format being different depending on room version.

### Pull Request Checklist

Preview: https://pr2207--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -2823,7 +2823,7 @@ the following places:
 * [`invite_state`](#get_matrixclientv3sync_response-200_invited-room) and
   [`knock_state`](#get_matrixclientv3sync_response-200_knocked-room) on
   [`/sync`](#get_matrixclientv3sync) responses.
-* When present on [`m.room.member`](#mroommember) events, the `invite_room_state`
+* On [`m.room.member`](#mroommember) events, the `invite_room_state`
   and `knock_room_state` under `unsigned` on the event.
 
 {{% boxes/note %}}
--- a/data/api/server-server/invites-v1.yaml
+++ b/data/api/server-server/invites-v1.yaml
@@ -97,8 +97,6 @@ paths:
                               Note that events have a different format depending on the room
                               version - check the [room version specification](/rooms) for
                               precise event formats.
-                          example:
-                            $ref: ./examples/invite_or_knock_state.json
               type: object
         required: true
       responses:
--- a/data/api/server-server/invites-v2.yaml
+++ b/data/api/server-server/invites-v2.yaml
@@ -98,8 +98,6 @@ paths:
                       Note that events have a different format depending on the room
                       version - check the [room version specification](/rooms) for
                       precise event formats.
-                  example:
-                    $ref: ./examples/invite_or_knock_state.json
               required:
                 - room_version
                 - event
--- a/data/api/server-server/knocks.yaml
+++ b/data/api/server-server/knocks.yaml
@@ -294,6 +294,7 @@ paths:
                     type: array
                     items:
                       type: object
+                      title: PDU
                       properties: {}
                       description: |-
                         Note that events have a different format depending on the room


data/api/server-server/invites-v1.yaml
```diff
@@ -71,13 +71,33 @@ paths:
...
+                            version - check the [room version specification](/rooms) for
+                            precise event formats.
                           items:
-                            $ref: ../../event-schemas/schema/core-event-schema/stripped_state.yaml
+                            type: object
+                            properties: {}
+                            description: |-
+                              Note that events have a different format depending on the room
+                              version - check the [room version specification](/rooms) for
+                              precise event formats.
```
turt2live: reviewer: we don't appear to have a useful type for this, so I've copied what transactions do.

content/client-server-api/_index.md
```diff
@@ -2810,7 +2810,7 @@ fresh state can be acquired from a join.
 Stripped state should contain some or all of the following state events, which
 should be represented as stripped state events when possible:
 
-* [`m.room.create`](#mroomcreate)
+* [`m.room.create`](#mroomcreate) ({{% changed-in v="1.16" %}} required on invites and knocks)
```
richvdh: I don't really know what "required on invites and knocks" means here. Something to do with `unsigned/invite_room_state` in invite events? And something equivalent for knocks?

It might be easier to specify when the `m.room.create` is *not* required?

richvdh: for readability, can we move the `changed-in` to the bottom of the section, to be something like:


```
{{% changed-in v="1.16" %}}: `m.room.create` is now **required** in some situations.
```

content/client-server-api/_index.md
```diff
@@ -2810,7 +2810,7 @@ fresh state can be acquired from a join.
 Stripped state should contain some or all of the following state events, which
 should be represented as stripped state events when possible:
```
richvdh: During MSC, we established that it was important that servers do *not* simply pass through what they receive over federation, but actively strip out any keys other than `sender`, `type`, `state_key` and `content`.

AFAIK Synapse does *not* do any stripping today. My point being, that this seems like an easy thing to overlook, and I think it is worth calling out as a note in the spec.

data/api/server-server/examples/stripped_state.json
richvdh, line 1: Can we please avoid calling this stripped state? The thing that makes stripped state stripped state is that it lacks properties other than `type`, `sender`, `content` and `state_key` -- which doesn't apply here.

data/api/server-server/invites-v1.yaml
```diff
@@ -71,9 +71,20 @@ paths:
                       properties:
                         invite_room_state:
                           type: array
+                          x-changedInMatrixVersion:
+                            "1.16": |
+                              `m.room.create` and format requirements were added.
```
richvdh, line 76: (I am belatedly wishing that we had just deprecated `/v1/invite` and mandated that everyone uses `/v2`, rather than adding more changes to `/v1`.)

Do you think it's worth trying to do some $ref trickery to reduce the duplication in the yaml, at least?

data/api/server-server/invites-v1.yaml
richvdh: transactions give the type a name of `PDU`, which I suggest you do here too, rather than just accepting the default of `object`.

richvdh (changes requested):

data/api/server-server/invites-v1.yaml
turt2live: I've been bitten by `$ref` not working in weird and unpredictable ways in the past, so I'm incentivized to copy/paste. The endpoints are also *slightly* different in error responses, which makes it especially annoying. 

If given a choice, I'd rather not fight `$ref` more than I already have to.

data/api/server-server/examples/stripped_state.json
turt2live: Suggestions on what to call this are very much needed. It's also not room state because it's incomplete, and it's used in different places so it can't be called "invite state" anymore either. 

Other candidates while writing this PR were "partial state" (already used elsewhere in context of joins in Synapse), "incomplete state" (doesn't feel descriptive enough), and "assistive state" (who is it assistive to?) - "stripped state" felt like the least worst option, but I'm open to opinions.

content/client-server-api/_index.md
turt2live: will add a note/warning box

content/client-server-api/_index.md
turt2live: This is mostly an awkward artifact of the paragraphs prior: stripped state is only used in invites and knocks, but the section states that it can be used in "other situations" too. We don't currently leverage those "other situations". The note is mostly trying to say "the SHOULD is a MUST here for 2 specific cases".

So, it's not required for features that don't exist, which is more awkward.

turt2live: sure, though I'm not seeing a concern about readability: is it that the list becomes a sentence?

turt2live: I've made an attempt to clarify it here: a678f0b52b7203932bb47a4aeac69763caac29ba

data/api/server-server/examples/stripped_state.json
turt2live: I've called it "invite or knock state" now. If something else is better/clearer, let me know.

86358e7e1749f653403dab42482e5da0dfd0ea40

content/client-server-api/_index.md
```diff
@@ -2810,14 +2810,22 @@ fresh state can be acquired from a join.
...
 * [`m.room.join_rules`](#mroomjoin_rules)
 * [`m.room.canonical_alias`](#mroomcanonical_alias)
 * [`m.room.encryption`](#mroomencryption)
 
+{{% changed-in v="1.16" %}} The `m.room.create` event is now **required** in
+the following places:
+* [`invite_state`](#get_matrixclientv3sync_response-200_invited-room) and
+  [`knock_state`](#get_matrixclientv3sync_response-200_knocked-room) on
+  [`/sync`](#get_matrixclientv3sync) responses.
+* When present on [`m.room.member`](#mroommember) events, the `invite_room_state`
```
richvdh: the "when present" here is hard to grok.

"The `m.room.create` event is now required... when present". Sounds a bit circular.

I'm guessing you don't mean that it's the `m.room.create` event that may or may not be present, but please clarify.

content/client-server-api/_index.md
richvdh: the concern was:

(a) that the list was confusing because it was a list of events that *may* be present and then some of them were required and some of them were not.
(b) it mixed up what is true *now* with insights about how things have changed.

The first of those is now fixed. The second... kinda. I'd still prefer we clearly state what is true *now* and then maybe have some text at the bottom of the section hinting that there has been a change in the spec at some point in the past. As an example of what I'm aiming for, see https://docs.python.org/3/library/stdtypes.html#dict, which says in the text "Dictionaries preserve insertion order" and explains exactly what that means, and then at the bottom has a short "Changed in version 3.7: Dictionary order is guaranteed to be insertion order."

But what you have is fine.

data/api/server-server/invites-v1.yaml
richvdh: I think that @zecakeh has done a bunch of work to make `$ref` work better, so it might be worth taking another look.

> The endpoints are also slightly different in error responses, which makes it especially annoying.

Ok, but you don't have to duplicate the whole endpoint -- you can just share some parameter type definitions, for example.

I'm really not a big fan of having two copies of all this stuff - it makes it much harder to maintain and review. Doing this review was particularly annoying because I kept getting confused about which bits I'd seen and which I hadn't.  Making the job harder for the reviewer to save a bit of authoring time is not a great investment.

I'll not insist on it this time around though.

data/api/server-server/knocks.yaml
```diff
@@ -293,19 +293,40 @@ paths:
                   knock_room_state:
                     type: array
                     items:
-                      $ref: ../../event-schemas/schema/core-event-schema/stripped_state.yaml
+                      type: object
```
richvdh, line 296: needs `title: PDU`

data/api/server-server/invites-v1.yaml
```diff
@@ -71,13 +71,34 @@ paths:
...
+                            type: object
+                            title: PDU
+                            properties: {}
+                            description: |-
+                              Note that events have a different format depending on the room
+                              version - check the [room version specification](/rooms) for
+                              precise event formats.
                           example:
-                            $ref: ../../event-schemas/examples/invite_room_state.json
+                            $ref: ./examples/invite_or_knock_state.json
```
richvdh: This example isn't actually used in the spec (since we have an example for the response as a whole) and it might be worth removing it for simplicity/clarity.

data/api/server-server/invites-v2.yaml
```diff
@@ -72,13 +72,34 @@ paths:
...
+                    type: object
+                    title: PDU
+                    properties: {}
+                    description: |-
+                      Note that events have a different format depending on the room
+                      version - check the [room version specification](/rooms) for
+                      precise event formats.
                   example:
-                    $ref: ../../event-schemas/examples/invite_room_state.json
+                    $ref: ./examples/invite_or_knock_state.json
```
richvdh: ditto

content/client-server-api/_index.md
turt2live: Trying to clarify that the `*_room_state` fields are the conditionally present ones, but maybe it's best to just ignore that detail here and rely on docs through the links.

data/api/server-server/invites-v1.yaml
turt2live: The concern with the error responses is that it's harder to deduplicate that part is all - the request parameters can be deduplicated, but that's only so helpful. 

I'll experiment with `$ref` separate from this, and open a PR if it works.

data/api/server-server/invites-v1.yaml
turt2live: I'm on the fence about whether to remove it - on one hand, it's indeed not used, but on the other it may be useful to consumers of the OpenAPI? I think for now I'll keep it maintained to just make a decision. We really should go through and clean up our schema files one day though - I imagine there are plenty of other examples of this.

turt2live: actually, let's remove it. The response example is now a reference itself.

PR #2210, opened 2025-09-16
Spec PR - MSC4142: Remove unintentional intentional mentions in replies

Johennes:
Spec for https://github.com/matrix-org/matrix-spec-proposals/pull/4142.

### Pull Request Checklist

Preview: https://pr2210--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/rich_replies.md
+++ b/content/client-server-api/modules/rich_replies.md
@@ -64,6 +64,11 @@ as if it were to be under `m.relates_to` directly instead.
 
 #### Mentioning the replied to user
 
+{{% boxes/note %}}
+{{% changed-in v="1.16" %}}
+Clients SHOULD no longer propagate mentioned users in the replied to event.
+{{% /boxes/note %}}
+
 In order to notify users of the reply, it MAY be desirable to include the `sender`
 of the replied to event. See [user and room mentions](#user-and-room-mentions) for
 additional information.


content/client-server-api/modules/rich_replies.md
```diff
@@ -64,11 +64,11 @@ as if it were to be under `m.relates_to` directly instead.
 
 #### Mentioning the replied to user
```
clokep, line 65: Should this have a note that we changed something in 1.16?

Johennes: Hm, possibly but I'm struggling a bit to formulate it. The part we removed was a MAY. So clients can in fact still do what we removed in 1.16. It also didn't have a RECOMMENDED label. So we cannot say we removed a recommendation either. 🤔

clokep: Maybe:

> Clients SHOULD no longer propagate mentioned users in the replied to event.

Johennes: Ok, have added that.

clokep: @tulir Does that make sense to you?

tulir: Yeah that sounds reasonable

tulir (approved):

PR #2212, opened 2025-09-17
Add example to each endpoint when the capability is not available

hughns:
### Pull Request Checklist

Preview: https://pr2212--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2212.clarification
+++ b/changelogs/client_server/newsfragments/2212.clarification
@@ -1,1 +1,1 @@
-Add note to each endpoint that uses capability negotiation and document expected response when the capability is not available.
+Add example to each endpoint when the capability is not available.


data/api/client-server/profile.yaml
```diff
@@ -29,6 +29,11 @@ paths:
         Servers MAY reject `null` values. Servers that accept `null` values SHOULD store
         them rather than treating `null` as a deletion request. Clients that want to delete a
         field, including its key and value, SHOULD use the `DELETE` endpoint instead.
+
+        This endpoint uses [capabilities negotiation](/client-server-api/#capabilities-negotiation)
+        depending on the `keyName`.
+        For the `displayname` key, clients SHOULD check the value of the [`m.set_displayname` capability](/client-server-api/#mset_displayname-capability).
+        For the `avatar_url` key, clients SHOULD check the value of the [`m.set_avatar_url` capability](/client-server-api/#mset_avatar_url-capability).
```
Johennes: These capabilities have been deprecated in the recent 1.16 release and are now superseded by `m.profile_fields`.

```suggestion
        This endpoint uses [capabilities negotiation](/client-server-api/#capabilities-negotiation)
        depending on the `keyName`. Clients SHOULD check the value of the
        [`m.profile_fields` capability](/client-server-api/#mprofile_fields-capability) to detect
        which `keyName`s they are allowed to modify.
```

data/api/client-server/administrative_contact.yaml
```diff
@@ -176,6 +180,18 @@ paths:
                   value: {
                     "submit_url": "https://example.org/path/to/submitToken"
                   }
+        "400":
```
anoadragon453, line 179: Where do these status codes choices come from?

hughns: The original MSC3283 did not specify the expected response code and referred to the Synapse implementation. So, I took these values from the Synapse implementation.

Another approach to the clarification would be to specify a range. e.g. 400-499

data/api/client-server/password_management.yaml
```diff
@@ -82,6 +86,18 @@ paths:
             application/json:
               schema:
                 $ref: definitions/auth_response.yaml
+        "403":
```
hughns, line 85: This code came from approach described at https://github.com/matrix-org/matrix-spec/pull/2212/files#r2378179370

data/api/client-server/profile.yaml
```diff
@@ -116,6 +121,12 @@ paths:
                       "errcode": "M_INVALID_PARAM",
                       "error": "Invalid profile key.",
                     }
+                capability_disabled:
```
hughns, line 119: Again, this code was based on the Synapse implementation.

data/api/client-server/login_token.yaml
```diff
@@ -110,6 +110,18 @@ paths:
             application/json:
               schema:
                 $ref: definitions/auth_response.yaml
+        "404":
```
hughns, line 113: I'm re-reading MSC3882 now (which I happened to author). The suggested 404 here is based on Synapse implementation.

An alternative would be:

- 404 M_UNRECOGNIZED = the server does not implement the optional login token feature
- 400 (or 403) M_FORBIDDEN = the user is not allowed to do this (i.e. capability would return false)

hughns:
I've reduced the scope of this and separated out the changes which purely add a note that capability negotiation exists for an endpoint (into https://github.com/matrix-org/matrix-spec/pull/2223).

hughns:
> Could you write up a quick MSC detailing each of these new error responses (status code and error codes)?

I've started this in MSC4369.

> And do you have a sense of whether introducing a new `M_CAPABILITY_NOT_ENABLED` error code instead of using the existing ones would break lots of functionality?

Not off the top of my head, but I will consider in the MSC.

Issue #2098, opened 2025-03-17
`room_id` should be required in `GET /events` for peeking


zecakeh:
**Link to problem area**: https://spec.matrix.org/v1.13/client-server-api/#peeking_get_matrixclientv3events

**Issue**

If the request doesn't include a `room_id` it cannot work for peeking because the homeserver wouldn't know which room to watch.

In Synapse, the `room_id` [is required on this endpoint for guests](https://github.com/element-hq/synapse/blob/59a15da43364c6b3319983e68f686750a918ec9a/synapse/rest/client/events.py#L57-L59). If the user is not a guest, [this logic](https://github.com/element-hq/synapse/blob/59a15da43364c6b3319983e68f686750a918ec9a/synapse/notifier.py#L820-L830) is used to determine the room(s) to watch: if there is no explicit `room_id` the rooms that are joined by the user are watched, which is the behavior of the non-peeking version of the endpoint.

Construct only supports the peeking version of this endpoint and [requires a room ID](https://github.com/matrix-construct/construct/blob/0624b69246878da592d3f5c2c3737ad0b5ff6277/modules/client/events.cc#L112-L116).

PR #2216, opened 2025-09-26
room_id is required for peeking

Johennes:
Fixes: #2098

As per @zecakeh's excellent research in #2098.

### Pull Request Checklist

Preview: https://pr2216--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2216.clarification
+++ b/changelogs/client_server/newsfragments/2216.clarification
@@ -1,0 +1,1 @@
+`room_id` is required for peeking via `/_matrix/client/v3/events`.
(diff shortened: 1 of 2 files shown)


anoadragon453 (approved):
I welcome this change. Yet even afterwards, it's confusing that we have two versions of the same endpoint method and path, where the only difference is that one has a `room_id` parameter, and the other does not.

Seems they should be combined, and only the ability to not provide the `room_id` endpoint be deprecated. Something to solve in a separate PR though.

Johennes:
> Seems they should be combined, and only the ability to not provide the room_id endpoint be deprecated. Something to solve in a separate PR though.

Given that these endpoints have been deprecated since r0, maybe we could remove them by now?

PR #2221, opened 2025-10-03
Spec device ID masquerading

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4326

**Cannot be merged until MSC passes FCP** (but can be reviewed).

### DIFF
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -356,6 +356,7 @@ service would like to masquerade as.
 Inputs:
 -   Application service token (`as_token`)
 -   User ID in the AS namespace to act as.
+-   Device ID belonging to the User ID to act with.
 
 Notes:
 -   This applies to all aspects of the Client-Server API, except for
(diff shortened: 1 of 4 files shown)


content/application-service-api.md
```diff
@@ -375,9 +376,19 @@ service's `user` namespaces. If the parameter is missing, the homeserver
...
 
+{{% added-in v="1.17" %}} Application services MAY similarly masquerade
+as a specific device ID belonging the user ID through use of the `device_id`
+query string parameter on the request. If the given device ID is not known
+to belong to the user, the server will return a 400 `M_UNKNOWN_DEVICE` error.
+If no `user_id` is supplied, the `device_id` MUST belong to the user implied
+by the `sender_localpart` property of the application service's registration.
+If no `device_id` is supplied, the homeserver is to assume the request is
+being made without a device ID and will fail to complete operations which
+require a device ID (such as uploading one-time keys).
```
onestacked, line 387: Should we also specify what happens when this is used on requests where the device id is not used?

turt2live: That should be specified implicitly: the request continues as normal.

onestacked: From how I read the current text the server would have to check the device id even when it is unneeded.
It should probably include a sentence to explicitly allow the server to not check the device id in this case.

Something like:

For requests that do not use the device id the server MAY omit to verify the `device_id` argument.

tulir: If a device ID is passed, it should be verified regardless of whether the endpoint needs it

onestacked: What is the point of that, wouldn't that just be a (mostly) useless database query on the server?

turt2live: Having re-read the MSC and implementation, this PR is wrong anyhow: the device ID is always validated if supplied.

turt2live: .. which is what the PR already says. I don't know what I read between my tabs there, but it was wrong.

turt2live: > What is the point of that, wouldn't that just be a (mostly) useless database query on the server?

The purpose of the MSC was to mirror the `user_id` validation rather than try to define new semantics. A future MSC can remove the requirement to validate the device ID, though it may be valuable to keep around to maintain the "last seen IP" information within a homeserver.

onestacked: I guess its fine if this mirrors the user id behavior.
I still thing it might be good to allow not checking for lightweight servers, but its probably not that much of a difference anyways.

Issue #1543, opened 2023-05-25
Olm and Megolm should be specced in the actual spec


dkasak:
Rather than just on gitlab.matrix.org[[1]](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/docs/olm.md)[[2]](https://gitlab.matrix.org/matrix-org/olm/-/blob/master/docs/megolm.md).

~Additionally, the spec [currently has a broken link](https://spec.matrix.org/v1.6/client-server-api/#mmegolmv1aes-sha2) for the Megolm spec.~ duplicate of https://github.com/matrix-org/matrix-spec/issues/1555

Johennes:
I put up a draft in https://github.com/matrix-org/matrix-spec/pull/2226 which has a few loose ends that we'll need to decide on.

clokep:
> That said: it's not a big problem, I'm not sure anyone else has ever used Olm or Megolm (I thought maybe OMEMO, but that spec refers to Signal's specs rather than ours), and I certainly agree that putting them on spec.matrix.org makes sense. So, fine.

Version 0.2 and earlier of OMEMO referred to Olm instead of Signal: https://xmpp.org/extensions/xep-0384.html#revision-history-v0.2

Johennes:
> (Also: I just noticed that https://spec.matrix.org/unstable/client-server-api/#molmv1curve25519-aes-sha2 and https://spec.matrix.org/unstable/client-server-api/#mmegolmv1aes-sha2 still link to the external spec, which should be fixed)

Sorry, I missed these. Have opened https://github.com/matrix-org/matrix-spec/pull/2242.

turt2live:
This appears to have shipped.

PR #2226, opened 2025-10-08
Include Olm & Megolm specifications in main spec

Johennes:
This inlines the Olm and Megolm specs into Matrix spec as suggested in #1543. It does this by:

- Adding a new top-level section
- Defining `\[ ... \]` and `\( ... \)` as passthrough delimiters for block and inline math expressions
- Calling [`transform.ToMath`](https://gohugo.io/functions/transform/tomath/) in a passthrough render hook to convert Latex to MathML with additional Katex styling
- Bundling the https://github.com/KaTeX/KaTeX CSS and fonts with the spec to avoid loading external resources
- Allowing `unsafe-inline` as `style-src` in CSP. This might be a deal breaker but I couldn't find another way around it so far. `transform.ToMath` will add a myriad of inline styles for vertical positioning without which many formulas are illegible.
- Adding `nohighlight` as language on any untyped code block. This is necessary because Docsy itself will otherwise enable its own Katex integration which is not CSS-only.

Things to figure out:

- Whether we want to do this at all
  - Best discussed in #1543.
- Placement within the spec
  - Top-level chapter – this is what the PR currently does
  - Appendix subsection
  - Entirely separate page (might require extracting a common theme)
  - ?
- How to best enable rendering of math expressions without CSP issues
  - [`transform.ToMath`](https://gohugo.io/functions/transform/tomath/) – this is what the PR currently does while relaxing CSP rules
  - [Rendering math on the client-side using JavaScript](https://gohugo.io/content-management/mathematics/)
  - Pure [MathML](https://developer.mozilla.org/en-US/docs/Web/MathML) (would probably be rather hard to maintain)
  - ?

Fixes: #1543

### Pull Request Checklist

Preview: https://pr2226--matrix-spec-previews.netlify.app

### DIFF
--- a/scripts/download-katex-assets.sh
+++ b/scripts/download-katex-assets.sh
@@ -36,6 +36,6 @@ install -vm644 "$tmp_dir/katex/katex.min.css" "$root/static/css/katex.min.css"
 
 # Remove any existing fonts and move the new ones into place.
 rm -rvf "$root"/static/css/fonts/KaTeX*
-while read -r file; do
+while IFS= read -r -d '' file; do
     install -vm644 "$file" "$root/static/css/fonts"
-done < <(find "$tmp_dir/katex/fonts" -maxdepth 1 -name "KaTeX*.woff2")
+done < <(find "$tmp_dir/katex/fonts" -maxdepth 1 -name "KaTeX*.woff2" -print0)


config/_default/hugo.toml
```diff
@@ -121,7 +127,7 @@ sidebar_menu_compact = true
 [[server.headers]]
   for = '/**'
   [server.headers.values]
-    Content-Security-Policy = "default-src 'self'; style-src 'self'; script-src 'self'; img-src 'self' data:; connect-src 'self'; font-src 'self' data:; media-src 'self'; child-src 'self'; form-action 'self'; object-src 'self'"
+    Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; img-src 'self' data:; connect-src 'self'; font-src 'self' data:; media-src 'self'; child-src 'self'; form-action 'self'; object-src 'self'"
```
richvdh, line 135: As the PR description says:

> * Allowing `unsafe-inline` as `style-src` in CSP. This might be a deal breaker but I couldn't find another way around it so far. `transform.ToMath` will add a myriad of inline styles for vertical positioning without which many formulas are illegible.

@dbkr and I scratched our heads for a bit about what was so bad about inline `style` attributes: [Stack Overflow to the rescue](https://stackoverflow.com/a/31759553/637864).

It's obviously not ideal, but given the spec is essentially statically-generated HTML, the opportunity for cross-site styling attacks seems minimal, and the other solutions seem worse.

I opened https://github.com/KaTeX/KaTeX/issues/4096.

Let's add a comment, though:

```suggestion
    # `style-src 'unsafe-inline'` is needed to correctly render the maths in the Olm spec:
    # https://github.com/KaTeX/KaTeX/issues/4096
    Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; img-src 'self' data:; connect-src 'self'; font-src 'self' data:; media-src 'self'; child-src 'self'; form-action 'self'; object-src 'self'"
```

config/_default/hugo.toml
```diff
@@ -43,6 +43,12 @@ description = "Home of the Matrix specification for decentralised communication"
     [markup.goldmark.renderer]
       # Enables us to render raw HTML
       unsafe = true
+    [markup.goldmark.extensions]
+      [markup.goldmark.extensions.passthrough]
```
richvdh, line 50:
```suggestion
      # Tell Goldmark to pass delimited blocks through the `render-passthrough` render hook.
      # This is used to render the maths in the Olm spec.
      # See: https://gohugo.io/functions/transform/tomath/#step-1.
      [markup.goldmark.extensions.passthrough]
```

content/olm-megolm/_index.md
```diff
@@ -0,0 +1,711 @@
+---
+title: "Olm & Megolm"
```
richvdh, line 2: I think this might work better as two separate pages (both under the "Olm & Megolm" table-of-contents entry)?

richvdh, line 2: Also: don't forget to add this to the list of "parts of the specification" at https://pr2226--matrix-spec-previews.netlify.app/#matrix-apis

layouts/_markup/render-passthrough.html
```diff
@@ -0,0 +1,9 @@
+{{- $opts := dict "output" "htmlAndMathml" "displayMode" (eq .Type "block") }}
```
richvdh, line 11: Could you add a comment, similar to that used for the "render-headings" hook?

```
{{- /*

    This is a passthrough render hook (https://gohugo.io/render-hooks/passthrough/).

    We use it to send the delemited passthrough element through KaTeX; this is
    used to render the maths in the Olm spec.

    See: https://gohugo.io/functions/transform/tomath/#step-2

*/ -}}
```

static/css/fonts/KaTeX_AMS-Regular.woff2
richvdh, line 1: Could you add a script or something that will download these font files, and the CSS?

For reference, there is some similar stuff for the "Inter" font in https://github.com/matrix-org/matrix-spec/tree/main/assets/css/fonts.

richvdh (commented):
looks great, other than a few comments

content/olm-megolm/_index.md
Johennes: Sounds good. Shall we put anything on the index page? Or should it just redirect to the Olm subpage?

Johennes: Oops, I missed that list. 🤦‍♂️

Added an entry with 4a0d7c5a509156aecf64c1556ced4fd0cb67ca9b.

layouts/_markup/render-passthrough.html
Johennes: Have added a minimally adapted variant of the comment in d0929ef4f05976b656ba932e8d417385a7100c20.

static/css/fonts/KaTeX_AMS-Regular.woff2
Johennes: Have added a very minimal `./scripts/download-katex-assets.sh` that can be invoked with the KaTeX version (current assets are from v0.16.23) in 8f3769af3049c1f494856003156a844c698e8bc3. Let me know if you had something more sophisticated in mind.

content/olm-megolm/_index.md
Johennes: Have introduced subpages and a very minimal index page in 7fa012b67fbac9803e96d2a6fe3928c83a9bb2e3. Let me know if this is ok.

Edit: On second thought the text on the index page sounds sort of silly. I don't really know what to write there to be honest. Maybe we should just redirect to the first subpage like we do in the changelog section.

content/olm-megolm/_index.md
```diff
@@ -4,708 +4,9 @@ weight: 61
 type: docs
 ---
 
-## Olm: A Cryptographic Ratchet
+The Olm and Megolm cryptographic ratchets are not formally part
+of the Matrix specification. In lack of a better place, their
+specification is reproduced here, regardless.
```
richvdh: How about just:

```suggestion
Matrix uses the Olm and Megolm cryptographic ratchets for [end-to-end encryption](../client-server-api/#end-to-end-encryption).
```

Honestly, just the page links are all we need, really.

static/css/fonts/KaTeX_AMS-Regular.woff2
richvdh: nope, this is fine, thanks!

scripts/download-katex-assets.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
```
richvdh: Let's give this a comment saying what it does.
```suggestion
#!/bin/bash
#
# Download the KaTeX fonts and CSS, and copy them into `static`.
```

scripts/download-katex-assets.sh
```diff
@@ -0,0 +1,40 @@
...
+set -e
+
+root=$(git rev-parse --show-toplevel)
```
richvdh: might as well keep this working even outside git:


```suggestion
root=$(dirname "$0")/..
```

scripts/download-katex-assets.sh
```diff
@@ -0,0 +1,40 @@
...
+tar -xzvf "$archive" -C "$tmp_dir"
+
+# Move the CSS file into place.
+install -vm644 "$tmp_dir/katex/katex.min.css" "$root/static/css/katex.min.css"
+
+# Remove any existing fonts and move the new ones into place.
+rm -rvf "$root"/static/css/fonts/KaTeX*
+while read -r file; do
+    install -vm644 "$file" "$root/static/css/fonts"
+done < <(find "$tmp_dir/katex/fonts" -maxdepth 1 -name "KaTeX*.woff2")
```
richvdh: not that it really makes any difference here, but just as a matter of good practice:

```suggestion
while IFS= read -r -d '' file; do
    install -vm644 "$file" "$root/static/css/fonts"
done < <(find "$tmp_dir/katex/fonts" -maxdepth 1 -name "KaTeX*.woff2" -print0)
```

https://mywiki.wooledge.org/BashFAQ/020

content/olm-megolm/_index.md
richvdh: Have added a suggestion below, but honestly what you've done is great

scripts/download-katex-assets.sh
Johennes: Somehow I could have sworn that `read -r` was enough because I've been on this exact quest many years ago. But evidently my memory has failed me. Thanks for spotting this and linking the explanation.

Johennes:
Thanks for the review! ❤️

I hadn't expected this to actually go anywhere initially.

PR #2232, opened 2025-10-15
List M_RESOURCE_LIMIT_EXCEEDED under correct error code section

hughns:
I think that the `M_RESOURCE_LIMIT_EXCEEDED` error code is currently incorrectly listed under the "other error codes" section. This is because it implicitly *can* apply to any API endpoint making it a "common error code".

The description currently [reads](https://spec.matrix.org/v1.16/client-server-api/#other-error-codes):

> ... Typically, this error will appear on routes which attempt to modify state (e.g.: sending messages, account data, etc) and not routes which only read state (e.g.: [/sync](https://spec.matrix.org/v1.16/client-server-api/#get_matrixclientv3sync), [/user/{userId}/account_data/{type}](https://spec.matrix.org/v1.16/client-server-api/#get_matrixclientv3useruseridaccount_datatype), etc).

Looking at it from another angle, if it is specific to particular endpoints then the endpoints that could return it should say so. However, there are no references to it in the individual endpoints.

As such, I conclude that `M_RESOURCE_LIMIT_EXCEEDED` should therefore be listed as a common error code.

For reference, the original "MSC" in [Google Doc form](https://docs.google.com/document/d/1_CgwkfLznU56fwLUFZXFnKivzVUPjjNsTO0DpCB9ncM/edit?tab=t.0#heading=h.eaxawgsbu6t2) (it predates the present MSC practices) which became a spec [PR](https://github.com/matrix-org/matrix-spec-proposals/pull/1874).

As a sanity check for future changes to the spec: it could make sense to ensure that each "other error code" is referenced by at least one specific API endpoint. Otherwise it is effectively "orphaned".

### Pull Request Checklist

Preview: https://pr2232--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2232.clarification
+++ b/changelogs/client_server/newsfragments/2232.clarification
@@ -1,1 +1,1 @@
-M_RESOURCE_LIMIT_EXCEEDED is now listed as a common error code.
+`M_RESOURCE_LIMIT_EXCEEDED` is now listed as a common error code.


changelogs/client_server/newsfragments/2232.clarification
```diff
@@ -0,0 +1 @@
+M_RESOURCE_LIMIT_EXCEEDED is now listed as a common error code.
```
anoadragon453:
```suggestion
`M_RESOURCE_LIMIT_EXCEEDED` is now listed as a common error code.
```

PR #2234, opened 2025-10-17
Spec for MSC4312: Resetting cross-signing keys in the OAuth world

Johennes:
Spec for https://github.com/matrix-org/matrix-spec-proposals/pull/4312.

This does not include the `org.matrix.cross_signing_reset` account management action because account management URLs are only introduced into the spec with https://github.com/matrix-org/matrix-spec-proposals/pull/4191 which has not landed yet.

### Pull Request Checklist

Preview: https://pr2234--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/m.oauth_params.yaml
+++ b/data/api/client-server/definitions/m.oauth_params.yaml
@@ -24,6 +24,7 @@ properties:
       A URL pointing to the homeserver's OAuth account management web UI
       where the user can approve the action. MUST be a valid URI with scheme
       `http://` or `https://`, the latter being RECOMMENDED.
+    pattern: "^https?://"
 example: {
     "url": "https://example.org/account/reset-cross-signing"
 }


data/api/client-server/definitions/m.oauth_params.yaml
```diff
@@ -0,0 +1,29 @@
...
+description: Schema for `m.oauth` entry in the `params` object in a User-Interactive Authentication response.
+required: ['url']
+properties:
+  url:
+    type: string
+    format: uri
+    description: |
+      A URL pointing to the homeserver's OAuth account management web UI
+      where the user can approve the action. MUST be a valid URI with scheme
+      `http://` or `https://`, the latter being RECOMMENDED.
```
anoadragon453, line 26: Shall we enforce the `MUST` here?

```suggestion
      where the user can approve the action. MUST be a valid URI with scheme
      `http://` or `https://`, the latter being RECOMMENDED.
    pattern: "^https?://"
```

Johennes: Yes, good call.

PR #2238, opened 2025-10-24
Silence redocly-cli rule checking examples

zecakeh:
This rule makes CI fail because it was added in the latest release to the `recommended` ruleset that we use.

`$ref` inside examples is non-compliant for OpenAPI / JSON Schema, so this rule fails to check the examples properly.

We already perform this check with `scripts/check-event-schema-examples.py`.

### Pull Request Checklist

Preview: https://pr2238--matrix-spec-previews.netlify.app

### DIFF
--- a/redocly.yaml
+++ b/redocly.yaml
@@ -9,5 +9,6 @@ rules:
   no-path-trailing-slash: off
   operation-2xx-response: off
   spec-strict-refs: error
-  # $ref in examples are not spec-compliant so we perform this check in a python script.
+  # $ref in examples are not spec-compliant, yet we find them useful. Disable this rule
+  # and instead check example schema using `check-event-schema-examples.py`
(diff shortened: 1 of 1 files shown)


redocly.yaml
```diff
@@ -8,4 +8,6 @@ rules:
   no-invalid-media-type-examples: off
   no-path-trailing-slash: off
   operation-2xx-response: off
-  spec-strict-refs: error
\ No newline at end of file
+  spec-strict-refs: error
+  # $ref in examples are not spec-compliant so we perform this check in a python script.
```
anoadragon453: Even though the script name may change, I find it useful to link so one can follow the chain of logic.

```suggestion
  # $ref in examples are not spec-compliant, yet we find them useful. Disable this rule
  # and instead check example schema using `check-event-schema-examples.py`
```

Issue #1057, opened 2022-05-10
"Optional" keyword on `device_one_time_keys_count` is confusing


turt2live:
**Link to problem area**: https://spec.matrix.org/v1.2/client-server-api/#extensions-to-sync-1

**Issue**

The spec currently says:
> Optional. For each key algorithm, the number of unclaimed one-time keys currently held on the server for this device. If an algorithm is unlisted, the count for that algorithm is assumed to be zero. If this entire parameter is missing, the count for all algorithms is assumed to be zero.

This leads the reader to believe they can just leave the field off of the sync response, but if they read on then they find out that has associated behaviour. As a cheap solution, drop "Optional." and leave the rest of the description as-is. Bonus points for rewording more appropriately though.

neilalexander:
> This leads the reader to believe they can just leave the field off of the sync response, but if they read on then they find out that has associated behaviour. 

Honestly the specced behaviour of assuming that a missing value means zero is a footgun anyway. A missing value should mean *nothing*. If the server doesn't support handling OTKs for example, it should be possible to omit the field and expect that a client will do *nothing*. We should not be encouraging clients to implement magic behaviours.

PR #2245, opened 2025-11-12
Clarify that device_one_time_keys_count is only optional if no unclaimed one-time keys exist

Johennes:
Fixes: #1057

This doesn't fix the issue raised in https://github.com/matrix-org/matrix-spec/issues/1057#issuecomment-1122824696 but given that this behavior is already spec'ed, I believe changing it would require an MSC.

I've also allowed myself to remove some excess whitespace from this table and the next. Editing these lengthy tables feels cumbersome enough without it.

### Pull Request Checklist

Preview: https://pr2245--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1778,7 +1778,7 @@ a way to identify the server's support for fallback keys.
 | Parameter                        | Type              | Description                                                                                                            |
 |----------------------------------|-------------------|------------------------------------------------------------------------------------------------------------------------|
 | device_lists                     | DeviceLists       | Optional. Information on e2e device updates. Note: only present on an incremental sync.                                |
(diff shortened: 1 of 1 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1775,19 +1775,18 @@ property is required for inclusion, though previous versions of the
...
-| device_unused_fallback_key_types | [string]           | **Required.** The unused fallback key algorithms.                                                                      |
+| Parameter                        | Type              | Description                                                                                                            |
+|----------------------------------|-------------------|------------------------------------------------------------------------------------------------------------------------|
+| device_lists                     | DeviceLists       | Optional. Information on e2e device updates. Note: only present on an incremental sync.                                |
+| device_one_time_keys_count       | {string: integer} | **Required if unclaimed one-time keys exist.** For each key algorithm, the number of unclaimed one-time keys currently held on the server for this device. If the count for an algorithm is zero, servers MAY omit that algorithm. If the count for all algorithms is zero, servers MAY omit this parameter entirely. |
```
anoadragon453: small nit:

```suggestion
| device_one_time_keys_count       | {string: integer} | **Required if any unclaimed one-time keys exist.** For each key algorithm, the number of unclaimed one-time keys currently held on the server for this device. If the count for an algorithm is zero, servers MAY omit that algorithm. If the count for all algorithms is zero, servers MAY omit this parameter entirely. |
```

PR #2246, opened 2025-11-14
Clarify that servers may choose not to use `M_USER_DEACTIVATED` when they don't know who is asking.

reivilibre:
Synapse has started wiping password hashes on deactivation and now can't authenticate deactivated users who are trying to log in.

We seem to be agreed internally that wiping password hashes is good data hygiene (but, notably, not universal) and also that not revealing deactivated status to the public is also good privacy etiquette, so for password users we don't appear to be able to do anything better.

In that case, it seems like we should clarify the spec that this is an optional error code.

Because clients will already have to anticipate login attempts using the wrong password, this should not particularly be a breaking change, although Synapse's behavioural change did unfortunately break one test suite.

See: https://github.com/element-hq/synapse/issues/15747

### Pull Request Checklist

Preview: https://pr2246--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/login.yaml
+++ b/data/api/client-server/login.yaml
@@ -263,7 +263,7 @@ paths:
                 ID.
               * `M_USER_DEACTIVATED`: The user has been deactivated.
                 Servers MAY instead use `M_FORBIDDEN` when they can no longer authenticate
-                the deactivated user (i.e. their password has been wiped).
+                the deactivated user (e.g. their password has been wiped).
           content:
             application/json:
               schema:


data/api/client-server/login.yaml
```diff
@@ -262,6 +262,8 @@ paths:
                 or the requested device ID is the same as a cross-signing key
                 ID.
               * `M_USER_DEACTIVATED`: The user has been deactivated.
+                Note that servers MAY choose not to use this error code and instead use `M_FORBIDDEN`,
+                particularly when the server can't authenticate the deactivated user.
```
anoadragon453: I think it'd be good to clarify what we're hinting at.

```suggestion
                Servers MAY instead use `M_FORBIDDEN` when they can no longer authenticate
                the deactivated user (i.e. their password has been wiped).
```

reivilibre:
CI failure doesn't look like anything I touched

Issue #1026, opened 2022-04-08
event_id_only should not be called "the only format available"


Johennes:
**Link to problem area**: https://github.com/matrix-org/matrix-spec/blob/ca466b5a572a867de79ba9077e6445265b841c96/data/api/client-server/pusher.yaml#L237

**Issue**

The `format` parameter in `PusherData` is documented as

> Currently the only format available is 'event_id_only'.

This is confusing because while `event_id_only` is the only allowed value for `format`, the parameter is optional and omitting it sends a lot more than just the event ID which arguably could be called a "format" of its own.

babolivier:
Related: https://github.com/matrix-org/matrix-spec-proposals/pull/2782

PR #2255, opened 2025-11-28
Clarify that usage of event_id_only is not mandatory

Johennes:
This makes it more explicit that the `format` parameter on pushers is optional and that not specifying it results in behavior that is drastically different from what `event_id_only` does.

Fixes: #1026

### Pull Request Checklist

Preview: https://pr2255--matrix-spec-previews.netlify.app


changelogs/client_server/newsfragments/2255.clarification
```diff
@@ -0,0 +1 @@
+Usage of `event_id_only` is not mandatory.
```
anoadragon453:
```suggestion
Usage of the `event_id_only` format for push notifications is not mandatory.
```

PR #2256, opened 2025-12-01
Add version picker in the navbar

Johennes:
This adds a version picker in the top right of the navbar. To populate the picker without having to rebuild historical specs on new releases, a new file is `/versions.json` is generated that can be loaded from `/latest/versions.json` by any spec version.

I haven't been able to fully test this because the local and netlify build only contain the unstable version. For the same reason, the PR is currently using `/version.json` rather than `/latest/versions.json`.

Screenshot 2025-12-01 at 09 01 07

Fixes: #951

### Pull Request Checklist

Preview: https://pr2256--matrix-spec-previews.netlify.app

### DIFF
--- a/assets/js/versions.template.js
+++ b/assets/js/versions.template.js
@@ -82,8 +82,7 @@ function appendVersion(parent, name, url) {
     a.appendChild(text);
 }
 
-// TODO: Load /latest/versions.json
-fetch("/versions.json")
+fetch("/latest/versions.json")
     .then(r => r.json())
     .then(versions => {
         // Find the surrounding list element


assets/js/versions.template.js
```diff
@@ -0,0 +1,78 @@
...
+        window.location.href = href.replace(matches[0], `/${name}/`);
+    });
+
+    // The link text
+    const text = document.createTextNode(name);
+    a.appendChild(text);
+}
+
+// TODO: Load /latest/versions.json
+fetch("/versions.json")
```
anoadragon453: To be completed before merging this PR.

assets/js/versions.js
```diff
@@ -0,0 +1,78 @@
...
+        // Add an entry for the unstable version
+        appendVersion(ul, "unstable", "https://spec.matrix.org/unstable");
+
+        // Add an entry for each proper version
+        for (const version of versions) {
+            appendVersion(ul, version.name, `https://spec.matrix.org/${version.name}`);
+        }
+
+        // For historical versions, simply link to the changelog
+        appendVersion(ul, "historical", "/changelog/historical/");
```
anoadragon453: I think this will link to https://spec.matrix.org/changelog/historical/, which is a 404. And what you want is https://spec.matrix.org/$version/changelog/historical/?

layouts/_partials/navbar-version-selector.html
```diff
@@ -0,0 +1,18 @@
...
+
+{{ $changelog := site.GetPage "changelog" }}
+{{ $pages := $changelog.RegularPages.ByDate.Reverse }}
+
+<div class="dropdown">
+    <a class="nav-link dropdown-toggle" href="#" role="button" data-bs-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+        All Versions
+    </a>
+    <ul class="dropdown-menu" id="version-selector">
+        {{- /* The menu is built by version.js */ -}}
```
anoadragon453:
```suggestion
        {{- /* The menu is built by versions.js */ -}}
```

assets/js/versions.template.js
```diff
@@ -0,0 +1,78 @@
...
+        ev.preventDefault();
+        ev.stopPropagation();
+
+        // Try to find the current version segment
+        const href = window.location.href;
+        const matches = href.match(/\/unstable\/|\/latest\/|\/v\d+.\d+\//g);
+        if (!matches) {
+            window.location.href = url;
+            return;
+        }
```
anoadragon453: Optional extra: can we highlight the current version we're on in the dropdown list?

assets/js/versions.js
Johennes: Oh, right. I converted the script to a template in 5fe78ced533a926df095e745b906176e6283e07a so that we can let Hugo resolve the path (which I _think_ should make this work both locally and in the deployment).

layouts/_partials/navbar-version-selector.html
Johennes: Good spot. I have addressed this in 5fe78ced533a926df095e745b906176e6283e07a because I had to rename the file there anyway.

assets/js/versions.template.js
Johennes: I have given it the blue in eea13b0ff2aa1a8e39df6700962b38d10825b244.

I have also given the unstable element the red in 9170eacfcd7103aa63c19c97247cde46cbe64bad if it's unselected. Curious what you think about that. I was also unsure if we wanted to have the unstable link at the very top or not?

Johennes:
I noticed that the dropdown didn't scroll if the window was too small. Have made it scrollable in 77b37cb9a9e96027200b63ab27e0921a0d125832.

assets/js/versions.template.js
anoadragon453: Hmm, after trying it locally I still don't feel the blue communicates "you have selected this version". Where after changing the CSS rule to `font-weight: bold` I feel definitely does.

I'm also not sure if `unstable` should be coloured red. There's nothing inherently wrong with looking at the unstable spec. Even though it's "unstable" and "subject to change", we encourage implementation from the point of FCP merge.

I do like the idea of "latest" being blue though - to sort of cement that as the default, and the one that most developers should be looking at.

What do you think?

Johennes: Yeah, after trying it out, I think you're right. Bold feels better than blue for the selected item. Also agree on red maybe being too much for the "unstable" item. Have made those changes in 70904ff1a83cdff206afe640604d4ee8f64c4985 and also added a "latest" item in blue.

Johennes:
Have created https://github.com/matrix-org/matrix-spec/pull/2257 for the 1.16 backport as a start (which I _think_ should make `/latest/versions.json` available).

PR #2257, opened 2025-12-04
Add version picker in the navbar (v1.16 backport)

Johennes:
This backports https://github.com/matrix-org/matrix-spec/pull/2256 into the 1.16 release. The patch applied cleanly. I merely renamed the changelog file.

### DIFF
--- a/assets/js/versions.template.js
+++ b/assets/js/versions.template.js
@@ -1,0 +1,106 @@
+/*
+Copyright 2025 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Determine the current version as defined in hugo.toml. This will either be
+// "unstable" or "vX.X" and doesn't depend on the current URL.
+//
+// The oddity below is an attempt at producing a readable Hugo template while
+// avoiding JS syntax errors in your IDE.
+const currentVersion = `{{ if eq .Site.Params.version.status "unstable" }}
+    {{- /**/ -}}
+    unstable
+    {{- /**/ -}}
+{{ else }}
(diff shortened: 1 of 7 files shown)


assets/js/versions.template.js
```diff
@@ -0,0 +1,106 @@
+/*
```
MadLittleMods, line 1: > I think the problem with the version picker is that if we don't backport it into all older releases, you'll be left stranded once you click out into older versions because there'll be no version picker to take you back.
>
> *-- @Johennes, https://github.com/matrix-org/matrix-spec/pull/2257#issuecomment-3612817662*

> I think that's probably fine, honestly. We won't be updating the older pre-1.1 versions to have the picker either, and deploying 16 backports is non-trivial.
>
> *-- @turt2live, https://github.com/matrix-org/matrix-spec/pull/2257#issuecomment-3612966124*

Why not roll it out to all of the spec versions?

This isn't changing spec content and is very useful to be able to jump around. I immediately ran into this when I saw it in [Matrix Live](https://youtu.be/3Axa0Kw3XPI?t=884) and tried to use it myself.

turt2live: The spec build is not 100% reproducible and can introduce errors ranging from slight visual annoyances to inoperable pages. We mitigate this variance by building the spec at release time in two ways: one as the "latest" version and another as the "historical" version (the same as latest but with banners to say it's old). 

When we're deploying a new spec release, we swap out the previous spec version's actual deployed files with the "historical" one and append the new latest version to the web server.

Backporting this feature to previous spec releases would either involve attempting to rebuild the spec at that version (which may fail subtly or catastrophically in some cases) or manually injecting and testing a block of code into each spec version. Neither option is trivial, and would have to be done 16 times at least. 

We simply do not have the time available to do that work or fix any of the bugs it might cause. Instead, we choose to roll forward as much as possible, leaving older spec versions best described by their build tag: historical. This can mean that some build-related issues take [multiple releases](https://github.com/matrix-org/matrix-spec/issues/2116) to fix, but it's currently considered an acceptable loss. 

We'll also get to a point after enough time/releases where people will not hit the issue as prominently. It's certainly not great that the bug exists, but given the options of spending a full working week fighting 16 spec releases and waiting for it to be less of an issue in 1-3 years, we're more in favour of waiting it out.

MadLittleMods: > manually injecting and testing a block of code into each spec version

This seems like the most applicable path.

Hopefully someone wants to put in the effort. Even with bad builds, feels strange that we can't build the site until the diff shows just the single change we want.

Thanks for explaining!

PR #2259, opened 2025-12-05
Fix versions.json URL in unstable and local environment

Johennes:
This is another follow-up on #2258 which has left the version picker broken on https://spec.matrix.org/unstable. I switched to `.Site.Home.Permalink` for the base URL. This works locally and I hope it'll work in the deployment, too, because a similar pattern is used in the navbar.

If this doesn't work either, we can switch to `/unstable/versions.json` which, however, has the downside of leaving the picker empty on local builds.

CC @anoadragon453 

### Pull Request Checklist

Preview: https://pr2259--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Let's try it.

PR #2260, opened 2025-12-05
Prevent PurgeCSS from deleting dynamically-added version picker CSS classes

anoadragon453:
Follow-on from https://github.com/matrix-org/matrix-spec/pull/2256. These classes were getting removed from the final, minified CSS as they were only present after the dynamic JavaScript that builds the version picker HTML ran.

They showed up just fine in a local `hugo serve` run as our CSS minification pipeline doesn't run there.

### Pull Request Checklist

Preview: https://pr2260--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2260.clarification
+++ b/changelogs/internal/newsfragments/2260.clarification
@@ -1,0 +1,1 @@
+Add version picker in the navbar.


assets/scss/_styles_project.scss
```diff
@@ -51,19 +51,24 @@ Custom SCSS for the Matrix spec
...
+  /* Make the version dropdown scroll if it's too large */
   ul#version-selector {
     max-height: 80vh;
     overflow-y: auto;
   }
 
+  /* As these styles are only applied by JavaScript, PurgeCSS doesn't see them
+   * in the source code and removes them unless we explicitly tell it not to.
+   */
+  /* purgecss start ignore */
```
anoadragon453, line 63: I decided to explicitly exclude these classes, rather than adding `selected` / `latest` to the PurgeCSS config, as that may affect unrelated classes.

Issue #784, opened 2021-03-10
Create a quick reference to API endpoints


richvdh:
**Link to problem area**:

https://spec.matrix.org/unstable/client-server-api

**Issue**

I used to use the old ToC as a quick-reference to the API endpoints, but that's no longer a thing you can do. That's probably for the best, but it would be nice if there were a list of endpoints *somewhere*.

HarHarLinks:
How did the old functionality look like, what was the selling point? How does it compare to playground.matrix.org?

PR #2262, opened 2025-12-05
Add a list of endpoints to the top of each spec

anoadragon453:
Fixes https://github.com/matrix-org/matrix-spec/issues/784

Add a collapsible list of endpoints to the top of the page for each distinct spec. We do this by storing endpoint metadata on `$page` and creating a new partial, `endpoints-toc.html`, which renders it.

There was a bit of trickery involved in rendering this partial separate from the rest of the page's content. I'm not entirely sure if what I did is the best way to do it - so feedback welcome.

Recommended to review commit-by-commit, as each commit contains further descriptions.

### Pull Request Checklist

Preview: https://pr2262--matrix-spec-previews.netlify.app

### DIFF
--- a/layouts/_partials/endpoints-toc.html
+++ b/layouts/_partials/endpoints-toc.html
@@ -5,8 +5,8 @@
   The outer page's Scratch must contain an "api_endpoints" key, which is either
   a slice of maps (a list of endpoint metadata dicts) or a map of module name ->
   slice of endpoint metadata dicts (representing the API modules and the
-  endpoints they each contain). Each slice item must contain the Each endpoint
-  dict must contain the following keys:
+  endpoints they each contain). Each endpoint dict must contain the following
+  keys:
 
     * `anchor`: the HTML anchor for the endpoint
     * `method`: the HTTP method


tulir:
Seems like it doesn't include endpoints from c-s api modules?

layouts/_partials/openapi/render-api.html
```diff
@@ -15,6 +15,7 @@
 {{ $api_data := index .api_data }}
 {{ $base_url := .base_url }}
 {{ $anchor_base := .anchor_base }}
+{{ $page := .page }}
```
richvdh, line 20: please can you document this in the list of params at the top of the file?

layouts/_partials/openapi/render-operation.html
```diff
@@ -22,13 +22,24 @@
 {{ $method := .method }}
 {{ $endpoint := .endpoint }}
 {{ $operation_data := .operation_data }}
+{{ $page := .page }}
```
richvdh, line 27: ditto

layouts/docs/list.html
richvdh, line 1: what is this file used for?

layouts/content.html
```diff
@@ -1,4 +1,3 @@
-<div class="td-content">
-	<h1>{{ .Title }}</h1>
-	{{ .Content }}
-</div>
+{{ define "main" }}
```
richvdh: somewhat wondering why we need this `define` where we didn't before?

layouts/docs/single.html
```diff
@@ -0,0 +1,3 @@
+{{ define "main" }}
```
richvdh, line 17: might be nice to add a comment to the top of this file describing when it is used. Presumably hugo looks for a file here? might be nice to add a link to some docs?

Likewise for `content.html` while you're there

layouts/_partials/openapi/render-api.html
richvdh: ... and .module

layouts/docs/list.html
anoadragon453: It determines the content of the CS API page. https://gohugo.io/templates/types/#structure has some details of how all the pieces fit together.

In hugo, you have base templates, which are overridden by user-defined templates. We define ours [here](https://github.com/matrix-org/matrix-spec/blob/7c39427d8b33073f55883debd0913f998620d122/layouts/docs/baseof.html). This template contains the navbar, footer, etc.

Note that it has the following:

```
{{ block "main" . }}{{ end }}
```

A [block](https://gohugo.io/functions/go-template/block/) is a placeholder that can be overridden. `define "main"` defines a [block](https://gohugo.io/functions/go-template/block/) named "main" that can replace this placeholder.

A [list template](https://gohugo.io/templates/types/#list) renders "lists" of things, like an archive of blog posts. Here we're using it to render the Client-Server API page (which has a list of modules).

A [single template](https://gohugo.io/templates/types/#single) renders a single thing, like a single blog post. We're using this to render each non-CS API page (S-S, AS, etc.).

layouts/content.html
anoadragon453: Ah, that was left over from me experimenting.

This entire file is actually no longer necessary, now that we're rendering the non-CS API pages via `single.html` instead.

Previously, we were quietly making use of [the default `single.html` template from Docsy](https://github.com/matrix-org/docsy/blob/5df72519f5af2355348b8cfe4f4cdf320efe37b8/layouts/docs/single.html), which just renders `layouts/content.html`.

We're now overriding `single.html`, which removes an invisible layer of indirection.

layouts/_partials/spec-content.html
anoadragon453: We could delete this partial once it was no longer used by `content.html`.

layouts/_partials/endpoints-toc.html
```diff
@@ -0,0 +1,42 @@
+{{/* Minimal list of API endpoints for the current page. */}}
```
richvdh: it might be nice to document here, that this expects some values in `Scratch`, and what format it expects.

layouts/docs/single.html
```diff
@@ -0,0 +1,19 @@
...
+
+  Override the `layouts/docs/single.html` template from Docsy. While this template
+  is equivalent to `list.html`, it's necessary to override both templates to
+  avoid falling back to the default Docsy templates.
+
+  https://gohugo.io/templates/types/#single
+
+  We use this single template to render the all API spec pages *except* the
+  Client-Server API. The Client-Server API spec page contains modules, and thus
+  is handled separately.
```
richvdh:
```suggestion
  is handled separately, via the `layouts/docs/list.html` template.
```

layouts/docs/single.html
```diff
@@ -0,0 +1,19 @@
...
+
+  The contents of the "main" block below are inserted into the `./baseof.html`
+  base template.
+
+*/}}
+{{ define "main" }}
+  {{ partial "spec-content.html" . }}
+{{ end }}
```
richvdh: (missing NL at EOF)

data/cs_modules.yaml
richvdh, line 1: This file could do with a comment to explain its purpose in life.

Generally it feels like a thing that is going to be annoying to maintain: it's rather subtle that you have to add a line here when adding a new module. For now, we should add some pointers in https://github.com/matrix-org/matrix-spec/blob/main/content/client-server-api about it, but in future it might be nice to use this data as a source of truth for the list at the bottom of `client-server-api/_index.md`.

layouts/content.html
```diff
@@ -1,4 +0,0 @@
-<div class="td-content">
-	<h1>{{ .Title }}</h1>
-	{{ .Content }}
-</div>
```
richvdh, line 4: Note for future archaeologists: this file is no longer necessary:

> Previously, we were quietly making use of the default `single.html` template from Docsy, which just renders `layouts/content.html`.
>
> We're now overriding `single.html`, which removes an invisible layer of indirection.

data/cs_modules.yaml
anoadragon453: Good catch - I've ended up just inlining it into the partial invocation in ab795c23.

layouts/_partials/endpoints-toc.html
```diff
@@ -0,0 +1,61 @@
+{{/*
+  
+  Renders a list of API endpoints for the current page, given:
+
+  The outer page's Scratch must contain an "api_endpoints" key, which is either
+  a slice of maps (a list of endpoint metadata dicts) or a map of module name ->
+  slice of endpoint metadata dicts (representing the API modules and the
+  endpoints they each contain). Each slice item must contain the Each endpoint
+  dict must contain the following keys:
```
richvdh:
```suggestion
  endpoints they each contain). Each endpoint dict must contain the following
  keys:
```

PR #2264, opened 2025-12-05
Use more specific names for the version picker CSS classes and add them to the PurgeCSS safelist

Johennes:
### Pull Request Checklist

Preview: https://pr2264--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Yep, looks like this is the way to go. I felt you had definitely hit the nail on the head with https://github.com/matrix-org/matrix-spec/pull/2263, but ah well. Let's not spend too much more time on this and just get it working.

Thanks for your persistence!

PR #2267, opened 2025-12-16
Spec device management for application services

zecakeh:
As per MSC4190.

### Pull Request Checklist

Preview: https://pr2267--matrix-spec-previews.netlify.app

### DIFF
--- a/content/application-service-api.md
+++ b/content/application-service-api.md
@@ -547,18 +547,12 @@ client-server endpoint.
 Application services need to be able to create and delete devices to manage the
 encryption for their users without having to rely on `/login`, which also
 generates an access token for the user, and which might not be available for
-homeservers that only support the [OAuth 2.0 API](client-server-api/#oauth-20-api).
+homeservers that only support the [OAuth 2.0 API](/client-server-api/#oauth-20-api).
 
 ##### Creating devices
 
 Application services can use the [`PUT /_matrix/client/v3/devices/{deviceId}`](/client-server-api/#put_matrixclientv3devicesdeviceid)
 endpoint to create new devices.
-
-When a new device was created, the homeserver MUST return a 201 HTTP status code.
-It MUST still return a 200 HTTP status code if a device was updated.
-
-This endpoint is rate-limited for device creation. Servers MAY want to use login
-rate limits.
 
 ##### Deleting devices
 


data/api/client-server/device_management.yaml
```diff
@@ -87,8 +87,21 @@ paths:
       tags:
         - Device management
     put:
-      summary: Update a device
-      description: Updates the metadata on the given device.
+      summary: Create or update a device
+      description: |-
+        Updates the metadata on the given device.
```
richvdh:
```suggestion
        Updates the metadata on the given device, or creates a new device.
```

data/api/client-server/device_management.yaml
```diff
@@ -87,8 +87,21 @@ paths:
...
+
+        {{% boxes/note %}}
```
richvdh: Not sure this belongs in an info-box: it feels like an important part of the API, rather than being a sidebar as an info box is normally used for. Suggest removing the `{{% boxes/note %}}` wrapper.

data/api/client-server/device_management.yaml
```diff
@@ -87,8 +87,21 @@ paths:
...
+        {{% added-in v="1.17" %}}
```
richvdh: Suggest moving this to the bottom, with some explanatory text:


```
        {{% changed-in v="1.17" %}} The ability to create new devices was added.
```

data/api/client-server/device_management.yaml
```diff
@@ -87,8 +87,21 @@ paths:
...
+        This endpoint can be used by application services to create a device.
```
richvdh: I'd like to see it emphasised more that *only* ASes can create devices this way


```suggestion
        The ability to create new devices is only available to application services: regular users may
        only update existing devices.
```

data/api/client-server/device_management.yaml
```diff
@@ -87,8 +87,21 @@ paths:
...
+
+        When a new device was created, the homeserver MUST return a 201 HTTP
+        status code. It MUST still return a 200 HTTP status code if a device was
+        updated.
+
+        This endpoint is rate-limited for device creation. Servers MAY want to
+        use login rate limits.
```
richvdh: some spare words here.
```suggestion
        This endpoint is rate-limited for device creation. Servers MAY use login rate limits.
```

data/api/client-server/device_management.yaml
```diff
@@ -142,6 +165,12 @@ paths:
         Since this endpoint uses User-Interactive Authentication, it cannot be used when the access token was obtained
         via the [OAuth 2.0 API](/client-server-api/#oauth-20-api).
         {{% /boxes/warning %}}
+
+        {{% boxes/note %}}
+        {{% added-in v="1.17" %}}
+        When this endpoint is used by an application service, the server MUST NOT
+        require User-Interactive Authentication.
+        {{% /boxes/note %}}
```
richvdh: Again, the use of an info-box here is confusing: the main text ("This API endpoint uses the User-Interactive Authentication API") now directly contradicts the info box. Instead, we need to change the main text:

```
       This API endpoint uses the [User-Interactive Authentication API](/client-server-api/#user-interactive-authentication-api), except when used by an application service.

       Deletes the given device, and invalidates any access token associated with it.

       {{% boxes/warning %}}
       Since this endpoint uses User-Interactive Authentication, it cannot be used when the access token was obtained
       via the [OAuth 2.0 API](/client-server-api/#oauth-20-api).
       {{% /boxes/warning %}}

       {{% changed-in v="1.17" %}} This endpoint no longer requires User-Interactive Authentication when used by an application service
```

richvdh: likewise elsewhere in this PR, for the other endpoints.

content/client-server-api/_index.md
```diff
@@ -1562,6 +1562,10 @@ If the access token does correspond to an appservice, but the user id does
 not lie within its namespace then the homeserver will respond with an
 errcode of `M_EXCLUSIVE`.
 
+{{% added-in v="1.17" %}} If this login type is used and the server doesn't
+support logging in via the Legacy authentication API, it MUST return a 400 HTTP
+status code with an `M_APPSERVICE_LOGIN_UNSUPPORTED` error code.
```
richvdh, line 1572: it might also be good to add this to the definition of a 400 response to the login API, under https://pr2267--matrix-spec-previews.netlify.app/client-server-api/#post_matrixclientv3login, here: 

Image

we could say something like: 

> Specific error codes used with this status code include:
>
>  * {{% added-in v="1.17" %}} `M_APPSERVICE_LOGIN_UNSUPPORTED`: an application service used the [`m.login.application_service`](https://pr2267--matrix-spec-previews.netlify.app/client-server-api/#appservice-login) type, but the server doesn't support logging in via the Legacy authentication API.

data/api/client-server/registration.yaml
```diff
@@ -60,6 +60,18 @@ paths:
 
         Any user ID returned by this API must conform to the grammar given in the
         [Matrix specification](/appendices/#user-identifiers).
+
```
richvdh, line 63: Aside: it's unfortunate that this description doesn't mention the fact that `/register` takes a `type` param which means we can skip UIAA, but that is an existing problem. I've opened https://github.com/matrix-org/matrix-spec/issues/2269.

data/api/client-server/registration.yaml
```diff
@@ -60,6 +60,18 @@ paths:
...
+        {{% boxes/note %}}
+        {{% added-in v="1.17" %}}
+        Even if the server doesn't support the Legacy authentication API, it
+        MUST support this endpoint for application services to be able to
+        [create users](/application-service-api/#server-admin-style-permissions).
+
+        In that case application services MUST set the `"inhibit_login": true`
+        parameter as they cannot use it to log in as users. If the
+        `inhibit_login` parameter is not set to `true`, the server MUST return a
+        400 HTTP status code with an `M_APPSERVICE_LOGIN_UNSUPPORTED` error code.
```
richvdh, line 73: Again, it would be nice to mention the new error code in the "Responses" section of the API doc

content/application-service-api.md
```diff
@@ -444,33 +446,57 @@ achieved by including the `as_token` on a `/register` request, along
...
+```
+POST /_matrix/client/v3/register
+Authorization: Bearer YourApplicationServiceTokenHere
+
+Content:
+{
+  "type": "m.login.application_service",
+  "username": "_irc_example"
+}
+```
```
richvdh: Since you're updating this anyway:

In other places where we give example requests, we don't say `Content`, and leave it implied. See https://pr2267--matrix-spec-previews.netlify.app/client-server-api/#user-interactive-api-in-the-rest-api for some examples.

Also, using two code blocks means we can decorate the content as json.

content/application-service-api.md
```diff
@@ -512,6 +538,44 @@ client-server endpoint.
...
+##### Creating devices
+
+Application services can use the [`PUT /_matrix/client/v3/devices/{deviceId}`](/client-server-api/#put_matrixclientv3devicesdeviceid)
+endpoint to create new devices.
+
+When a new device was created, the homeserver MUST return a 201 HTTP status code.
+It MUST still return a 200 HTTP status code if a device was updated.
+
+This endpoint is rate-limited for device creation. Servers MAY want to use login
+rate limits.
```
richvdh: I don't think we need to repeat these details of the API here: they are clearly defined over on the spec for `PUT /_matrix/client/v3/devices/{deviceId}`. It's not wrong as such, just a bit duplicative, and there is a risk that we might change the API definition and forget to update this bit.

richvdh (changes requested):
Thanks very much for this, @zecakeh! I have a few comments.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

data/api/client-server/device_management.yaml
zecakeh: Fixed in b36630c36375651c1407c653bf0fd6aae1f4a689.

content/client-server-api/_index.md
zecakeh: Fixed in c9523087b8992726f7e1bb685a352904685d02cc.

data/api/client-server/registration.yaml
zecakeh: Fixed in c9523087b8992726f7e1bb685a352904685d02cc.

content/application-service-api.md
zecakeh: Fixed in c71c1e225bb94801f509edd2fc589945b4080d90.

content/application-service-api.md
zecakeh: Fixed in 59cfacf94fe87b15e190c19f29dec272c22fcaf4.

PR #2275, opened 2025-12-18
Auto-create draft releases when building release tags

richvdh:
Update the CI so that it drafts a release, with the correct artifacts attached and the release notes prepared, when building a release tag.

Preview: https://pr2275--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -243,7 +243,13 @@ jobs:
     name: "🔎 Validate generated HTML"
     runs-on: ubuntu-latest
     needs: [calculate-baseurl, build-spec]
-    # run even if generate-changelog was skipped
+    # Run even if `generate-changelog` was skipped.
+    #
+    # `build-spec` has a dependency on `generate-changelog` to ensure order of execution
+    # and to access `needs.generate-changelog.result`. However, `generate-changelog` is
+    # skipped on tag builds; even a transient dependency on `generate-changelog` is then
+    # enough for this step to also be skipped by default on tag builds. Hence the need for
+    # this explicit `if`.
     if: ${{ !failure() && !cancelled() }}
     steps:
       - name: "📥 Source checkout"


.github/workflows/main.yml
```diff
@@ -243,6 +243,8 @@ jobs:
     name: "🔎 Validate generated HTML"
     runs-on: ubuntu-latest
     needs: [calculate-baseurl, build-spec]
+    # run even if generate-changelog was skipped
+    if: ${{ !failure() && !cancelled() }}
```
richvdh: this is actually an unrelated fix; I was just confused about why this didn't run as part of the tag build.

.github/workflows/main.yml
anoadragon453: I believe the reason we're `need`ing `generate-changelog` here is ensuring that `generate-changelog` runs before `build-spec` runs.

Could we add a comment saying as much?


```suggestion
    # Ensure these jobs run and complete before building the spec.
    needs: [calculate-baseurl, build-openapi, generate-changelog]
```

.github/workflows/main.yml
```diff
@@ -243,6 +243,8 @@ jobs:
     name: "🔎 Validate generated HTML"
     runs-on: ubuntu-latest
     needs: [calculate-baseurl, build-spec]
+    # run even if generate-changelog was skipped
```
anoadragon453: build-spec depends on `generate-changelog`, and we then depend on `build-spec` here, hence getting skipped if `generate-changelog` doesn't fire (which it doesn't on tags).

Yet `build-spec` must put `generate-changelog` in `needs` so that:

* `generate-changelog` runs before `build-spec` runs
* `build-spec` has access to `needs.generate-changelog.result`.

Could we add a comment saying as much?


```suggestion
    # run even if generate-changelog was skipped.
    # build-spec `needs` generate-changelog, but generate-changelog is skipped
    # on tags. The `needs` is required to ensure order of execution and to access
    # `needs.generate-changelog.result`.
```

Otherwise I was initially tempted to just remove `generate-changelog` from `build-spec`s `needs`.

anoadragon453 (approved):
Useful to automate this step. Thanks!

.github/workflows/main.yml
richvdh:
```suggestion
    # Run even if `generate-changelog` was skipped.
    #
    # `build-spec` has a dependency on `generate-changelog` to ensure order of execution
    # and to access `needs.generate-changelog.result`. However, `generate-changelog` is
    # skipped on tag builds; even a transient dependency on `generate-changelog` is then
    # enough for this step to also be skipped by default on tag builds. Hence the need for
    # this explicit `if`.
```

PR #2276, opened 2025-12-19
Include spec release in filenames in built tarball

richvdh:
I found it slightly confusing that the tarballs generated by CI just have everything under `spec/`, rather than including the version number.

This PR changes things so that the paths in the tarball mirror the URLs contained within the spec. In other words:

 * Tarballs built from a `vNN` tag will contain the generated spec under `/vNN/`.
 * Tarballs built from the `main` branch will contain the generated spec under `/unstable/`.
 * Tarballs built for a PR branch will have the generated spec at the top level.

If we merge this, we will need to followup with a change to the autodeployer for https://spec.matrix.org/unstable, so that it knows to find the generated spec at `/unstable` rather than `/spec`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -250,6 +250,14 @@ jobs:
     name: "🔎 Validate generated HTML"
     runs-on: ubuntu-latest
     needs: [calculate-baseurl, build-spec]
+    # Run even if `generate-changelog` was skipped.
+    #
+    # `build-spec` has a dependency on `generate-changelog` to ensure order of execution
+    # and to access `needs.generate-changelog.result`. However, `generate-changelog` is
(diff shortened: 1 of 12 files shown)


anoadragon453 (approved):
tar magic looks fine.

> If we merge this, we will need to followup with a change to the autodeployer

Looks like we're not currently using any option to determine the extracted path: https://github.com/matrix-org/matrix-ansible-private/blob/782bc351a97ed52e401056c26805d2c668c5778e/roles/gha_webhook_listener/templates/webhook-listener.service.j2#L2 (internal link).

However [the script itself](https://github.com/matrix-org/gha-webhook-listener) has a `--hook-script` option which allows running some commands after the fact, which could be used to rename directories.

PR #2277, opened 2025-12-19
Spec Additional Error Codes for submitToken endpoint

zecakeh:
As per MSC4183

### Pull Request Checklist

Preview: https://pr2277--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/definitions/request_token_response.yaml
+++ b/data/api/client-server/definitions/request_token_response.yaml
@@ -32,8 +32,8 @@ properties:
       An optional field containing a URL where the client must submit the
       validation token, with identical request and response parameters and error
       codes to the Identity Service API's
-      [`POST /_matrix/identity/v2/validate/email/submitToken`](/identity-service-api/#post_matrixidentityv2validateemailsubmittoken)
-      and [`POST /_matrix/identity/v2/validate/msisdn/submitToken`](/identity-service-api/#post_matrixidentityv2validatemsisdnsubmittoken)
+      [`POST /validate/email/submitToken`](/identity-service-api/#post_matrixidentityv2validateemailsubmittoken)
+      and [`POST /validate/msisdn/submitToken`](/identity-service-api/#post_matrixidentityv2validatemsisdnsubmittoken)
       endpoints (without the requirement for an access token). The homeserver
       must send this token to the user (if applicable), who should then be
       prompted to provide it to the client.


data/api/identity/v2_email_associations.yaml
```diff
@@ -153,6 +153,25 @@ paths:
                   value: {
```
zecakeh, line 153: The introduction of the MSC also talks about applying the same error codes to the GET endpoints, but not the proposal section, so I didn't touch them because I am not sure which is more correct?

anoadragon453: @dbkr as the author of the MSC in question, what was your intent?

anoadragon453 (commented):
Taking off the review queue until https://github.com/matrix-org/matrix-spec/pull/2277#discussion_r2655719282 is answered.

data/api/identity/v2_email_associations.yaml
dbkr: I think the intro is wrong and the proposal section is correct: the GET endpoint should return a user-facing HTML page and therefore no error code.

richvdh: added a note at https://github.com/matrix-org/matrix-spec-proposals/pull/4183#discussion_r2687585633.

data/api/identity/v2_email_associations.yaml
```diff
@@ -153,6 +153,25 @@ paths:
...
                     "success": true
                   }
+        "400":
+          x-addedInMatrixVersion: "1.18"
```
richvdh: this added-in doesn't appear in the rendered spec. Also, putting it here rather implies that this endpoint wasn't allowed to return a 400 until 1.18, which isn't really true: we just didn't define the error codes.

I think the easiest solution to both problems is probably to add **`(added in v1.18)`** to each of the error codes?

data/api/client-server/definitions/request_token_response.yaml
```diff
@@ -24,16 +24,22 @@ properties:
...
+        the Identity Service API's endpoint.
     description: |-
       An optional field containing a URL where the client must submit the
-      validation token to, with identical parameters to the Identity Service
-      API's `POST /validate/email/submitToken` endpoint (without the requirement
-      for an access token). The homeserver must send this token to the user (if
-      applicable), who should then be prompted to provide it to the client.
+      validation token, with identical request and response parameters and error
+      codes to the Identity Service API's
+      [`POST /_matrix/identity/v2/validate/email/submitToken`](/identity-service-api/#post_matrixidentityv2validateemailsubmittoken)
```
richvdh: I find the abbreviated form easier to read. The link clarifies that we are talking about v2.



```suggestion
      [`POST /validate/email/submitToken`](/identity-service-api/#post_matrixidentityv2validateemailsubmittoken)
```

data/api/client-server/definitions/request_token_response.yaml
```diff
@@ -24,16 +24,22 @@ properties:
...
+      endpoint (without the requirement for an access token). The homeserver
```
richvdh: it reads oddly that we reference `/validate/email/submitToken` even when we are talking about `**msisdn**/requestToken`. Maybe something like:


```suggestion
      and [`POST /_matrix/identity/v2/validate/msisdn/submitToken`](/identity-service-api/#post_matrixidentityv2validatemsisdnsubmittoken)
      endpoints (without the requirement for an access token). The homeserver
```

data/api/client-server/definitions/request_token_response.yaml
```diff
@@ -24,16 +24,22 @@ properties:
...
+      must send this token to the user (if applicable), who should then be
+      prompted to provide it to the client.
 
       If this field is not present, the client can assume that verification
       will happen without the client's involvement provided the homeserver
-      advertises this specification version in the `/versions` response
-      (ie: r0.5.0).
+      advertises this specification version in the [`GET /versions`](/client-server-api/#get_matrixclientversions)
```
richvdh:
```suggestion
      advertises a sufficiently recent specification version in the
      [`GET /versions`](/client-server-api/#get_matrixclientversions)
```

data/api/client-server/definitions/request_token_response.yaml
zecakeh: Don't we usually use the full path when we link to a different spec than the one that is rendered, to make it clear that we are talking about an endpoint from a different spec?

data/api/identity/v2_email_associations.yaml
zecakeh: Fixed in 611ba3b4f9f1fac9ccacf4a3a64079bdae10afa2, using the same syntax as we do a few times already in the CS API.

data/api/client-server/definitions/request_token_response.yaml
zecakeh: Applied in 4297109c1198ba7a5d50a8072d07fd49465c810b.

data/api/client-server/definitions/request_token_response.yaml
zecakeh: Applied in 4297109c1198ba7a5d50a8072d07fd49465c810b.

data/api/client-server/definitions/request_token_response.yaml
richvdh: hrm, maybe, but given it says "the Identity Service API" just before, that feels pretty obvious to me

data/api/client-server/definitions/request_token_response.yaml
```diff
@@ -24,16 +24,23 @@ properties:
...
     description: |-
       An optional field containing a URL where the client must submit the
-      validation token to, with identical parameters to the Identity Service
-      API's `POST /validate/email/submitToken` endpoint (without the requirement
-      for an access token). The homeserver must send this token to the user (if
-      applicable), who should then be prompted to provide it to the client.
+      validation token, with identical request and response parameters and error
+      codes to the Identity Service API's
+      [`POST /_matrix/identity/v2/validate/email/submitToken`](/identity-service-api/#post_matrixidentityv2validateemailsubmittoken)
+      and [`POST /_matrix/identity/v2/validate/msisdn/submitToken`](/identity-service-api/#post_matrixidentityv2validatemsisdnsubmittoken)
```
richvdh:
```suggestion
      and [`POST /validate/msisdn/submitToken`](/identity-service-api/#post_matrixidentityv2validatemsisdnsubmittoken)
```

PR #2278, opened 2025-12-19
Spec User suspension & locking endpoints

zecakeh:
As per MSC4323.

### Pull Request Checklist

Preview: https://pr2278--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2278.feature
+++ b/changelogs/client_server/newsfragments/2278.feature
@@ -1,1 +1,1 @@
-Add endpoints to lock and suspend server-local users for administrations and add the `m.account_management` capability, as per [MSC4323](https://github.com/matrix-org/matrix-spec-proposals/pull/4323).
+Add administrator endpoints to lock and suspend server-local users and add the `m.account_management` capability, as per [MSC4323](https://github.com/matrix-org/matrix-spec-proposals/pull/4323).
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -2282,7 +2282,7 @@ using their accounts, for instance, due to safety or security concerns.
 In contrast to account deactivation, locking is a non-destructive action
 that can be reversed.
 
-{{% added-in v="1.18" %}} To lock or unlock an account, the administrators
+{{% added-in v="1.18" %}} To lock or unlock an account, administrators
 SHOULD use the [`PUT /admin/lock/{userId}`](#put_matrixclientv1adminlockuserid)
 endpoint. They MAY also use [`GET /admin/lock/{userId}`](#get_matrixclientv1adminlockuserid)
 to check whether a user's account is locked.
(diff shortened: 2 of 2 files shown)


changelogs/client_server/newsfragments/2278.feature
```diff
@@ -0,0 +1 @@
+Add endpoints to lock and suspend server-local users for administrations and add the `m.account_management` capability, as per [MSC4323](https://github.com/matrix-org/matrix-spec-proposals/pull/4323).
```
anoadragon453:
```suggestion
Add administrator endpoints to lock and suspend server-local users and add the `m.account_management` capability, as per [MSC4323](https://github.com/matrix-org/matrix-spec-proposals/pull/4323).
```

content/client-server-api/_index.md
```diff
@@ -2280,9 +2280,12 @@ The server SHOULD return one of the following responses:
 Server administrators may apply locks to prevent users from usefully
 using their accounts, for instance, due to safety or security concerns.
 In contrast to account deactivation, locking is a non-destructive action
-that can be reversed. This specification describes the behaviour of clients
-and servers when an account is locked. It deliberately leaves the methods
-for locking and unlocking accounts as a server implementation detail.
+that can be reversed.
+
+{{% added-in v="1.18" %}} To lock or unlock an account, the administrators
```
anoadragon453:
```suggestion
{{% added-in v="1.18" %}} To lock or unlock an account, administrators
```

content/client-server-api/_index.md
```diff
@@ -2331,6 +2334,11 @@ from that account. The effect is similar to [locking](#account-locking), though
 without risk of the client losing state from a logout. Suspensions are reversible,
 like locks and unlike deactivations.
 
+{{% added-in v="1.18" %}} To suspend or unsuspend an account, the administrators
```
anoadragon453:
```suggestion
{{% added-in v="1.18" %}} To suspend or unsuspend an account, administrators
```

data/api/client-server/admin.yaml
```diff
@@ -107,6 +107,391 @@ paths:
...
+              examples:
+                response:
+                  value: {
+                    "errcode": "M_INVALID_PARAM",
+                    "error": "User does not belong to the local server."
+                  }
+        "403":
+          description: |-
+            The requesting user is not a server administrator, or the target user is another
+            administrator. The errcode is `M_FORBIDDEN`.
```
anoadragon453, line 171: @nexy7574 did you intend for server administrators to not be able to **look up** the suspend/lock status of another administrator?

I'll admit that this how MSC4323 is written, but I'm not sure what the justification for it would be.

timedoutuk: I believe the original intent was that server administrators couldn't be locked or suspended in the first place, ergo looking up their status would be pointless. Per https://github.com/nexy7574/matrix-spec-proposals/blob/nexy7574/agnostic-suspend-and-lock/proposals/4323-agnostic-suspend-and-lock.md#errors-and-restrictions, the "target cannot be an administrator" covers both the getter and setter endpoints:

> `403 / M_FORBIDDEN`: The requesting user is not a server administrator, is trying to suspend/lock their own account, or the target user is another administrator.

It could easily just be a no-op 200 `{"suspended"/"locked": false}` if that's more idiomatic, but just blanket refusing lookups that made no sense required less writing :^)

anoadragon453: I see. I could see both sides. Since it made it through the MSC process, let's keep it in the spec for now.

PR #2284, opened 2025-12-21
Specify basic validation for federation membership endpoints

tulir:
Preview: https://pr2284--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/2284.clarification
+++ b/changelogs/server_server/newsfragments/2284.clarification
@@ -1,1 +1,1 @@
-Specified input validation for PDUs passed to and returned from federation membership endpoints.
+Specify validation for PDUs passed to and returned from federation membership endpoints.


data/api/server-server/invites-v1.yaml
```diff
@@ -172,6 +172,17 @@ paths:
                   }
         "400":
           description: |-
+            The `M_INVALID_PARAM` error code is used to indicate one or more of the following:
```
tulir: This error code is made up. Synapse seems to use `M_UNKNOWN` for most of the checks, but writing that into the spec wouldn't be very nice. It can be reworded to not use any specific error code if necessary

timedoutuk: Wouldn't `M_BAD_JSON` make sense? [Hammerhead uses this](https://codeberg.org/timedout/hammerhead/src/commit/72e2c9d364ea0a39062580881b0059158f15e680/pkg/hammerhead/router/routes/federation/v2/send_join.go#L92)

> Request contained valid JSON, but it was malformed in some way, e.g. missing required keys, invalid values for keys.

richvdh: yeah, we do tend to use `M_BAD_JSON` for other things like this, for better or worse.

data/api/server-server/invites-v1.yaml
```diff
@@ -172,6 +172,17 @@ paths:
...
+
+            * The invite event fails a [signature check](/server-server-api/#validating-hashes-and-signatures-on-received-events).
```
richvdh: this seems to conflict with the advice below that `M_MISSING_PARAM` is used when "one or more events fails a signature check". Is your goal to single out the invite event for special treatment? If so then (a) I'm not sure that's necessary; (b) if you think it is necessary, then better be more explicit.

richvdh (changes requested):
Thanks very much for taking this on. It's certainly a valuable improvement.

I'm a bit concerned that putting the validation requirements way down in the "error codes" section buries the lede a bit, and that people aren't going to notice it. I'd be more inclined to put it in the endpoint description.

Similarly, I'd phrase it as "Servers MUST validate the following: ..." rather than "[error code] is used in the following situations: ...".

data/api/server-server/invites-v1.yaml
tulir: I read it as those checks only applying to `invite_room_state`, which doesn't include the invite itself. The specific error code is fairly arbitrary (as mentioned in the other review thread)

richvdh: ok. I would suggest using the same error code for both `invite_room_state` and the invite event itself.

richvdh:
> > I'm a bit concerned that putting the validation requirements way down in the "error codes" section buries the lede a bit, and that people aren't going to notice it. I'd be more inclined to put it in the endpoint description.
> > Similarly, I'd phrase it as "Servers MUST validate the following: ..." rather than "[error code] is used in the following situations: ...".
> 
> The main reason for both points was that invites already had `invite_room_state` validation in the response spec rather than the endpoint description. Should I move that stuff as well?

Ideally, I think so, yes, though at least in the case of `invite_room_state` we have a note on the request param that there is validation that must be done.

tulir:
Added the validation for GHSA-m5p2-vccg-8c9v, moved all of it from the response schemas to the request description and switched to use `M_MISSING_PARAM` for all errors

changelogs/server_server/newsfragments/2284.clarification
```diff
@@ -0,0 +1 @@
+Specified input validation for PDUs passed to and returned from federation membership endpoints.
```
richvdh:
```suggestion
Specify validation for PDUs passed to and returned from federation membership endpoints.
```

data/api/server-server/invites-v2.yaml
```diff
@@ -40,6 +40,30 @@ paths:
...
+        The `M_MISSING_PARAM` error code is also used to indicate one or more of the following
+        problems in the `invite_room_state` field:
+
+        * The `m.room.create` event is missing from `invite_room_state`.
+        * One or more entries in `invite_room_state` are not formatted according
+          to the room's version.
+        * One or more events fails a [signature check](/server-server-api/#validating-hashes-and-signatures-on-received-events).
+        * One or more events does not reside in the same room as the invite.
+          Note: Some room versions may require calculating the room ID for an
+          event rather than relying on the presence of `room_id`.
```
richvdh, line 64: I feel like we've gone in the wrong direction here in terms of the error code. `M_MISSING_PARAM` seems like a terrible choice for param that isn't missing.

Can we update both halves to use something more sensible like `M_INVALID_PARAM`? Or just leave the precise error code unspecified here?

data/api/server-server/joins-v2.yaml
```diff
@@ -247,6 +256,10 @@ paths:
             The error should be passed through to clients so that they
             may give better feedback to users.
 
+            If `M_MISSING_PARAM` is returned and the request is associated with a
```
richvdh: The description for this endpoint specifies `M_INVALID_PARAM` for validation failures. Let's make sure they are consistent.

data/api/server-server/invites-v2.yaml
tulir: If we want it to match implementations, probably just need to leave it unspecified. I guess it might be fine to specify something even if implementations don't match right now? Servers shouldn't be trying to catch and handle these errors anyway

richvdh: >  Servers shouldn't be trying to catch and handle these errors anyway

Except that, in the section about response codes, the spec *does* mandate trying to catch these errors and handle them differently, at least in the case of `/v2/invite/{roomId}/{eventId}`: it says that errors should be passed through to the client, except for `M_MISSING_PARAM` which should be turned into a 500 error. 

Ugh, I don't really know how best to proceed here. Certainly the best solution would be an MSC to clarify things, but that's a bunch of time and effort. 

Maybe we can just get away with saying something like "errors which indicate a faulty request such as `M_MISSING_PARAM` and `M_INVALID_PARAM` should be turned into 500 errors", and then leaving the exact error code to be returned under specific circumstances unspecified for now?

richvdh: (or, leave the "response codes" section just talking about `M_MISSING_PARAM`, and if we decide we need better error handling, MSC the response codes later?)

tulir: I made that case say both `M_MISSING_PARAM` and `M_INVALID_PARAM` should be hidden from clients, I guess the wording could be even more relaxed rather than specifying just 2 error codes

richvdh (approved):
lgtm. thanks.

PR #2291, opened 2026-01-08
Spec for MSC4356: Recently used emoji

Johennes:
Spec for [MSC4356](https://github.com/matrix-org/matrix-spec-proposals/pull/4356).

### Pull Request Checklist

Preview: https://pr2291--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/recent_emoji.md
+++ b/content/client-server-api/modules/recent_emoji.md
@@ -4,9 +4,8 @@
 
 This module enables clients to track a user's cumulated emoji usage across different
 devices. The data is stored in the [`m.recent_emoji`](#mrecent_emoji)
-global [account data](https://spec.matrix.org/v1.15/client-server-api/#client-config) and can, 
-among other things, be used to generate recommendations
-in emoji pickers.
+global [account data](#client-config) and can, among other things, be used to
+generate recommendations in emoji pickers.
 
 #### Events
 


content/client-server-api/modules/recent_emoji.md
```diff
@@ -0,0 +1,36 @@
...
+unspecified.
+
+To prevent excessive growth of the event as new emoji are being used, clients
+SHOULD limit the length of the `recent_emoji` array by dropping elements from
+its end. A RECOMMENDED maximum length is 100 emoji.
+
+To enable future extension, clients MUST tolerate and preserve array elements
+within `recent_emoji` that they don't understand or support. This means ignoring
+the entries when deciding what to display to the user while retaining them when
+modifying the array (unless the modification is for truncation).
```
Johennes, line 34: This is per [MSC4356](https://github.com/matrix-org/matrix-spec-proposals/pull/4356). However, the proposal also contained:

> Malicious or buggy clients could cause undefined behavior on other clients by setting emoji counters to negative values or the maximum allowed value. To prevent this, clients SHOULD drop any emojis with invalid entries from the list.

We cannot both preserve and drop elements that seem invalid to a client. Since allowing future extensibility seemed more beneficial to me, I have only retained that side.

data/event-schemas/schema/m.recent_emoji.yaml
Johennes, line 1: The spec renders this as "Message event" which feels a little confusing given that it lives in account data.

Image

However, it looks like this is the same for other account data events such as [`m.push_rules`](https://spec.matrix.org/v1.16/client-server-api/#mpush_rules), for instance. So I assume this needs to be fixed elsewhere.

content/client-server-api/modules/recent_emoji.md
```diff
@@ -0,0 +1,36 @@
+### Recently used emoji
+
+{{% added-in v="1.18" %}}
+
+This module enables clients to track a user's cumulated emoji usage across different
+devices. The data is stored in the [`m.recent_emoji`](/client-server-api/#mrecent_emoji)
```
richvdh:
```suggestion
devices. The data is stored in the [`m.recent_emoji`](#mrecent_emoji)
```

data/event-schemas/schema/m.recent_emoji.yaml
richvdh: Indeed. Related: https://github.com/matrix-org/matrix-spec/issues/897

content/client-server-api/modules/recent_emoji.md
```diff
@@ -0,0 +1,36 @@
...
+global account data event and can, among others, be used to generate recommendations
```
richvdh: The "event" is the notification that the account data has been changed. The data storage itself is not an event (though there are plenty of bits of spec that might suggest otherwise).

```suggestion
global [account data](https://spec.matrix.org/v1.15/client-server-api/#client-config) and can, 
among other things, be used to generate recommendations
```

content/client-server-api/modules/recent_emoji.md
richvdh: My interpretation of the MSC was that entries with an invalid value of `emoji` should be ignored and preserved, while entries that cannot be parsed or that have -ve usage counts should be dropped.

content/client-server-api/modules/recent_emoji.md
Johennes: Sorry, I thought we needed the prefix for it to be correctly anchored in https://playground.matrix.org but I probably mixed that up?

content/client-server-api/modules/recent_emoji.md
Johennes: Good point, thanks.

content/client-server-api/modules/recent_emoji.md
Johennes: Ah, so the schema is fixed. We just reserve the possibility for `emoji` to contain something else than an emoji. I think that makes sense. Thanks for explaining my MSC to me. 😅 

How about this?

```suggestion
To enable future extension, clients MUST tolerate and preserve array elements
within `recent_emoji` regardless of whether they understand or support the
contained `emoji` value. This means ignoring entries with unrecognised values
of `emoji` when deciding what to display to the user while retaining them when
modifying the array (unless the modification is for truncation).

To prevent undefined behavior, clients SHOULD remove array elements that
don't conform to the event schema such as elements with negative counters.
```

content/client-server-api/modules/recent_emoji.md
richvdh: we do need the prefix in openapi definitions (in `/data/api`, etc) but this content just ends up in the spec text, so no need for a prefix.

content/client-server-api/modules/recent_emoji.md
richvdh: s/This means ignoring the entries/This means ignoring entries with unrecognised values of `emoji`/

Otherwise yes sgtm

content/client-server-api/modules/recent_emoji.md
Johennes: Ahh, right and of course. 🤦‍♂️ 

Apologies and thanks for explaining.

content/client-server-api/modules/recent_emoji.md
Johennes: Have edited it in my previous comment and committed the suggestion from there. Thanks for helping to resolve this.

content/client-server-api/modules/recent_emoji.md
richvdh: No problem at all; it's not very obvious!

content/client-server-api/modules/recent_emoji.md
```diff
@@ -0,0 +1,41 @@
+### Recently used emoji
+
+{{% added-in v="1.18" %}}
+
+This module enables clients to track a user's cumulated emoji usage across different
+devices. The data is stored in the [`m.recent_emoji`](#mrecent_emoji)
+global [account data](https://spec.matrix.org/v1.15/client-server-api/#client-config) and can, 
+among other things, be used to generate recommendations
+in emoji pickers.
```
richvdh: sorry, my fault

```suggestion
global [account data](#client-config) and can, among other things, be used to
generate recommendations in emoji pickers.
```

richvdh (approved):

PR #2297, opened 2026-01-25
Clarify meaning of floating point `m.room.power_levels`

velikopter:
Not sure if I did this the right way..

### Pull Request Checklist

* [X] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog)
* [X] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#sign-off)
* [ ] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#other-changes)

Signed-off-by: Helix K <vel@riseup.net>

Preview: https://pr2297--matrix-spec-previews.netlify.app

### DIFF
--- a/content/rooms/fragments/v1-floaty-power-levels.md
+++ b/content/rooms/fragments/v1-floaty-power-levels.md
@@ -6,7 +6,7 @@ When the value is a float
  * Second, the value is truncated at the decimal point: `51146.98` becomes `51146`.
 
 Values outside the range represented by IEE754 binary64 (a "double") cause the
-powerlevel event to be rejected, as does `Infinity` and `NaN`.
+powerlevel event to be rejected, as do `Infinity`, `-Infinity` and `NaN`.
 
 For example, this is a valid `m.room.power_levels` event in this room version:
 


content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,34 @@
+
+##### `m.room.power_levels` Events Accept Values as Floats
+
+When the value is a float, anything after the decimal point is removed,
```
uhoreg: There shouldn't be any float values in an event `content`, because events need to be encoded as Canonical JSON when they're hashed, and Canonical JSON doesn't allow floats.

1vel: Room versions 5 and below don't strictly enforce canonical JSON

changelogs/room_versions/newsfragments/2297.clarification
```diff
@@ -0,0 +1 @@
+Clarify floaty powerlevels.
```
richvdh:
```suggestion
Clarify meaning of floating-point powerlevels.
```

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,34 @@
+
+##### `m.room.power_levels` Events Accept Values as Floats
```
richvdh:
```suggestion
##### `m.room.power_levels` events accept values as floats
```

content/rooms/fragments/v1-floaty-power-levels.md
richvdh: This would be worth clarifying in the text.

> Note that, since this room version does not enforce that events comply with the requirements of canonical JSON [link to "Canonical JSON" section in the room version], power levels can be formatted as floats.

... or words to that effect.

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,34 @@
+
+##### `m.room.power_levels` events accept values as floats
+
+When the value is a float, anything after the decimal point is removed,
+making e.g. `5.17`, `5.42`, and `5` functionally identical.
```
richvdh: while you're here, you'd better define how `1.34E3` is handled.

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,34 @@
+
+##### `m.room.power_levels` events accept values as floats
```
richvdh, line 2: I see we already have a section about stringy power levels. Is it the case that strings representing floats are accepted here too? In other words, are `"5.17"` and `"1.34E3"` accepted?

If so, we should probably combine the two sections into a single "`m.room.power_levels` events accept values as strings or floats" section.

If not, we could do with a clarification in the stringy power levels section.

1vel: floating-point powerlevels aren't allowed in strings:

```python
>>> int("5.63")
Traceback (most recent call last):
  File "<python-input-1>", line 1, in <module>
    int("5.63")
    ~~~^^^^^^^^
ValueError: invalid literal for int() with base 10: '5.63'
>>> 
```

I will add a clarification in that section.

content/rooms/fragments/v1-floaty-power-levels.md
tulir:
```python
>>> int(json.loads("1.34E3"))
1340
```

I guess it needs more words to say numbers with exponential notation are unpacked before stripping the decimal part, i.e. `1.345678E3` -> 1345.678 -> 1345

content/rooms/fragments/v1-floaty-power-levels.md
1vel: I believe this is already described well:
> a single base 10 integer, **no float values or decimal points**, optionally with any number of leading zeroes (`"100"`, `"000100"`);

(emphasis mine)

richvdh: Right, yes.

I think the bit that's confusing me is the words "When the value is representative of an integer", which makes it sound like we haven't decided what to if the string contains something that isn't an integer.

I think it should probably say something like:

> For fields where the schema for `m.room.power_levels` specifies an integer, a string complying with the following rules is also permitted:

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,40 @@
...
+
+When the value is a float, anything after the decimal point is removed,
+making e.g. `5.17`, `5.42`, and `5` functionally identical.
```
richvdh: rather than having two separate paragraphs (where one is lost way down below the example), I'd suggest making a clear list of transformations to apply:

```suggestion
When the value is a float
 * First, exponential notation is applied:  `5.114698E4` becomes `51146.98`
 * Second, the value is truncated at the decimal point: `51146.98` becomes `51146`.
```

richvdh:
Question: how is a power level of `99.999999999999999` handled?

velikopter:
> Question: how is a power level of `99.999999999999999` handled?
> 
> Or, since we're talking edge cases, `1e400` ?

```python
>>> int(1e400)
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    int(1e400)
    ~~~^^^^^^^
OverflowError: cannot convert float infinity to integer
>>> int(99.999999999999999)
100
>>> 
```

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,41 @@
+
+##### `m.room.power_levels` events accept values as floats
+
+When the value is a float
+ * First, exponential notation is applied: `5.114698E4` becomes `51146.98`
+ * Second, the value is truncated at the decimal point: `51146.98` becomes `51146`.
+
+Any float greater than or equal to `1e309` must result in the powerlevel event
+being rejected.
```
richvdh: `2e308` is also rejected. The maximum is actually `1.797694....e308`.

Let's instead say something like:

```suggestion
Values outside the range represented by IEE754 binary64 (a "double") cause the powerlevel event to be rejected, as do infinities and NaN.
```

1vel: infinities and NaN aren't actually part of the JSON spec so I don't think that bit should be included.

richvdh: and yet Synapse will happily accept JSON that includes `Infinity` or `NaN`.

richvdh: ... so I feel like we should make it explicit that they are not accepted in power levels.

content/rooms/fragments/v1-floaty-power-levels.md
```diff
@@ -0,0 +1,41 @@
+
+##### `m.room.power_levels` events accept values as floats
+
+When the value is a float
+ * First, exponential notation is applied: `5.114698E4` becomes `51146.98`
+ * Second, the value is truncated at the decimal point: `51146.98` becomes `51146`.
+
+Values outside the range represented by IEE754 binary64 (a "double") cause the
+powerlevel event to be rejected, as does `Infinity` and `NaN`.
```
richvdh:
```suggestion
powerlevel event to be rejected, as do `Infinity`, `-Infinity` and `NaN`.
```

PR #2299, opened 2026-01-26
Spec for MSC4313: Require HTML `<ol>` `start` Attribute support

zecakeh:
Spec for MSC4313

### Pull Request Checklist

Preview: https://pr2299--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/instant_messaging.md
+++ b/content/client-server-api/modules/instant_messaging.md
@@ -84,9 +84,9 @@ Additionally, web clients should ensure that *all* `a` tags get a
 `rel="noopener"` to prevent the target page from referencing the
 client's tab/window.
 
-{{% added-in v="1.18" %}} Clients that support the `ol` tag MUST also support
-the `start` attribute in order to prevent loss of meaning of a message due to
-the numbering of list items.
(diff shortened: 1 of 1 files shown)


content/client-server-api/modules/instant_messaging.md
```diff
@@ -84,6 +84,10 @@ Additionally, web clients should ensure that *all* `a` tags get a
 `rel="noopener"` to prevent the target page from referencing the
 client's tab/window.
 
+{{% added-in v="1.18" %}} Clients that support the `ol` tag MUST also support
+the `start` attribute in order to prevent loss of meaning of a message due to
+the numbering of list items.
+
```
HarHarLinks: Thanks for tackling this!

Note the MSC is only about rendering, since converting input to `formatted_body` is an implementation detail. Maybe something like

```suggestion
{{% added-in v="1.18" %}} Clients that support rendering numbered lists via the `ol` tag MUST also support
the `start` attribute in order to prevent loss of meaning of a message due to
the numbering of list items.

```

zecakeh: Applied in e8faaf461de5583e7bb736e98d8f7691464abc44.

PR #2301, opened 2026-01-27
Spec for MSC4153: Exclude non-cross-signed devices

zecakeh:
### Pull Request Checklist

Preview: https://pr2301--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -178,7 +178,7 @@ Example:
 ```
 
 `ed25519` and `curve25519` keys are used for [device keys](#device-keys).
-Additionally, `ed25519` keys are used for [cross-signing keys](#cross-signing).
+Additionally, `ed25519` keys are used for [cross-signing](#cross-signing).
 
 `signed_curve25519` keys are used for [one-time and fallback keys](#one-time-and-fallback-keys).
 
@@ -760,8 +760,8 @@ The process between Alice and Bob verifying each other would be:
 15. Assuming they match, Alice and Bob's devices each calculate Message
     Authentication Codes (MACs) for:
     * {{% changed-in v="1.18" %}} Each of the keys that they wish the other user
-      to verify (usually their device ed25519 key and their master cross-signing
-      key). The master cross-signing key SHOULD be included when two different
+      to verify (usually their device ed25519 key and their master signing key, 
+      see below). The master signing key SHOULD be included when two different
       users are verifying each other. Verifying individual devices of other
       users is deprecated.
     * The complete list of key IDs that they wish the other user to verify.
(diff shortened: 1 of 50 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -5,6 +5,83 @@ Matrix optionally supports end-to-end encryption, allowing rooms to be
...
+  Alice's cross-signing keys change, Alice's own devices MUST alert her to this
+  fact, and prompt her  to re-cross-sign those devices. If Bob is in an
+  encrypted room with Alice,  Bob's devices SHOULD inform him of Alice's key
+  change and SHOULD prevent him from sending an encrypted message to Alice
+  without acknowledging the change. Bob's clients may behave differently
+  depending on whether Bob had previously [verified](#device-verification)
+  Alice or not. For example, if Bob had previously verified Alice, and Alice's
+  keys change, Bob's client may require Bob to re-verify, or may display a more
+  aggressive warning.
+* Clients SHOULD NOT send encrypted [to-device](#send-to-device-messaging)
```
andybalaam, line 54: If I'm reading this right, this is only talking about to-device messages, but I think it should cover all encrypted events.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -5,6 +5,83 @@ Matrix optionally supports end-to-end encryption, allowing rooms to be
...
+
+While a user may be unable to [verify](#device-verification) every other user
+that they communicate with, or may be unaware of the need to verify other users,
+[cross-signing](#cross-signing) gives some measure of protection and so SHOULD
+be used where possible. In particular, clients SHOULD implement the following
+recommendations.
+
+* Clients SHOULD create new [cross-signing keys](#cross-signing) for users who
+  do not yet have cross-signing keys.
+* Users SHOULD have [Secret Storage](#storage) set up to avoid needing to reset
```
andybalaam: Is it appropriate to use SHOULD to talk about users? I would have thought it would be better to say clients should encourage users or similar.

content/client-server-api/modules/end_to_end_encryption.md
tulir: You can't selectively send room events to specific devices. Withholding the megolm session shared via to-device is the way to cover all encrypted messages in rooms.

It might be worth adding a note explicitly calling that out for extra clarity, but technically it's already completely correct.

andybalaam: Yeah, I was confusing events with sessions. In our implementation, sending sessions was separate from the other to-device messages, but here both are covered in this bullet point and I am talking rubbish :-)

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -5,6 +5,83 @@ Matrix optionally supports end-to-end encryption, allowing rooms to be
...
+  keys.
+* Clients MAY make provisions for encrypted bridges. Some bridges are structured
+  in a way such that only one user controlled by the bridge (often called the
+  bridge bot) participates in encryption, and encrypted messages from other
+  bridge users are encrypted by the bridge bot. Thus encrypted messages sent by
+  one user could be encrypted by a [Megolm](#mmegolmv1aes-sha2) session sent by
+  a different user. Clients MAY accept such messages, provided the session
+  creator's device is [cross-signed](#cross-signing). However, the client MUST
+  annotate the message with a warning, unless the client has a way to check that
+  the bridge bot is permitted to encrypt messages on behalf of the user.
```
tulir: Probably worth linking to https://github.com/matrix-org/matrix-spec-proposals/pull/4350 with a sentence along the lines of "Future MSCs such as MSC4350 may provide a secure way to allow such impersonation". The spec already links to future MSCs in a couple of places

content/client-server-api/modules/end_to_end_encryption.md
zecakeh: Added a note in 2fca4789ca537302c786c8dc87b3a684c4b31f30.

content/client-server-api/modules/end_to_end_encryption.md
zecakeh: Changed in 2fca4789ca537302c786c8dc87b3a684c4b31f30.

content/client-server-api/modules/end_to_end_encryption.md
zecakeh: Added in 2fca4789ca537302c786c8dc87b3a684c4b31f30.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -674,8 +759,11 @@ The process between Alice and Bob verifying each other would be:
     their devices if they match or not.
 15. Assuming they match, Alice and Bob's devices each calculate Message
     Authentication Codes (MACs) for:
-    * Each of the keys that they wish the other user to verify (usually their
-      device ed25519 key and their master cross-signing key).
+    * {{% changed-in v="1.18" %}} Each of the keys that they wish the other user
+      to verify (usually their device ed25519 key and their master cross-signing
+      key). The master cross-signing key SHOULD be included when two different
+      users are verifying each other. Verifying individual devices of other
+      users is deprecated.
```
richvdh, line 766: It's worth noting that this flow is used when verifying individual devices of the *same* user (i.e. when Alice and Bob are the same user on different devices), and I wonder if this sentence makes that unclear.

richvdh: I don't feel strongly enough about this to hold up the PR.

PR #2307, opened 2026-02-03
Add identifier pronunciation

HarHarLinks:
This is more of an info than normative text, but it seems appropriate to add it in this area.

Source are the [upcoming brand guidelines](https://github.com/matrix-org/matrix.org/issues/2966#issuecomment-3794989840) of the Foundation, which in this case are based on a really old but to my knowledge not contradicted presentation https://matrix.org/blog/wp-content/uploads/2014/09/Introducing-matrix.pdf (page 11).

We could also put this in an infobox etc if that seems appropriate to you?

### Pull Request Checklist

Preview: https://pr2307--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/appendices/newsfragments/2307.clarification
+++ b/changelogs/appendices/newsfragments/2307.clarification
@@ -1,0 +1,1 @@
+Add identifier pronunciation guidelines. Contributed by @HarHarLinks.


anoadragon453 (approved):
Useful, thank you!

I can confirm that this seems to be what people have adopted canonically in IRL discussions.

PR #2311, opened 2026-02-12
Spec for MSC4277: Harmonizing the reporting endpoints

Johennes:
### Pull Request Checklist

Preview: https://pr2311--matrix-spec-previews.netlify.app

### DIFF
--- a/data/api/client-server/report_content.yaml
+++ b/data/api/client-server/report_content.yaml
@@ -105,9 +105,9 @@ paths:
         the appropriate people. The caller must be joined to the room to report
         it.
 
-        Clients could infer whether a reported event exists based on the 404 response.
-        Homeservers that wish to conceal this information MAY return 200 responses
-        regardless of the existence of the reported event.
+        Clients could infer whether a reported event or room exists based on the 404
+        response. Homeservers that wish to conceal this information MAY return 200
+        responses regardless of the existence of the reported event or room.
 
         Furthermore, it might be possible for clients to deduce whether a reported
         event exists by timing the response. This is because only a report for an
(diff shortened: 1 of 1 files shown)


data/api/client-server/report_content.yaml
```diff
@@ -136,6 +148,10 @@ paths:
       x-changedInMatrixVersion:
         1.8: |
           This endpoint now requires the user to be joined to the room.
+        1.18: |
+          The `score` request parameter was removed. Additionally, servers
+          may deny event ID enumeration by using the 200 response
```
richvdh:
```suggestion
          may prevent event ID enumeration by using the 200 response
```

(The word 'deny' sounds a bit more active -- like the server should actively monitor for clients that are enumerating event IDs and return a 4xx or something. 'Prevent' reads a bit more passive.)

richvdh: The MSC isn't explicit about it, but can servers apply the same logic to non-existent *room* IDs?

data/api/client-server/report_content.yaml
```diff
@@ -52,6 +61,11 @@ paths:
       security:
         - accessTokenQuery: []
         - accessTokenBearer: []
+      x-changedInMatrixVersion:
+        1.18: |
+          Servers MAY deny room ID enumeration by using the 200 response
```
richvdh:
```suggestion
          Servers MAY prevent room ID enumeration by using the 200 response
```

data/api/client-server/report_content.yaml
Johennes: Hm, yeah, I guess by extension of the same argument, that would make a lot of sense. I hadn't thought of that before.

How do we deal with the MSC missing this?

richvdh: the MSC just says "MAY respond with 200 and no content regardless of whether the *reported subject* exists". I think we can squint at that "reported subject" and claim that it allows non-existent room IDs here.

Johennes: Oh, good point. Have added this change with fd9989820a7f19dadf5d433ce7c9c9432009f543.

richvdh (approved):
lgtm

PR #2315, opened 2026-02-20
Spec for MSC4335: M_USER_LIMIT_EXCEEDED error code

hughns:
For MSC4335.

### Pull Request Checklist

Preview: https://pr2315--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -148,35 +148,35 @@ only read state (e.g.: [`/sync`](#get_matrixclientv3sync),
 [`/user/{userId}/account_data/{type}`](#get_matrixclientv3useruseridaccount_datatype), etc).
 
 `M_USER_LIMIT_EXCEEDED`
-{{% added-in v="1.18" %}} The request cannot be completed because the user has
+: {{% added-in v="1.18" %}} The request cannot be completed because the user has
 exceeded (or the request would cause them to exceed) a limit associated with
 their account. For example, a user may have reached their allocated storage
 quota, reached a maximum number of allowed rooms, devices, or other
 account-scoped resources, or exceeded usage limits for specific features.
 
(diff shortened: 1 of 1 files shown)


content/client-server-api/_index.md
```diff
@@ -147,6 +147,37 @@ state (e.g.: sending messages, account data, etc) and not routes which
 only read state (e.g.: [`/sync`](#get_matrixclientv3sync),
 [`/user/{userId}/account_data/{type}`](#get_matrixclientv3useruseridaccount_datatype), etc).
 
+`M_USER_LIMIT_EXCEEDED`
```
hughns, line 150: Is this too long for this section of the doc?

I personally think the details and example are helpful, but they make the others look sparse!

I could add an extra section like `M_LIMIT_EXCEEDED` has?

richvdh: I think the main problem is that it's easy to miss the fact that the whole thing, including the example response, is part of `M_USER_LIMIT_EXCEEDED`.

I think we could solve this by *either* 1. reformatting the section to make it clearer (one option would be to use `<dl>` items, as in the [Push notifications](https://pr2315--matrix-spec-previews.netlify.app/client-server-api/#push-notifications) section), or 2. moving the body of the definition out to another section, as you suggest.

I have no strong opinion on which is preferable.

hughns: I think the `<dl>` works well. I've drafted https://github.com/matrix-org/matrix-spec/pull/2323 to explore this. If that looks like a good direction then I would rebase on that approach.

richvdh:
> One other thought (which I would do in a follow-up PR): would it make more sense if these error codes were sorted alphabetically?

I think it probably would, tbh, yes.

PR #2318, opened 2026-02-20
Typos and clarifications

HarHarLinks:
### Pull Request Checklist

Preview: https://pr2318--matrix-spec-previews.netlify.app

### DIFF
--- a/data/event-schemas/schema/core-event-schema/event.yaml
+++ b/data/event-schemas/schema/core-event-schema/event.yaml
@@ -7,7 +7,7 @@ properties:
       When interacting with the REST API, this is the HTTP body.
     type: object
   type:
-    description: The type of event, as defined by [the event type specification](#types-of-room-events).
+    description: The type of event, as defined by [the event type specification](/client-server-api/#types-of-room-events).
     type: string
 required:
   - type


layouts/_partials/events/render-event.html
```diff
@@ -44,7 +44,7 @@ <h1 id="{{ $anchor }}">
 
 <table class="basic-info">
   <tr>
-   <th>Event type:</th>
+   <th>Event type</th>
```
HarHarLinks: related, there is also this questionable `.` at the end of the state key when the spec using this template deems it right:
image
Feels a bit inconsistent but not clear enough to me to definitely fix if intended to be a description even though it's not a full sentence.

data/event-schemas/schema/core-event-schema/event.yaml
```diff
@@ -8,7 +8,7 @@ properties:
     type: object
   type:
     description: The type of event. This SHOULD be namespaced similar to Java package
-      naming conventions e.g. 'com.example.subdomain.event.type'
+      naming conventions e.g. `com.example.subdomain.event.type`
```
HarHarLinks: arguably this whole explanation is in the wrong place here since how event types work is defined elsewhere. Would you agree to me changing the description respectively?

content/client-server-api/modules/threading.md
```diff
@@ -107,7 +107,7 @@ flag to `true`.
 ```
 
 {{% boxes/note %}}
-Clients which are acutely aware of threads (they do not render threads, but are otherwise
+Clients which are aware of threads (they do not render threads, but are otherwise
```
HarHarLinks, line 110: I asked a native speaker and we concluded that "acutely" does not add anything but confusion here.

changelogs/client_server/newsfragments/2318.clarification
```diff
@@ -0,0 +1 @@
+Fix typos, formatting, wording. Contributed by @HarHarLinks.
```
richvdh: Unless you're particularly keen for your own entry in the changelog (and if you are, you're welcome to leave this), we normally go with "Fix various typos throughout the specification." for this sort of thing.

```suggestion
Fix various typos throughout the specification.
```

data/event-schemas/schema/core-event-schema/event.yaml
richvdh: > Would you agree to me changing the description respectively?

sure, this isn't very elegant as things stand.

changelogs/internal/newsfragments/2318.clarification
```diff
@@ -0,0 +1 @@
+Fix typos, formatting, wording. Contributed by @HarHarLinks.
```
richvdh: probably no need for a separate `internal` changelog here

HarHarLinks: I added this because I technically modified the footer in config/_default/hugo.toml 🤷

data/event-schemas/schema/core-event-schema/event.yaml
HarHarLinks: done in a34a6f8c

PR #2319, opened 2026-02-22
Spec for MSC4376: Remove /v1/send_join and /v1/send_leave

### LINKED ISSUE #1514: S2S v1/send_join: What is the integer in the response?
**Link to problem area**: https://spec.matrix.org/v1.6/server-server-api/#put_matrixfederationv1send_joinroomideventid

**Issue**

Spec says:

> 200 response
> 
> Array of integer, Room State.

... however, only the Room State is specified, and it's not specified what the `integer` is exactly. Is it always `200`, like in the example?

### LINKED ISSUE #1515: S2S v1 send_join: `event` response property not documented
**Link to problem area**: https://spec.matrix.org/v1.6/server-server-api/#put_matrixfederationv1send_joinroomideventid

**Issue**

Aside from `auth_chain`, `origin`, and `state`, the example response also shows an `event` property, but it is not specified what this is (although it *seems* to be the signed event generated by the resident homeserver).

As this is top-level in the response and not part of an event, it's not already covered by the various "event format may change" warnings, and indeed this does not seem to be specified in the room version spec either.

velikopter:
Specs MSC4376

fixes #1514
fixes #1515

### Pull Request Checklist

Signed-off-by: Kierre Sametti <vel@riseup.net>

Preview: https://pr2319--matrix-spec-previews.netlify.app


data/api/server-server/joins-v2.yaml
1vel, line 1: should I add a note about where to find the removed v1 API for server developers interested in that?

richvdh: I don't think that's necessary. A developer writing a client from scratch shouldn't need to care. A developer maintaining an existing implementation will know where to find older versions of the spec.

PR #2320, opened 2026-02-23
Spec for MSC4341: Support for RFC 8628 Device Authorization Grant

hughns:
For MSC4341.

### Pull Request Checklist

Preview: https://pr2320--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -1729,12 +1729,10 @@ authentication type.
 #### Login flows
 
 Logging in and obtaining an access token with the OAuth 2.0 API should be done
-using either the [authorization code flow](#authorization-code-flow) or
-[device authorization flow](#device-authorization-flow).
-
-The client allocates a device ID and then requests authorization for a
-[scope](#scope) that includes full Matrix client-server API read/write access
-and use of the device ID.
+using either the [authorization code grant](#authorization-code-grant) or
+[device authorization grant](#device-authorization-grant). In the context of the
+Matrix specification, this means requesting a [scope](#scope) including full
+client-server API read/write access and allocating a device ID.
 
 ##### Authorization code flow
 


content/client-server-api/_index.md
```diff
@@ -1873,6 +1884,142 @@ Sample response:
 Finally, the client can call the  [`/whoami`](#get_matrixclientv3accountwhoami)
 endpoint to get the user ID that owns the access token.
 
+#### Device authorization flow
+
+{{% added-in v="1.18" %}}
+
+The device authorization flow allows clients on devices with limited input
+capabilities (such as CLI applications or embedded devices) to obtain an
```
sandhose: I would argue that it's also useful for native desktop apps. I can see how Element Web could switch to it on desktop, as sometimes the redirect isn't reliable

content/client-server-api/_index.md
```diff
@@ -1873,6 +1884,142 @@ Sample response:
...
+[RFC 8628 section 3.2](https://datatracker.ietf.org/doc/html/rfc8628#section-3.2),
+containing:
+
+| Parameter                    |                                                                                                                                               |
+|------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------|
+| `device_code`                | The device verification code.                                                                                                                 |
+| `user_code`                  | An end-user verification code.                                                                                                                |
+| `verification_uri`           | The end-user verification URI on the authorization server.                                                                                    |
+| `verification_uri_complete`  | Optionally, the URI including the `user_code`, so the user does not need to type it in manually.                                              |
```
sandhose: It technically doesn't require the `user_code` in the URI but rather information that substitutes the end-user entering that code

The RFC says

> OPTIONAL.  A verification URI that includes the "user_code" (or
>      other information with the same function as the "user_code"),
>      which is designed for non-textual transmission.


```suggestion
| `verification_uri_complete`  | Optionally, the URI which doesn't require the user to manually type the `user_code`, designed for non-textual transmission.                   |
```

data/api/client-server/oauth_server_metadata.yaml
```diff
@@ -67,8 +67,8 @@ paths:
                     type: string
                     format: uri
                     description: |-
-                      URL of the token endpoint, necessary to use the authorization code grant and
-                      the refresh token grant.
+                      URL of the token endpoint, necessary to use the authorization code grant,
+                      device authorization grant and the refresh token grant.
```
sandhose:
```suggestion
                      URL of the token endpoint, necessary to use the authorization code grant,
                      device authorization grant and refresh token grant.
```

or


```suggestion
                      URL of the token endpoint, necessary to use the authorization code grant,
                      the device authorization grant and the refresh token grant.
```

hughns: Re-reading, I'm not sure the is any benefit in listing the grant types out? i.e. the token endpoint is required for all grant types that Matrix supports.

```suggestion
                      URL of the token endpoint, used during the grants.
```

content/client-server-api/_index.md
hughns: Good point. In 68732e76cf52201be16edbee8d8576443f92153b I have attempted to cover these use cases too.

data/api/client-server/oauth_server_metadata.yaml
sandhose: That's fair, or even `used by different grants`? The RFC/IANA registry is pretty vague anyway:

> URL of the authorization server's token endpoint

content/client-server-api/_index.md
```diff
@@ -1720,14 +1720,25 @@ authentication type.
...
+   - Use the [authorization code grant](#authorization-code-grant) via the
+     [login flow](#login-flow) for clients with access to a web browser.
+   - {{% added-in v="1.18" %}} Use the [device authorization grant](#device-authorization-grant) via the
+     [device authorization flow](#device-authorization-flow) for clients on
+     devices with limited input capabilities (e.g. CLI applications or embedded devices)
+     where the user completes authorization on a separate device.
 4. [Refresh the access token](#token-refresh-flow) with the [refresh token grant](#refresh-token-grant) when it expires.
 5. [Revoke the tokens](#token-revocation) when the users wants to log out of the client.
 
 #### Login flow
```
zecakeh: I believe that this section should be renamed to "Login flows" (plural) and contain sub-sections for the authorization code flow and the device authorization flow. The device authorization flow shouldn't be a separate section, because it is used for logging in too.

IIRC, this section uses the "login" term to be easy to find by developers that want to know how to log in with this API, and it was always intended for this section to be split into subsections for the different grant types that allow to log in. It just didn't make sense when it was introduced because there was only a single supported grant type for this.

hughns: I did wonder this when I started making the PR. I will refactor into this to see how it feels.

data/api/client-server/oauth_server_metadata.yaml
hughns:
```suggestion
                      URL of the token endpoint, used by the grants.
```

content/client-server-api/_index.md
hughns: Attempted in 3b2f11bd7afb04f01c03c5c02236aa36ee3da69f

content/client-server-api/_index.md
```diff
@@ -1720,19 +1720,33 @@ authentication type.
...
+The client allocates a device ID and then requests authorization for a
+[scope](#scope) that includes full Matrix client-server API read/write access
+and use of the device ID.
+
+##### Authorization code flow
+
+This login flow uses the [authorization code grant](#authorization-code-grant)
+and is suitable for clients where the following criteria are met:
+
+- The is a web browser available for the user to complete authentication and
```
zecakeh: I believe this is a typo:

```suggestion
- There is a web browser available for the user to complete authentication and
```

content/client-server-api/_index.md
```diff
@@ -1720,19 +1720,33 @@ authentication type.
...
-grant](#authorization-code-grant). In the context of the Matrix specification,
-this means requesting a [scope](#scope) including full client-server API
-read/write access and allocating a device ID.
+Logging in and obtaining an access token with the OAuth 2.0 API should be done
+using either the [authorization code flow](#authorization-code-flow) or
+[device authorization flow](#device-authorization-flow).
 
-Once the client has retrieved the [server metadata](#server-metadata-discovery),
-it needs to generate the following values:
+The client allocates a device ID and then requests authorization for a
```
zecakeh: I believe that the previous wording made more sense. The actual "allocation" is made when the scope is sent to the server, not before.

So I'd suggest to keep the previous wording or replace "allocates" with "creates" or something similar.

zecakeh (changes requested):

content/client-server-api/_index.md
hughns: You are indeed correct. I've moved it back closer to the original in 6f39b946fd2c34ed5a9db81ce064a4b6f36254f1

PR #2324, opened 2026-02-28
Clarify what clients are required to do for attachment encryption

tulir:
This patches some of the gripes with the spec brought up in https://lotte.chir.rs/2024/08/18/Missing-Salamanders-Matrix-Media-can-be-decrypted-to-multiple-valid-plaintexts-using-different-keys/. The actual vulnerability isn't fixed, as that would require a new hash (or extending the existing hash) to cover the key and iv fields, which requires an MSC.

In particular, most `should`s have been removed and replaced with imperatives or MUSTs. Generating unique AES keys and IVs and only using them once is a MUST, as is zeroing out the latter 64 bits of the counter (to ensure that there's no confusion about what is a unique IV/counter).

Additionally, verifying the hash before using the content is a MUST now. I'm pretty sure that was always the idea, it just wasn't written down.

Preview: https://pr2324--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -270,9 +270,9 @@ key, and encrypts the file using AES-CTR. The counter is 64 bits long,
 starting at 0 and prefixed by a random 64-bit Initialization Vector (IV),
 which together form a 128-bit unique counter block.
 
-Clients MUST generate both the AES key and IV using a secure random
-source and MUST NOT use the same key or IV multiple times. The latter
-64 bits of the 128-bit counter block MUST start at zero.
+Clients MUST generate both the AES key and IV using a cryptographically 
+secure random source and MUST NOT use the same key or IV multiple
+times. The latter 64 bits of the 128-bit counter block MUST start at zero.
 
 {{% boxes/warning %}}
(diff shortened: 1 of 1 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -350,11 +265,14 @@ field, and as a result should remove her from its list of tracked users.
...
-Initialization Vector (IV), which together form a 128-bit unique counter
-block.
+In order to achieve this, the client generates a single-use 256-bit AES
+key, and encrypts the file using AES-CTR. The counter is 64 bits long,
+starting at 0 and prefixed by a random 64-bit Initialization Vector (IV),
+which together form a 128-bit unique counter block.
+
+Clients MUST generate both the AES key and IV using a secure random
+source and MUST NOT use the same key or IV multiple times. The latter
+64 bits of the 128-bit counter block MUST start at zero.
```
dkasak:
```suggestion
Clients MUST generate both the AES key and IV using a cryptographically 
secure random source and MUST NOT use the same key or IV multiple
times. The latter 64 bits of the 128-bit counter block MUST start at zero.
```

dkasak: Say "cryptographically" to make it clearer what "secure" means

PR #2206, opened 2025-09-05
Add time zone profile field from MSC4175

clokep:
### Pull Request Checklist

Spec PR for MSC4175

* [x] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)
* [x] Pull request includes a [sign off](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#sign-off)
* [x] Pull request is classified as ['other changes'](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#other-changes)

Preview: https://pr2206--matrix-spec-previews.netlify.app

PR #2326, opened 2026-03-02
s2s/query: clarify profile behaviour and API responses

famfo:
Since the existing specification of the behavior is relatively fuzzy, I have chosen the stricter interpretations from multiple places saying something similar.

### Pull Request Checklist

Preview: https://pr2326--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/2326.clarification
+++ b/changelogs/server_server/newsfragments/2326.clarification
@@ -1,0 +1,1 @@
+Clarify the s2s profile query behaviour and responses.
--- a/changelogs/server_server/newsfragments/2326.clarification
+++ b/changelogs/server_server/newsfragments/2326.clarification
@@ -1,1 +1,1 @@
-Clarify the s2s profile query behaviour and responses.
+Clarify the behaviour and response format of `GET /_matrix/federation/v1/query/profile`.
--- a/data/api/server-server/query.yaml
+++ b/data/api/server-server/query.yaml
@@ -115,12 +115,12 @@ paths:
         to the target server (identified by the [server name](/appendices/#server-name)
         in the user ID).
 
-        Responding servers MAY
-          - allow users to set arbitrary key/value pairs in their profile in addition to the
-            specified pairs
-          - deny profile look-up over federation by responding with 403 and an error code of
-            `M_FORBIDDEN`
-          - omit certain key/value pairs in the response
+        Responding servers MAY:
+          - Allow users to set arbitrary key/value pairs in their profile in addition to the
+            specified pairs.
+          - Deny profile look-up over federation by responding with 403 and an error code of
+            `M_FORBIDDEN`.
+          - Omit certain key/value pairs in the response.
 
         Requesting servers MAY wish to cache the response to this query to avoid requesting the
         information too often.
@@ -149,10 +149,12 @@ paths:
       responses:
         "200":
           description: |-
-            The profile for the user. If a `field` is specified in the request, only the
-            matching field MUST included in the response. If no `field` was specified,
-            the response MUST include the fields of the user's profile that can be made
-            public, such as the display name and avatar.
+            The profile for the user. If a `field` is specified in the request, the response
+            MUST only included the specified `field`. If no `field` was specified, the response
+            SHOULD include all of the fields of the user's profile, which can be made public.
+
+            If a field in the user's profile can't be made public over the federation, the
+            responding server MAY choose to exclude it in the response.
 
             If the user does not have a particular field set on their profile, the server
             MUST either exclude it from the response body or give it the value `null`.
@@ -175,8 +177,10 @@ paths:
                     example: mxc://example.org/MyC00lAvatar
                   m.tz:
                     type: string
-                    description: The name of the user's time zone. The name MUST be registered in
-                      the [IANA Time Zone Database](https://www.iana.org/time-zones).
+                    description: The name of the user's time zone. The name SHOULD be registered in
+                      the [IANA Time Zone Database](https://www.iana.org/time-zones), requesting
+                      servers MUST tolerate invalid or unknown values.
+                    x-addedInMatrixVersion: "1.16"
                 additionalProperties:
                   description: Additional profile fields.
               examples:
--- a/data/api/server-server/query.yaml
+++ b/data/api/server-server/query.yaml
@@ -151,7 +151,8 @@ paths:
           description: |-
             The profile for the user. If a `field` is specified in the request, the response
             MUST only include the specified `field`. If no `field` was specified, the response
-            SHOULD include all of the fields of the user's profile, which can be made public.
+            SHOULD include all of the fields of the user's profile that can be made public, such
+            as the display name and avatar.
 
             If a field in the user's profile can't be made public over the federation, the
             responding server MAY choose to exclude it in the response.
--- a/data/api/server-server/query.yaml
+++ b/data/api/server-server/query.yaml
@@ -111,22 +111,26 @@ paths:
       summary: Query for profile information about a given user
       description: |-
         Performs a query to get profile information, such as a display name or avatar,
-        for a given user. Homeservers should only query profiles for users that belong
+        for a given user. Homeservers MUST only query profiles for users that belong
         to the target server (identified by the [server name](/appendices/#server-name)
         in the user ID).
 
-        Servers may wish to cache the response to this query to avoid requesting the
-        information too often.
+        Responding servers MAY
+          - allow users to set arbitrary key/value pairs in their profile in addition to the
+            specified pairs
+          - deny profile look-up over federation by responding with 403 and an error code of
+            `M_FORBIDDEN`
+          - omit certain key/value pairs in the response
 
-        Servers MAY deny profile look-up over federation by responding with 403 and an
-        error code of `M_FORBIDDEN`.
+        Requesting servers MAY wish to cache the response to this query to avoid requesting the
+        information too often.
       operationId: queryProfile
       security:
         - signedRequest: []
       parameters:
         - in: query
           name: user_id
-          description: The user ID to query. Must be a user local to the receiving homeserver.
+          description: The user ID to query. MUST be a user local to the receiving homeserver.
           required: true
           example: "@someone:example.org"
           schema:
@@ -134,24 +138,24 @@ paths:
         - in: query
           name: field
           description: |-
-            The field to query. If specified, the server will only return the given field
-            in the response. If not specified, the server will return the full profile for
-            the user.
+            The field of the profile to query. If specified, the server MUST only return the
+            given field in the response. If not specified, the server MUST return the full,
+            public, profile for the user.
+
+            Defined values are `displayname`, `avatar_url` and `m.tz`. In addition to these
+            servers MAY allow users to set additional key/value pairs.
           schema:
             type: string
-            enum:
-              - displayname
-              - avatar_url
       responses:
         "200":
           description: |-
             The profile for the user. If a `field` is specified in the request, only the
-            matching field should be included in the response. If no `field` was specified,
-            the response should include the fields of the user's profile that can be made
+            matching field MUST included in the response. If no `field` was specified,
+            the response MUST include the fields of the user's profile that can be made
             public, such as the display name and avatar.
 
             If the user does not have a particular field set on their profile, the server
-            should exclude it from the response body or give it the value `null`.
+            MUST either exclude it from the response body or give it the value `null`.
           content:
             application/json:
               schema:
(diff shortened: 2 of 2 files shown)


data/api/server-server/query.yaml
```diff
@@ -146,12 +146,12 @@ paths:
...
+            matching field MUST included in the response. If no `field` was specified,
+            the response MUST include the fields of the user's profile that can be made
             public, such as the display name and avatar.
 
             If the user does not have a particular field set on their profile, the server
-            should exclude it from the response body or give it the value `null`.
+            MUST either exclude it from the response body or give it the value `null`.
           content:
             application/json:
               schema:
```
zecakeh, line 164: While we are touching this endpoint, I believe that the schema of the response should be updated to match the response of `GET /_matrix/client/v3/profile/{userId}`, i.e. add support for extended profile fields and `m.tz`.

data/api/server-server/query.yaml
zecakeh: While we are touching this endpoint, I believe that we should remove this `enum` definition, because it shows as:

> One of: [displayname, avatar_url].

But with extended profile fields, more values are possible.

data/api/server-server/query.yaml
famfo: Judging from the MSC4133 text on [errors and missing keys](https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4133-extended-profiles.md#404-not-found-target-cannot-be-found) this API should also return 404 instead on not found keys in the profile?

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,51 @@ paths:
...
+        Responding servers MAY
+          - allow users to set arbitrary key/value pairs in their profile in addition to the
+            specified pairs
+          - deny profile look-up over federation by responding with 403 and an error code of
+            `M_FORBIDDEN`
+          - omit certain key/value pairs in the response
 
-        Servers MAY deny profile look-up over federation by responding with 403 and an
-        error code of `M_FORBIDDEN`.
+        Requseting servers MAY wish to cache the response to this query to avoid requesting the
```
HarHarLinks:
```suggestion
        Requesting servers MAY wish to cache the response to this query to avoid requesting the
```

data/api/server-server/query.yaml
```diff
@@ -160,20 +164,20 @@ paths:
                   displayname:
                     type: string
                     description: |-
-                      The display name of the user. May be omitted if the user does not have a
-                      display name set.
+                      The display name of the user. MUST either be omitted or set to `null` if
+                      the user does not have a display name set.
```
richvdh, line 171: Aside: I was a bit surprised by this, but then I found https://github.com/matrix-org/matrix-spec/commit/d914c402e24e69536902703a305dbf4391c57303, and this seems to be just a clarification of that changes.

richvdh (dismissed):
LGTM

data/api/server-server/query.yaml
richvdh: @famfo are you planning to address this comment?

famfo: I have already changed this [here](https://github.com/matrix-org/matrix-spec/compare/5790b2f99036d156a2db4dd007bc5e0bdb811194..0608624f1d75b04acae79d6ed2b3b22d2a1f60cf?diff=unified&w):

```diff
        "404":
-         description: The user does not exist or does not have a profile.
+         description: The user does not exist, does not have a profile or the queried field does not exist.
          content:
```

richvdh: @zecakeh is that what you intended here?

zecakeh: I was also thinking about the `m.tz` field and the `<Other properties>` that appear in the CS API.

Actually it would probably be best if we could share the response schema between the CS and SS APIs because they should always be identical, but given the current code organization I wouldn't know where to put it.

famfo: I guess the most "correct" place would currently be the Appendix. A "common" section where some of these schemas are defined would probably help. This also applies to stuff like errors, they are just referencing the client-server API definition without further explanation.

I think that shuffling around a bunch of schemas is also a bit out of scope for this PR though.

zecakeh: Sure, then you should add those missing fields here.

famfo: I thought I did already? Should I change this again?

https://github.com/matrix-org/matrix-spec/pull/2326/changes#diff-1e41c6a0edb6b363a5318eb03b12f79e3620ac4fbcb17ae26b69bb21da95b822R145-R146

zecakeh: You mentioned the fields in the request parameters but not on the schema of the response, the one pointed to by this thread.

zecakeh (approved):

richvdh:
@famfo please don't force-push between rounds of review: it makes it very hard to see what has changed since the previous round.

changelogs/server_server/newsfragments/2326.clarification
```diff
@@ -0,0 +1 @@
+Clarify the s2s profile query behaviour and responses.
```
richvdh: "s2s" isn't a term that is used in the spec. Since this is the changelog for the server-server spec, it's somewhat redundant anyway.

```suggestion
Clarify the behaviour and response format of `GET /_matrix/federation/v1/query/profile`.
```

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,51 @@ paths:
...
         in the user ID).
 
-        Servers may wish to cache the response to this query to avoid requesting the
-        information too often.
+        Responding servers MAY
+          - allow users to set arbitrary key/value pairs in their profile in addition to the
+            specified pairs
+          - deny profile look-up over federation by responding with 403 and an error code of
+            `M_FORBIDDEN`
+          - omit certain key/value pairs in the response
```
richvdh: Per https://github.com/matrix-org/matrix-spec/blob/main/meta/documentation_style.rst#general:

```suggestion
          - Allow users to set arbitrary key/value pairs in their profile in addition to the
            specified pairs.
          - Deny profile look-up over federation by responding with 403 and an error code of
            `M_FORBIDDEN`.
          - Omit certain key/value pairs in the response.
```

data/api/server-server/query.yaml
```diff
@@ -160,20 +164,26 @@ paths:
...
-                      The avatar URL for the user's avatar. May be omitted if the user does not
-                      have an avatar set.
-                    example: mxc://matrix.org/MyC00lAvatar
+                      The avatar URL for the user's avatar. MUST either be omitted or set to
+                      `null` if the user does not have an avatar set.
+                    example: mxc://example.org/MyC00lAvatar
+                  m.tz:
+                    type: string
+                    description: The name of the user's time zone. The name MUST be registered in
+                      the [IANA Time Zone Database](https://www.iana.org/time-zones).
```
richvdh: It would be good to give this an `x-addedInMatrixVersion` tag. Since `m.tz` was added in spec v1.16 (it looks like this change got forgotten in https://github.com/matrix-org/matrix-spec/pull/2206), we should probably set it to `"1.16"`.

richvdh: I don't think the MUST part here is correct, for a couple of reasons:
 - A client might have set the value on a server that does not know about m.tz.
 - Per MSC4175, servers are not obliged to validate the value that clients set anyway

It would probably be better to say "SHOULD be registered in the IANA Time Zone Database, but receiving servers MUST tolerate invalid or unknown values."

richvdh (changes requested):

data/api/server-server/query.yaml
clokep: Why is this endpoint talking about setting data? Shouldn't that only be the CS docs?

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,51 @@ paths:
...
-              - displayname
-              - avatar_url
       responses:
         "200":
           description: |-
             The profile for the user. If a `field` is specified in the request, only the
-            matching field should be included in the response. If no `field` was specified,
-            the response should include the fields of the user's profile that can be made
+            matching field MUST included in the response. If no `field` was specified,
+            the response MUST include the fields of the user's profile that can be made
```
clokep: I think there's some missing words:

```suggestion
            matching field MUST be included in the response. If no `field` was specified,
            the response MUST include all of the fields of the user's profile that can be made
```

I'm still not sure about the fist sentence. I think you're trying to say "the response MUST only include that field", but I can't make that change as a suggestion.

data/api/server-server/query.yaml
famfo: To set expectations for requesting servers. A server doesn't have to implement the c2s API.

data/api/server-server/query.yaml
famfo: should the other fields also get a `x-addedInMatrixVersion: "1.16"`?

data/api/server-server/query.yaml
richvdh: @clokep leaving this thread open for you to assess if you're satisfied with the answer. I agree it's not terribly clear as written here but I think it's an improvement over the current spec and I don't have any immediate suggestions for improvements.

data/api/server-server/query.yaml
richvdh: well no, because they haven't been newly added to the spec?

data/api/server-server/query.yaml
```diff
@@ -160,20 +166,28 @@ paths:
...
-                      The avatar URL for the user's avatar. May be omitted if the user does not
-                      have an avatar set.
-                    example: mxc://matrix.org/MyC00lAvatar
+                      The avatar URL for the user's avatar. MUST either be omitted or set to
+                      `null` if the user does not have an avatar set.
+                    example: mxc://example.org/MyC00lAvatar
+                  m.tz:
+                    type: string
+                    description: The name of the user's time zone. The name SHOULD be registered in
+                      the [IANA Time Zone Database](https://www.iana.org/time-zones), requesting
```
richvdh: Comma splice

```suggestion
                      the [IANA Time Zone Database](https://www.iana.org/time-zones). Requesting
```

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,53 @@ paths:
...
-              - avatar_url
       responses:
         "200":
           description: |-
-            The profile for the user. If a `field` is specified in the request, only the
-            matching field should be included in the response. If no `field` was specified,
-            the response should include the fields of the user's profile that can be made
-            public, such as the display name and avatar.
+            The profile for the user. If a `field` is specified in the request, the response
+            MUST only included the specified `field`. If no `field` was specified, the response
```
richvdh:
```suggestion
            MUST only include the specified `field`. If no `field` was specified, the response
```

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,53 @@ paths:
...
+            SHOULD include all of the fields of the user's profile, which can be made public.
```
richvdh: This doesn't mean what you want it to mean. In any case, why have you removed the examples of display name and avatar?

```suggestion
            SHOULD include all of the fields of the user's profile that can be made
            public, such as the display name and avatar.
```

richvdh (changes requested):

data/api/server-server/query.yaml
clokep: I find it more confusing. It sounds like modifications can be made via this endpoint which is not true. The endpoint is read only.

clokep: Instead of "Allow user to set..." maybe say "Include arbitrary keys..." or something like that.

data/api/server-server/query.yaml
```diff
@@ -111,47 +111,53 @@ paths:
...
-        for a given user. Homeservers should only query profiles for users that belong
+        for a given user. Homeservers MUST only query profiles for users that belong
         to the target server (identified by the [server name](/appendices/#server-name)
         in the user ID).
 
-        Servers may wish to cache the response to this query to avoid requesting the
-        information too often.
+        Responding servers MAY:
+          - Allow users to set arbitrary key/value pairs in their profile in addition to the
+            specified pairs.
```
clokep:
```suggestion
          - Include arbitrary key/value pairs in the profile in addition to the
            specified pairs.
```

data/api/server-server/query.yaml
famfo: Well, kind of? Before MSC 4133 user couldn't set other properties themselves so if a server were to do validation that would be a change from the previous behavior, but this is good enough for me.

data/api/server-server/query.yaml
famfo: Huh, why doesn't that mean what I want it to?

Anyway, I recently encountered a case where this was explicitly not the case but yea, it might be slightly pedantic to remove the note.

data/api/server-server/query.yaml
clokep: Resolving this since my other suggestion was included.

data/api/server-server/query.yaml
richvdh: > Huh, why doesn't that mean what I want it to?

Because of the comma. "The response SHOULD include all of the fields of the user's profile, which can be made public" means "The response SHOULD include all of the fields of the user's profile. Also, the profile can be made public".

Obviously, it's not the case that the whole of the user's profile can be made public, but that's what's implied by your sentence.

Are you a native German speaker by any chance? I think this is a difference between German and English. (And, for the record, my German is far far worse than your English, so please don't take this as a criticism!)

data/api/server-server/query.yaml
richvdh: I think this has been addressed; resolving.

data/api/server-server/query.yaml
richvdh: Ok this is the only outstanding thread on this PR. If you're happy with my suggestion, please commit it and we can merge.

richvdh (approved):
Thanks!

PR #2331, opened 2026-03-06
Add page search widget (Pagefind)

Johennes:
This is a proof of concept for adding a page search feature to the spec. It uses https://pagefind.app/ via [Docsy's built-in search UI](https://www.docsy.dev/docs/content/search/#custom-search). This adds a search field in the top right corner of the navbar. Results display in a dropdown below the search field.

The result display logic is custom. Results are grouped under the main page's title with only the first three subresults revealed by default. Further subresults from the same page are hidden inside an expandable container.

Screenshot 2026-03-19 at 08 53 14

<del>⚠️ The PR currently disables CSP _entirely_ for the purpose of demonstration. The CSP rules will have to be tweaked to allow loading the Pagefind Wasm.</del>

--- 

The initial version of this PR tried to mimic how the search works on https://element-hq.github.io/synapse/latest/.  There was a new "Search" button in the navbar.

Screenshot 2026-03-06 at 15 19 19

Clicking it, exposed and focused a search widget at the top of the current page.

Screenshot 2026-03-06 at 15 19 57

Clicking the button again, hid the search widget. Search results were displayed inline below the search box.

Screenshot 2026-03-06 at 15 22 06

### Pull Request Checklist

Preview: https://pr2331--matrix-spec-previews.netlify.app

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -236,6 +236,10 @@ jobs:
         run: |
           tar -C "spec${baseURL}" --strip-components=1 -xzf openapi.tar.gz
 
+      - name: "🔍 pagefind indexing"
+        run: |
+          npx -y pagefind --site "spec${baseURL}"
+
       - name: "📦 Tarball creation"
         run: |
           cd spec
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -238,7 +238,7 @@ jobs:
 
       - name: "🔍 pagefind indexing"
         run: |
-          npx -y pagefind --site "spec${baseURL}"
+          npm run pagefind -- --site "spec${baseURL}"
 
       - name: "📦 Tarball creation"
         run: |
--- a/README.md
+++ b/README.md
@@ -79,6 +79,18 @@ We use a highly customized [Docsy](https://www.docsy.dev/) theme for our generat
 Awesome. If you're looking at making design-related changes to the spec site, please coordinate with us in
 [#matrix-docs:matrix.org](https://matrix.to/#/#matrix-docs:matrix.org) before opening a PR.
 
+## Page search
+
+The spec uses [Pagefind](https://pagefind.app/) to provide a page search widget. To test this locally, you'll need to generate the
+search index _after_ building the static site.
+
+```
+hugo build && npx -y pagefind --site public && hugo serve
+```
+
+Note that while `hugo serve` supports hot reloading, changes made to the site content won't reflect in the search index without
+rebuilding it.
+
 ## Building the specification
 
 If for some reason you're not a CI/CD system and want to render a static version of the spec for yourself, follow the above
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The spec uses [Pagefind](https://pagefind.app/) to provide a page search widget.
 search index _after_ building the static site.
 
 ```
-hugo build && npx -y pagefind --site public && hugo serve
+hugo build && npm run pagefind -- --site public && hugo serve
 ```
 
 Note that while `hugo serve` supports hot reloading, changes made to the site content won't reflect in the search index without
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -179,7 +179,7 @@ search backend.
 
               $searchResultBody.append($("<p>").append($expander));
               $wrapper = $("<div>")
-                .addClass("collapse")
+                .addClass("collapse td-offline-search-results__subresults")
                 .attr("id", wrapper_id)
                 .on("hide.bs.collapse", _ => $action.text("▶ Show"))
                 .on("show.bs.collapse", _ => $action.text("▼ Hide"));
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -15,18 +15,18 @@ limitations under the License.
 */
 
 /*
-Adapted from [1] to combine Docsy's built-in search UI with the Pagefind
+Adapted from [1] to combine Docsy"s built-in search UI with the Pagefind
 search backend.
 
   [1]: https://github.com/matrix-org/docsy/blob/71d103ebb20ace3d528178c4b6d92b6cc4f7fd53/assets/js/offline-search.js
 */
 
 (function ($) {
-  'use strict';
+  "use strict";
 
   $(document).ready(async function () {
     const pagefind = await import("/pagefind/pagefind.js");
-    const $searchInput = $('.td-search input');
+    const $searchInput = $(".td-search input");
 
     //
     // Lazily initialise Pagefind only when the user is about to start a search.
@@ -40,7 +40,7 @@ search backend.
     // Register handler
     //
 
-    $searchInput.on('change', (event) => {
+    $searchInput.on("change", (event) => {
       render($(event.target));
 
       // Hide keyboard on mobile browser
@@ -48,7 +48,7 @@ search backend.
     });
 
     // Prevent reloading page by enter key on sidebar search.
-    $searchInput.closest('form').on('submit', () => {
+    $searchInput.closest("form").on("submit", () => {
       return false;
     });
 
@@ -73,7 +73,7 @@ search backend.
       //
 
       const searchQuery = $targetSearchInput.val();
-      if (searchQuery === '') {
+      if (searchQuery === "") {
         return;
       }
 
@@ -88,75 +88,109 @@ search backend.
       // Make result html
       //
 
-      const $html = $('<div>');
+      const $html = $("<div>");
 
       $html.append(
-        $('<div>')
+        $("<div>")
           .css({
-            display: 'flex',
-            justifyContent: 'space-between',
-            marginBottom: '1em',
+            display: "flex",
+            justifyContent: "space-between",
+            marginBottom: "1em",
           })
           .append(
-            $('<span>').text('Search results').css({ fontWeight: 'bold' })
+            $("<span>").text("Search results").css({ fontWeight: "bold" })
           )
           .append(
-            $('<span>').addClass('td-offline-search-results__close-button')
+            $("<span>").addClass("td-offline-search-results__close-button")
           )
       );
 
-      const $searchResultBody = $('<div>').css({
+      const $searchResultBody = $("<div>").css({
         maxHeight: `calc(100vh - ${
           $targetSearchInput.offset().top - $(window).scrollTop() + 180
         }px)`,
-        overflowY: 'auto',
+        overflowY: "auto",
       });
       $html.append($searchResultBody);
 
       if (results.length === 0) {
         $searchResultBody.append(
-          $('<p>').text(`No results found for query "${searchQuery}"`)
+          $("<p>").text(`No results found for query "${searchQuery}"`)
         );
       } else {
-        results.forEach((r) => {
-          r.sub_results.forEach((s) => {
-            const href = s.url;
+        results.forEach((r, index_r) => {
+          // Add the main result"s page title.
+          $searchResultBody.append(
+            $("<a>")
+              .addClass("d-block")
+              .css({
+                fontSize: "1.2rem",
+              })
+              .attr("href", r.url)
+              .text(r.meta.title)
+          );
+          
+          // Render the first 3 subresults per page and wrap the rest
+          // in a collapsed container.
+          const LIMIT = 3;
+          let $wrapper = null;
 
-            const $entry = $('<div>').addClass('mt-4');
+          r.sub_results.forEach((s, index_s) => {
+            
+
+            if (index_s === LIMIT) {
+              const wrapper_id = `collapssible-subresults-${index_r}`;
+              const $expander = $("<a>")
+                .attr("data-bs-toggle", "collapse")
+                .attr("data-bs-target", `#${wrapper_id}`)
+                .attr("href", "#")
+                .attr("role", "button")
+                .attr("aria-expanded", "false")
+                .attr("aria-controls", wrapper_id)
+                .css("margin-left", "0.5rem")
+                .text(`${r.sub_results.length - index_s} more result(s) from ${r.meta.title}`);
+
+              $searchResultBody.append($("<p>").append($expander));
+              $wrapper = $("<div>")
+                .addClass("collapse")
+                .attr("id", wrapper_id);
+              $searchResultBody.append($wrapper);
+            }
+
+            const $entry = $("<div>")
+              .css("margin-top", "0.5rem")
+              .css("margin-left", "0.5rem");
 
             $entry.append(
-              $('<small>').addClass('d-block text-body-secondary').text(r.meta.title)
-            );
-
-            $entry.append(
-              $('<a>')
-                .addClass('d-block')
-                .css({
-                  fontSize: '1.2rem',
-                })
-                .attr('href', href)
+              $("<a>")
+                .addClass("d-block")
+                .attr("href", s.url)
                 .text(s.title)
             );
 
-            $entry.append($('<p>').html(s.excerpt));
+            $entry.append($("<p>").html(s.excerpt));
 
-            $searchResultBody.append($entry);
+            if (index_s < LIMIT) {
+              $searchResultBody.append($entry);
+            } else {
+              $wrapper.append($entry);
+            }
           });
         });
       }
 
-      $targetSearchInput.one('shown.bs.popover', () => {
-        $('.td-offline-search-results__close-button').on('click', () => {
-          $targetSearchInput.val('');
-          $targetSearchInput.trigger('change');
+      $targetSearchInput.one("shown.bs.popover", () => {
+        $(".td-offline-search-results__close-button").on("click", () => {
+          $targetSearchInput.val("");
+          $targetSearchInput.trigger("change");
         });
       });
 
       const popover = new bootstrap.Popover($targetSearchInput, {
         content: $html[0],
         html: true,
-        customClass: 'td-offline-search-results',
-        placement: 'bottom',
+        customClass: "td-offline-search-results",
+        placement: "bottom",
       });
       popover.show();
     };
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -34,7 +34,7 @@ search backend.
 
     $searchInput.focus(() => {
       pagefind.init();
-    }); 
+    });
 
     //
     // Register handler
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -87,7 +87,7 @@ search backend.
             .text("Loading...")))
         .append($("<p>")
           .text("Loading..."));
-      const popover = new bootstrap.Popover($targetSearchInput, {
+      const popover = new bootstrap.Popover($targetSearchInput[0], {
         content: $spinner[0],
         html: true,
         customClass: "td-offline-search-results",
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -226,5 +226,5 @@ search backend.
 //
 
 const resultsString = (n) => {
-  return n > 1 ? "results" : "result";
+  return n === 1 ? "result" : "results";
 };
--- a/assets/js/offline-search.js
+++ b/assets/js/offline-search.js
@@ -165,7 +165,7 @@ search backend.
           r.sub_results.forEach((s, index_s) => {
             if (index_s === LIMIT) {
               const wrapper_id = `collapssible-subresults-${index_r}`;
-              const $action = $("<span>").text("Show");
+              const $action = $("<span>").text("▶ Show");
               const $expander = $("<a>")
                 .attr("data-bs-toggle", "collapse")
                 .attr("data-bs-target", `#${wrapper_id}`)
@@ -173,7 +173,6 @@ search backend.
                 .attr("role", "button")
                 .attr("aria-expanded", "false")
                 .attr("aria-controls", wrapper_id)
-                .css("margin-left", "0.5rem")
                 .append($action)
                 .append($("<span>").text(` ${r.sub_results.length - index_s} more result(s) from ${r.meta.title}`));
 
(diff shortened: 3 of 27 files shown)


assets/js/offline-search.js
anoadragon453, line 1: Is it possible to change the searchbox text from "Search this site..." to "Search the spec..."?

assets/js/offline-search.js
```diff
@@ -0,0 +1,198 @@
...
+      render($(event.target));
+
+      // Hide keyboard on mobile browser
+      $searchInput.blur();
+    });
+
+    // Prevent reloading page by enter key on sidebar search.
+    $searchInput.closest("form").on("submit", () => {
+      return false;
+    });
```
anoadragon453, line 57: Is it possible to show any sort of "loading" indicator after pressing return? It wasn't immediately clear to me that pressing enter kicked off the search until the search results appeared, about 1s later.

assets/js/offline-search.js
```diff
@@ -0,0 +1,198 @@
...
+              const wrapper_id = `collapssible-subresults-${index_r}`;
+              const $expander = $("<a>")
+                .attr("data-bs-toggle", "collapse")
+                .attr("data-bs-target", `#${wrapper_id}`)
+                .attr("href", "#")
+                .attr("role", "button")
+                .attr("aria-expanded", "false")
+                .attr("aria-controls", wrapper_id)
+                .css("margin-left", "0.5rem")
+                .text(`${r.sub_results.length - index_s} more result(s) from ${r.meta.title}`);
```
anoadragon453: Can we toggle this text to "Hide x more result(s) ..." after it's been expanded?

assets/scss/_styles_project.scss
anoadragon453, line 1: Hovering over the close button doesn't change the mouse cursor to a "click" icon, whereas I would expect it to:

Image

assets/js/offline-search.js
anoadragon453, line 57: Is it possible to keep the search field highlighted after pressing return? I'd like to continue to refine my search term without needing to use the mouse to click the field again.

anoadragon453 (commented):
A few suggestions for UX. Otherwise the search results seem good!

assets/js/offline-search.js
Johennes: Yes, have changed it in 6f05c2c78e5f879121e913c17d722901ae0b9180.

assets/js/offline-search.js
Johennes: Yes, excellent call. Have added a spinner in 92b7e714e7992e7bb2340add5792b74232cbeefa.

assets/js/offline-search.js
Johennes: Good idea. Have implemented this with 8e383835b96ac004ad6ff39e3074625ac8506b26. There's still some weird vertical space that jumps around during expansion. I haven't been able to figure out why so far.

assets/scss/_styles_project.scss
Johennes: Fixed with 72205be3dcedb2a50e864ce88ccf2efc7168fc6c.

assets/js/offline-search.js
Johennes: It looks like the Docsy template did this with mobile in mind.

```js
// Hide keyboard on mobile browser
$searchInput.blur();
```

I have removed that with a56969149f31875e147cccfed39bbe0928b5e626 since mobile probably isn't that big of a use case for us.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+        );
+      } else {
+        results.forEach((r, index_r) => {
+          // Add the main result's page title.
+          $searchResultBody.append($("<div>")
+            .append($("<a>")
+              .css({
+                fontSize: "1.2rem",
+              })
+              .attr("href", r.url)
```
Johennes: I have no idea if these URLs will work in production where the spec is deployed under a subpath rather than on `/`. I don't know how to test this before releasing production either, sadly. 😕

assets/js/offline-search.js
anoadragon453: Since this is one of those small, annoying things, I gave the problem to an LLM. Apologies if it ends up being slop. The commit is https://github.com/matrix-org/matrix-spec/pull/2331/commits/2fd3c28a7a64b1b5d276cff3a5a47d642f030d2b, with its guess being:

> The jump is coming from margin-collapsing on the expanded Bootstrap collapse wrapper. During the height animation, Bootstrap applies `overflow: hidden`, which prevents the first hidden result’s top margin from collapsing. When the animation finishes, that constraint disappears and the content snaps upward by about 0.5rem (roughly 8px), which matches what you were seeing.
> 
> I fixed it by giving the hidden-results wrapper a dedicated class in `offline-search.js:180` and making the shown state establish its own formatting context in `_styles_project.scss:671`. The key rule is display: flow-root on `.td-offline-search-results__subresults.collapse.show`, which prevents that margin collapse without changing the collapse behaviour.

I failed to try it out locally due to being unable to load `http://localhost:1313/pagefind/pagefind.js` (via `hugo serve`). Even after indexing the site via pagefind:

<details>

<summary>Click to show pagefind indexing steps</summary>

```
➜ hugo -d spec  
Start building sites … 
hugo v0.152.2+extended+withdeploy linux/amd64 BuildDate=unknown VendorInfo=nixpkgs


                  │ EN  
──────────────────┼─────
 Pages            │ 187 
 Paginator pages  │   0 
 Non-page files   │   0 
 Static files     │  39 
 Processed images │   6 
 Aliases          │   0 
 Cleaned          │   0 

Total in 4644 ms

matrix-spec on  johannes/page-search [$!?] took 4.7s …
➜ npx -y pagefind --site "spec"          

Running Pagefind v1.4.0 (Extended)
Running from: "/home/work/code/matrix-spec"
Source:       "spec"
Output:       "spec/pagefind"

[Walking source directory]
Found 44 files matching **/*.{html}

[Parsing files]
Did not find a data-pagefind-body element on the site.
↳ Indexing all <body> elements on the site.

[Reading languages]
Discovered 1 language: en

[Building search indexes]
Total: 
  Indexed 1 language
  Indexed 44 pages
  Indexed 11939 words
  Indexed 0 filters
  Indexed 0 sorts

Finished in 1.764 seconds
```

</details>

But perhaps I'm missing something?

Edit: looks like it worked in the preview though :tada:

Johennes: Well amazing! Thank you. ❤️ 

For local running, you'll need `hugo build && npx -y pagefind --site public --serve`. Maybe this should go into the README somewhere.

anoadragon453: Aha, that works perfectly.

> Maybe this should go into the README somewhere.

Yes please! As a suggestion if you specifically want to test out the search functionality (as I assume `pagefind` is otherwise worse as `hugo` as it doesn't support hot-reloading of content).

Johennes: Have added it with de25ba52651d2210d98d9e8413c95ea8aeb70372 in the variant `hugo build && npx -y pagefind --site public && hugo serve`. That still supports hugo hot reloading just that any content changes will be missing in the search index (which I think is probably ok for testing).

anoadragon453: Oh, nice! Yep, that works perfectly here.

assets/js/offline-search.js
anoadragon453: I tried this locally:

```
hugo --baseURL '/v1.17/' -d 'spec/v1.17/'
pagefind --site spec/v1.17/
```

and the links in the generated indexes look fine:

```
http://127.0.0.1:8123/v1.17/server-server-api/#get_matrixfederationv1version
```

So I think we should be fine here.

README.md
```diff
@@ -79,6 +79,18 @@ We use a highly customized [Docsy](https://www.docsy.dev/) theme for our generat
...
 [#matrix-docs:matrix.org](https://matrix.to/#/#matrix-docs:matrix.org) before opening a PR.
 
+## Page search
+
+The spec uses [Pagefind](https://pagefind.app/) to provide a page search widget. To test this locally, you'll need to generate the
+search index _after_ building the static site.
+
+```
+hugo build && npx -y pagefind --site public && hugo serve
+```
```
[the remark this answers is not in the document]
anoadragon453: `hugo build` exists, and the provided command does indeed work, Copilot :)

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+
+      // Kick off the search.
+      const search = await pagefind.debouncedSearch(searchQuery);
+      if (search === null) {
+        // A more recent search call has been made, nothing to do.
+        return;
+      }
+
+      // Load all result details. We may want to switch to a paged UI in future for better performance.
+      const results = await Promise.all(search.results.slice(0, 100).map(r => r.data()));
```
anoadragon453: I noticed some odd behaviour when testing the site locally in Firefox and Chromium.

* In Firefox, the cursor disappears for a few seconds while the results are loaded.
* In Chromium, the browser locks up for about 15s before results are loaded.

In both cases, I think loading 100 promises in parallel is the problem here. It looks like Docsy [is using](https://github.com/matrix-org/docsy/blob/71d103ebb20ace3d528178c4b6d92b6cc4f7fd53/assets/js/offline-search.js#L103) `offline-search-max-results`, which [is defined here](https://github.com/matrix-org/docsy/blob/71d103ebb20ace3d528178c4b6d92b6cc4f7fd53/layouts/_partials/search-input.html#L52), and is linked to the configurable `offlineSearchMaxResults` site setting.

So:

1. I think we should use that existing setting instead of hardcoding 100.
2. We should reduce this to something smaller to prevent the browser thread from locking up.

If we really want >10 results, we can try batching `r.data()` calls instead of doing one large `Promise.all()`.

anoadragon453 (changes requested):

config/_default/hugo.toml
```diff
@@ -151,7 +152,9 @@ sidebar_menu_compact = true
   [server.headers.values]
     # `style-src 'unsafe-inline'` is needed to correctly render the maths in the Olm spec:
     # https://github.com/KaTeX/KaTeX/issues/4096
-    Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self'; img-src 'self' data:; connect-src 'self'; font-src 'self' data:; media-src 'self'; child-src 'self'; form-action 'self'; object-src 'self'"
+    # `script-src 'unsafe-eval'` is needed because Pagefind relies on it to load its Wasm:
+    # https://github.com/Pagefind/pagefind/blob/main/docs/content/docs/hosting.md
+    Content-Security-Policy = "default-src 'self'; style-src 'self' 'unsafe-inline'; script-src 'self' 'unsafe-eval'; img-src 'self' data:; connect-src 'self'; font-src 'self' data:; media-src 'self'; child-src 'self'; form-action 'self'; object-src 'self'"
```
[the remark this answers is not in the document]
Johennes: As per https://github.com/Pagefind/pagefind/blob/main/docs/content/docs/hosting.md, `wasm-unsafe-eval` doesn't yet work in Safari.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+    };
+  });
+})(jQuery);
+
+//
+// Helpers
+//
+
+const resultsString = (n) => {
+  return n > 1 ? "results" : "result";
```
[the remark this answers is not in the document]
Johennes: As per the calling code, this should never happen but still fair.

.github/workflows/main.yml
```diff
@@ -236,6 +236,10 @@ jobs:
         run: |
           tar -C "spec${baseURL}" --strip-components=1 -xzf openapi.tar.gz
 
+      - name: "🔍 pagefind indexing"
+        run: |
+          npx -y pagefind --site "spec${baseURL}"
```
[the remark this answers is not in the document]
Johennes: I pinned the version to 1.4.0 in 308bcfd20eb7ccfca30f89080e802cd36698b76b.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+search backend.
+
+  [1]: https://github.com/matrix-org/docsy/blob/71d103ebb20ace3d528178c4b6d92b6cc4f7fd53/assets/js/offline-search.js
+*/
+
+(function ($) {
+  "use strict";
+
+  $(document).ready(async function () {
+    const pagefind = await import("/pagefind/pagefind.js");
```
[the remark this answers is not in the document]
Johennes: Actually `./` doesn't work because this ends up in `${deployment}/js/main.js`. We need to navigate one level up first to be able to use a relative path to `${deployment}/pagefind/pagefind.js`. Have re-fixed this with bccbff878d6a8c8585bde19bc0bbc6546d4da404.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+            .text("Loading...")))
+        .append($("<p>")
+          .text("Loading..."));
+      const popover = new bootstrap.Popover($targetSearchInput, {
+        content: $spinner[0],
+        html: true,
+        customClass: "td-offline-search-results",
+        placement: "bottom",
+      });
+      popover.show();
```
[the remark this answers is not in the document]
Johennes: For some reason this worked all the same but I have fixed it with 0cf8fca4dae1c28455768ea7e0de52f1f6142eaa.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+          r.sub_results.forEach((s, index_s) => {
+            if (index_s === LIMIT) {
+              const num_hidden_results = r.sub_results.length - index_s;
+              const wrapper_id = `collapssible-subresults-${index_r}`;
+              const $action = $("<span>").text("▶ Show");
+              const $expander = $("<a>")
+                .attr("data-bs-toggle", "collapse")
+                .attr("data-bs-target", `#${wrapper_id}`)
+                .attr("href", "#")
+                .attr("role", "button")
```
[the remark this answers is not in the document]
Johennes: Have switched to buttons with 84b5e98f255cea9028aa982e201debf21abb11cb. Though this also changed the UI to some extent.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+          .css({ fontWeight: "bold" }))
+        .append($("<span>")
+          .addClass("td-offline-search-results__close-button")
+          .attr("role", "button")
+          .attr("aria-label", "Close")
+          .on("click", () => {
+            $targetSearchInput.val("");
+            $targetSearchInput.trigger("change");
+          })
+        )
```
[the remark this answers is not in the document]
Johennes: Have switched to a button with af1e9382cada771a362a371334c8488c73fba75f.

config/_default/hugo.toml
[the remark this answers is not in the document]
anoadragon453: Could you leave a comment talking about Safari support explicitly in the meantime? So we know to reduce the scope once that's out.

assets/js/offline-search.js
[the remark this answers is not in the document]
Johennes: Have improved the styling somewhat with 0cc8ee38d5875836b6170278f43168cfd14abb2f.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+
+      // Finally, show the search results.
+      //
+      // Ideally we would just call setContent but there appears to be a bug in Bootstrap
+      // that causes the popover to be hidden when setContent is called after the popover
+      // has been shown. To work around this, we use the hack from [1] to inject the HTML
+      // content manually.
+      //
+      // [1]: https://github.com/twbs/bootstrap/issues/37206#issuecomment-1259541205
+      $(popover.tip.querySelector('.popover-body')).html($html[0]);
```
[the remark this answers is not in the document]
Johennes: Evidently this worked fine but I suspect the other variant may be cleaner. Have applied it with 49fec082f5ead13d4126189447f44df26b99c618.

assets/js/offline-search.js
```diff
@@ -0,0 +1,230 @@
...
+
+    $searchInput.focus(() => {
+      pagefind.init();
+    });
+
+    //
+    // Set up search input handler.
+    //
+
+    $searchInput.on("change", (event) => {
```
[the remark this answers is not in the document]
Johennes: Have changed it to start searching upon Enter explicitly in 86675ad4a4d9e4df1b6a0214ca21ad6ac9699ff7.

config/_default/hugo.toml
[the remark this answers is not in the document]
Johennes: Good point. Have added a remark in 754e9c82b89e9e3a9e1d546383b4514a592a389b.

assets/js/offline-search.js
Johennes: Hm, good catch. I think what makes a static limit hard is the main result / subresults structure. The first main result may already have 100 subresults which would then make us omit any further main results or subresults.

Maybe we should load the first 3 subresults from each main result and display that (which should be safe because we don't have that many pages at the moment). Then once expanding further subresults, we could load these – possibly in a paged manner?

In other words, initially you'd be presented with:

> ### Client-Server API (80 results)
>
> - Some result 1
> - Some result 2
> - Some result 3
>
> <a href="#">Show more results from Client-Server API (77 remaining)</a>
>
> ### Server-Server API (13 results)
>
> ...

Then when you click "Show more...", you'd get:

> ### Client-Server API (80 results)
>
> - Some result 1
> - Some result 2
> - Some result 3

> - Some result 4
> - Some result 5
> - Some result 6
> - Some result 7
> - Some result 8
> - Some result 9
> - Some result 10
> - Some result 11
> - Some result 12
> - Some result 13
>
> <a href="#">Show more results from Client-Server API (67 remaining)</a>
>
> ### Server-Server API (13 results)
>
> ...

anoadragon453: It makes a lot of sense to gate loading additional results behind clicking the expansion button. Is that simple to implement (selectively loading the first three results, and then the next 10, from each page)?

If so, we can probably stick to our hardcoded counts (though we should comment that we're purposefully ignoring the option).

Johennes: Oh, actually the API is a little less flexible than I expected. It looks like loading a main result includes loading all subresults. So we cannot selectively load the subresults in pages, sadly. I guess the best we can do is load the main results in pages then.

anoadragon453: The slow part is calling `Promise.all()` on all of the results simultaneously. We can still do that in batches of, say 10, and then append that set of results to the existing list iteratively?

Johennes: I have switched it around to load and render main results one after the other in 8fb380d465fcc1dd51ff41bdf51db4b604e75234. This helps a little but when a main result contains many subresults, it still feels quite slow. I think the only way around that is to load further main results on demand only.

Johennes: e3ca1ba2b8faae4143ea9764f7f49178faf9bdfc now loads main results progressively in pages of three on button click. This still doesn't help a lot when a main result has hundreds of results. I think we won't get around this without a change of Pagefind's API to allow loading individual subresults. Overall this feels better than the initial "load everything" version though.

anoadragon453: Agreed. My browsers no longer freezes up when searching the fairly generic "endpoint".

It does still freeze for a bit while searching "the", but as you say, I don't think we can do much about that at the moment. And that should be an edge case.

assets/js/offline-search.js
```diff
@@ -0,0 +1,290 @@
...
+
+    $searchInput.focus(() => {
+      pagefind.init();
+    });
+
+    //
+    // Set up search input handler.
+    //
+
+    $searchInput.on("keypress", (event) => {
```
anoadragon453: `keypress` [has been deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event). Use `keydown` instead.

```js
$searchInput.on("keydown", (event) => {
  if (event.key === "Enter") {
    event.preventDefault();
    render($(event.target));
    return;
  }
});
```

assets/js/offline-search.js
```diff
@@ -0,0 +1,290 @@
...
+      .addClass("spinner-border")
+      .attr("role", "status")
+      .append($("<div>")
+        .addClass("visually-hidden")
+        .text("Loading...")))
+    .append($("<p>")
+      .text("Loading..."));
+}
+
+const loadAndRenderResults = async (results, offset, $spinner, $searchResultBody) => {
```
anoadragon453, line 184: Spawning from [a discussion](https://matrix.to/#/!DWdtrsMe_KmWrK7MFFUsbhPnxfR3jNPmW0vFXkiq-GU/$rwi3w0WJg1W8ExGvtbwpvnhuu3im1mClBNwA9Zt5VkU?via=element.io&via=matrix.org&via=mtrnord.blog) in the Matrix Spec & Docs Authoring room:

It would be nice to rank some pages above others. For instance, I found on a search for "authorisation" that Changelog entries were appearing above spec pages in the results.

Image

We could shove changelog entries at the end with something like the following:

```
if result.url.includes("/changelog/") {...} 
```

anoadragon453 (changes requested):

assets/js/offline-search.js
Johennes: Good catch. Fixed in 5fa9103ae571035cac39f493565bc1a9551027a0.

assets/js/offline-search.js
Johennes: This is now done in 78ab248a0b652e3da5cbe317bfc16f897f11b6df.

I chose a slightly different approach because the URL is only available after loading a result – which we previously found out to be slow. Instead, I added a filter to split the search into two steps. First all non-changelog pages are searched, then the changelogs are searched afterwards and the results concatenated.

Johennes:
Sorry for the delay in getting this updated. The latest comments should now be addressed.

anoadragon453 (approved):
This now looks good to me! Thanks for getting it over the line.

I played around with it a bit and it works very well now! Still freezes the page *momentarily* when searching for a very generic term, but I think this is fine for the vast majority of searches.

Thank you!

PR #2332, opened 2026-03-09
Specify Policy Servers (MSC4284)

turt2live:
MSC (rendered): https://github.com/matrix-org/matrix-spec-proposals/blob/main/proposals/4284-policy-servers.md
MSC (PR): https://github.com/matrix-org/matrix-spec-proposals/pull/4284

**Note for reviewer:** I've arbitrarily decided that "Policy Server" is a noun. Feel free to decide otherwise, and I'll fix it. 

### Pull Request Checklist

Preview: https://pr2332--matrix-spec-previews.netlify.app

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -473,6 +473,9 @@ server must ensure that the event:
     otherwise it is rejected.
 6.  Passes authorization rules based on the current state of the room,
     otherwise it is "soft failed".
+7. {{% added-in v="1.18" %}} Is [validated](#validating-policy-server-signatures)
+   by the Policy Server, if the room is [using a Policy Server](#determining-if-a-policy-server-is-enabled-in-a-room),
+   otherwise it is "soft failed".
 
 Further details of these checks, and how to handle failures, are
 described below.
@@ -688,6 +691,11 @@ then any new event `D'` will not reference `C`:
     B   C
     |
     D'
+
+{{% boxes/note %}}
+{{% added-in v="1.18" %}}
+Events can also be soft failed if they fail [Policy Server checks](#validating-policy-server-signatures).
+{{% /boxes/note %}}
 
 #### Retrieving event authorization information
 
@@ -1465,6 +1473,12 @@ The signature is then copied back to the original event object.
 For an example of a signed event, see the [room version
 specification](/rooms).
 
+{{% boxes/note %}}
+{{% added-in v="1.18" %}}
+Events sent in rooms with [Policy Servers](#policy-servers) MUST [ask](#asking-for-a-policy-server-signature-on-an-event)
+the Policy Server for a signature too.
+{{% /boxes/note %}}
+
 ### Validating hashes and signatures on received events
 
 When a server receives an event over federation from another server, the
@@ -1498,6 +1512,12 @@ If the hash check fails, then it is assumed that this is because we have
 only been given a redacted version of the event. To enforce this, the
 receiving server should use the redacted copy it calculated rather than
 the full copy it received.
+
+{{% boxes/note %}}
+{{% added-in v="1.18" %}}
+Events sent in rooms with [Policy Servers](#policy-servers) have [additional](#validating-policy-server-signatures)
+signature validation requirements.
+{{% /boxes/note %}}
 
 ### Calculating the reference hash for an event
 


data/api/client-server/policy_server.yaml
```diff
@@ -0,0 +1,77 @@
...
+paths:
+  /matrix/policy_server:
+    get:
+      summary: Gets information about a Policy Server's configuration, if available.
+      description: |-
+        Gets public key information for a [Policy Server](/client-server-api/#policy-servers).
+
+        {{% boxes/note %}}
+        Like the [well-known discovery URI](/client-server-api/#well-known-uris),
+        this endpoint should be accessed with the hostname of the homeserver's
```
zecakeh: Should it be:


```suggestion
        this endpoint should be accessed with the hostname of the Policy Server's
```
?

data/api/client-server/policy_server.yaml
```diff
@@ -0,0 +1,77 @@
...
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  public_keys:
+                    type: object
+                    description: |-
+                      The unpadded base64-encoded public keys for the Policy Server. MUST contain
+                      at least `ed25519`.
```
zecakeh, line 50: If the `ed25519` key is required, why is it not actually defined in the schema?

data/event-schemas/schema/m.room.policy.yaml
```diff
@@ -0,0 +1,36 @@
...
+    properties:
+      via:
+        description: |-
+          The server name to use as a Policy Server. MUST have a joined user in
+          the room.
+        type: string
+      public_keys:
+        description: |-
+          The unpadded base64-encoded public keys for the Policy Server. MUST contain at
+          least `ed25519`.
```
zecakeh, line 20: Same here.

data/api/client-server/policy_server.yaml
turt2live: either works in this case. "Homeserver" felt less confusing for client developers, but that's not a strong reason to keep it as-is.

data/api/client-server/policy_server.yaml
turt2live: because it yelled at me when I tried. I probably did it wrong though.

data/event-schemas/schema/m.room.policy.yaml
turt2live: ref https://github.com/matrix-org/matrix-spec/pull/2332#discussion_r2910257429

content/client-server-api/modules/policy_servers.md
```diff
@@ -0,0 +1,98 @@
...
+Room operators might prefer to use a "moderation bot" instead of a Policy Server.
+Moderation bots are reactive because they typically [redact](#redactions) events
+after they've already been sent.
+
+Room operators might also prefer to layer a moderation bot with a Policy Server
+for added protection.
+{{% /boxes/note %}}
+
+A room's Policy Server is designated by the [`m.room.policy`](#mroompolicy) state
+event described below. If the state event is not set in the room or is incomplete,
```
anoadragon453: What do you mean by an "incomplete" state event? A non-spec compliant one?

I might just say "If the state event ... or does not contain the required fields, then the room does *not* use a Policy Server".

content/server-server-api.md
```diff
@@ -1277,6 +1277,171 @@ endpoint MUST be protected as follows:
         room ID MUST be ignored if the sending server is denied access to
         the room identified by that ID.
 
+{{< added-in v="1.18" >}} The following endpoints MAY be protected:
+
+-   [`/_matrix/policy/v1/sign`](#post_matrixpolicyv1sign) - Protected if the server
+    is tracking the DAG and chooses to enforce the ACL.
```
anoadragon453: This might make more sense as:

```suggestion
The following endpoints MAY be protected:

-   [`/_matrix/policy/v1/sign`](#post_matrixpolicyv1sign) - {{< added-in v="1.18" >}}
    Protected if the server is tracking the DAG and chooses to enforce the ACL.
```

content/server-server-api.md
```diff
@@ -1277,6 +1277,171 @@ endpoint MUST be protected as follows:
...
+
+{{% boxes/warning %}}
+The Policy Server's [published signing key](#publishing-keys) SHOULD NOT be the
+same key contained in the `m.room.policy` state event. This is to keep "can send
+federation traffic" and "can recommend events for inclusion" separate, and to
+allow rooms to revoke the Policy Server's key without cooperation of the Policy
+Server.
+
+If the Policy Server is acting as a normal homeserver and attempting to send an
+event, that event will require the a signature from the server's published signing
```
anoadragon453:
```suggestion
event, that event will require a signature from the server's published signing
```

content/server-server-api.md
anoadragon453, line 1: Should checking the signature of an event from a policy server be included as one of the steps under ["Checks performed on receipt of a PDU"](https://spec.matrix.org/v1.17/server-server-api/#checks-performed-on-receipt-of-a-pdu)?

anoadragon453, line 1: Should section ["5.1.4 Soft failure"](https://spec.matrix.org/v1.17/server-server-api/#soft-failure) contain a quick reference to events that fail Policy Server checks?

anoadragon453, line 1: [Signing Events -> Adding hashes and signatures to outgoing events](https://spec.matrix.org/v1.17/server-server-api/#adding-hashes-and-signatures-to-outgoing-events) should note something along the lines of "when a valid policy server is configured, the event must also carry the policy server’s signature before it is sent."

Likewise with the "Validating hashes and signatures on received events" side, it should be expanded to cover validating the signature against an active Policy Server's public key.

turt2live: note that the requirement is on *asking* the policy server for the signature, not actually having the signature. If we required the signature, it'd break all existing room versions.

PR #2333, opened 2026-03-10
Add page search widget (Docsy / Lunr)

Johennes:
This is an alternative to https://github.com/matrix-org/matrix-spec/pull/2331 that uses [Docsy's built-in page search support](https://www.docsy.dev/docs/content/search/) based on Lunr.

The search input will display on the right side of the navbar. You'll need to hit enter to trigger the search which will make results display in a popup below the input.

Screenshot 2026-03-10 at 14 19 48

It looks like one major drawback of this is that the search results will only list entire pages. So you will get e.g. `/client-server-api/` in the results but not different matches of the search term on that page.

### Pull Request Checklist

### DIFF
--- a/assets/scss/_styles_project.scss
+++ b/assets/scss/_styles_project.scss
@@ -661,3 +661,7 @@ dd {
     margin: 0;
   }
 }
+
+.td-offline-search-results {
+  max-width: 460px;
+}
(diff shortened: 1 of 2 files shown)


Johennes:
Closing in favour of #2331 due to lack of customizability of the Lunr search backend.

PR #2334, opened 2026-03-10
Clarify that server ACLs are case-insensitive

1vel:
### Pull Request Checklist

These were implemented case insensitively in Synapse when they were first added: https://github.com/element-hq/synapse/commit/3cf3e08a97f4617763ce10da4f127c0e21d7ff1d

Some implementations (notably Ruma) treated them case sensitively however.

* [X] Pull request includes a [changelog file](https://github.com/matrix-org/matrix-spec/blob/master/CONTRIBUTING.rst#adding-to-the-changelog)

Signed-off-by: Helix K <vel@riseup.net>

Preview: https://pr2334--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/server_server/newsfragments/2334.clarification
+++ b/changelogs/server_server/newsfragments/2334.clarification
@@ -1,0 +1,1 @@
+Clarify that server ACLs are case-insensitive.
--- a/changelogs/server_server/newsfragments/2334.clarification
+++ b/changelogs/server_server/newsfragments/2334.clarification
@@ -1,1 +1,1 @@
-Clarify that server ACLs are case-insensitive.
+Clarify that server ACLs are case-insensitive, as per [MSC4436](https://github.com/matrix-org/matrix-spec-proposals/pull/4436).
--- a/data/event-schemas/schema/m.room.server_acl.yaml
+++ b/data/event-schemas/schema/m.room.server_acl.yaml
@@ -62,8 +62,9 @@ properties:
       allow:
         type: array
         description: |-
-          The case-insensitive server names to allow in the room, excluding any port information.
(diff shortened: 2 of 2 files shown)


turt2live (changes requested):
This change looks good to me, but I think it requires an MSC. I've written one and sent it for FCP: https://github.com/matrix-org/matrix-spec-proposals/pull/4436

data/event-schemas/schema/m.room.server_acl.yaml
```diff
@@ -62,7 +62,7 @@ properties:
       allow:
         type: array
         description: |-
-          The server names to allow in the room, excluding any port information.
+          The case-insensitive server names to allow in the room, excluding any port information.
           Each entry is interpreted as a [glob-style pattern](/appendices#glob-style-matching).
```
Johennes: It might just be me but I find it slightly confusing that the descriptions starts with saying that the elements are server names but then ends with saying that they are actually glob expression. Maybe it would be better to use something like this?


```suggestion
          A list of case-insensitive [glob expressions](/appendices#glob-style-matching) that are
          evaluated against server names after stripping any port information to determine the servers
          to allow in the room.
```

changelogs/server_server/newsfragments/2334.clarification
```diff
@@ -0,0 +1 @@
+Clarify that server ACLs are case-insensitive.
```
turt2live:
```suggestion
Clarify that server ACLs are case-insensitive, as per [MSC4436](https://github.com/matrix-org/matrix-spec-proposals/pull/4436).
```

turt2live (approved):
approval with changes when FCP finishes successfully

Issue #1471, opened 2023-03-21
If a server specifies multiple active verify keys, which one(s) are required to sign an object?


turt2live:
**Link to problem area**:
* https://spec.matrix.org/v1.6/server-server-api/#get_matrixkeyv2server
* https://spec.matrix.org/v1.6/appendices/#signing-details
* etc

**Issue**
No where do we actually say which of the server's keys have to sign a given message, though the general case is that a server has exactly 1 active (not-old) key in play.

richvdh:
Yes, it's never made explicit, but for the record: it is sufficient for an object to be signed by any one active key.

Additionally: neither https://spec.matrix.org/v1.6/server-server-api/#request-authentication nor https://spec.matrix.org/v1.6/server-server-api/#validating-hashes-and-signatures-on-received-events say anything about validity. Again for the record, the expectation is that:
 * the time when a S-S API request was **received**, or
 * the `origin_server_ts` on an event
 
... must lie within the validity period of the key used for the signature.

richvdh:
Fixed by #2341

PR #2341, opened 2026-03-21
Clarify how multiple signatures should be handled during verification

timedoutuk:
Clarifies the "[Validating hashes and signatures on received events](https://spec.matrix.org/unstable/server-server-api/#validating-hashes-and-signatures-on-received-events)" section of the server-to-server specification to clarify that signatures from unknown keys are skipped. This clarifies behaviour that is seen in [Synapse](https://github.com/element-hq/synapse/blob/40d699b1d4d7855ffb892723eac90cd34f22aa6f/synapse/crypto/keyring.py#L351-L357) and [ruwuma](https://forgejo.ellis.link/continuwuation/ruwuma/pulls/51), and will become necessary as policy servers are adopted, as noted in https://github.com/matrix-org/matrix-spec/issues/1471#issuecomment-4103659286.

### Pull Request Checklist

Preview: https://pr2341--matrix-spec-previews.netlify.app

### DIFF
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -1508,7 +1508,7 @@ expected to be published and therefore should be skipped at this stage.
 Additionally, any keys that are known to have expired prior to the event's
 `origin_server_ts` are ignored.
 
-If all signatures from known keys are found to be valid, the expected content hash is
+If all signatures from known unexpired keys from the originating server(s) are found to be valid, the expected content hash is
 calculated as described below. The content hash in the `hashes` property
 of the received event is base64-decoded, and the two are compared for
 equality.
--- a/content/server-server-api.md
+++ b/content/server-server-api.md
@@ -1492,37 +1492,35 @@ signature](/appendices#checking-for-a-signature). Note that this
 step should succeed whether we have been sent the full event or a
 redacted copy.
 
-The signatures expected on an event are:
+Unless the event is a 3rd party invite, only the signature(s) from the
+originating server (the server the `sender` belongs to) are required for
+verification. If a signature is from an unknown or expired key, it is skipped.
 
--   The `sender`'s server, unless the invite was created as a result of
-    3rd party invite. The sender must already match the 3rd party
-    invite, and the server which actually sends the event may be a
-    different server.
--   For room versions 1 and 2, the server which created the `event_id`.
-    Other room versions do not track the `event_id` over federation and
-    therefore do not need a signature from those servers.
+If the event is a 3rd party invite, the sender must already match the 3rd party
+invite, and the server which actually sends the event may be a different
+server.
 
-Only signatures from known server keys are validated here, any unknown keys are ignored.
+Only signatures from known server keys are validated here. Any unknown keys are ignored.
 In particular, the [policy server key](#validating-policy-server-signatures) is not
 expected to be published and therefore should be skipped at this stage.
 Additionally, any keys that are known to have expired prior to the event's
 `origin_server_ts` are ignored.
-
-If all signatures from known unexpired keys from the originating server(s) are found to be valid, the expected content hash is
-calculated as described below. The content hash in the `hashes` property
-of the received event is base64-decoded, and the two are compared for
-equality.
-
-If the hash check fails, then it is assumed that this is because we have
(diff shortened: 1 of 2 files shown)


content/server-server-api.md
```diff
@@ -1503,7 +1502,7 @@ The signatures expected on an event are:
     Other room versions do not track the `event_id` over federation and
     therefore do not need a signature from those servers.
 
-If the signature is found to be valid, the expected content hash is
+If all signatures from known keys are found to be valid, the expected content hash is
```
tulir: It might be nice to have a sentence explicitly calling this out rather than relying on "from known keys" to convey everything. Something like

> Only signatures from known server keys are validated here, any unknown keys are ignored. In particular, the policy server key is not expected to be published and therefore should be skipped at this stage.

richvdh: Also worth noting that keys that are known to have expired before the `origin_server_ts` of the event are ignored, per https://github.com/element-hq/synapse/blob/v1.150.0rc1/synapse/crypto/keyring.py#L359-L360.

richvdh: I feel like this change should live in the "checking for a signature" appendix, rather than here, since it applies to incoming requests as well as events?

timedoutuk: Requests can only have one signature, can't they?

richvdh: Hrm. In theory they could have multiple Authorization headers. Apparently that's not really compliant with RFC7320 though, so I guess they can't. Ignore me, then.

content/server-server-api.md
richvdh: should we clarify that this means that:

 * there must be at least one valid signature from these servers
 * signatures from other servers are ignored and not checked?

content/server-server-api.md
```diff
@@ -1503,7 +1502,13 @@ The signatures expected on an event are:
     Other room versions do not track the `event_id` over federation and
     therefore do not need a signature from those servers.
 
-If the signature is found to be valid, the expected content hash is
+Only signatures from known server keys are validated here, any unknown keys are ignored.
```
richvdh: comma splice.

```suggestion
Only signatures from known server keys are validated here. Any unknown keys are ignored.
```

content/server-server-api.md
richvdh: In fact, I wonder if it might help to give this list of servers a name. We seem to refer to them as "the originating server" [sic] on line 1489-90: maybe "originating server(s)" is a good enough name.

content/server-server-api.md
```diff
@@ -1503,7 +1502,13 @@ The signatures expected on an event are:
...
+In particular, the [policy server key](#validating-policy-server-signatures) is not
+expected to be published and therefore should be skipped at this stage.
+Additionally, any keys that are known to have expired prior to the event's
+`origin_server_ts` are ignored.
+
+If all signatures from known keys are found to be valid, the expected content hash is
```
richvdh:
```suggestion
If all signatures from known unexpired keys from the originating server(s) are found to be valid, the expected content hash is
```

richvdh (commented):
thanks for working on this. At the risk of gold-plating, I think it would be nice to tighten up the language a bit more, while we're in the area.

content/server-server-api.md
timedoutuk: I've shuffled this entire section around and dropped the confusing mention w.r.t. v1 & v2 rooms ([context](https://matrix.to/#/!nD4Jy1hp0We0VmIM9ubjqWLBX_uV8YlTBBPa3a_v2uk/%24WSMt0HNWVgt4LRogUVCFbuh19BaeXtBfZ7E1rw5vFDo?via=starstruck.systems&via=matrix.org&via=4d2.org)), and explicitly spelt out which server's signatures should be checked (and that others should be ignored). Let me know if you've got any other suggestions.

richvdh (approved):
TBH I'm still not sure that it's 100% clear, but this is certainly a material improvement on what's there currently, and we should probably stop attempting to gold-plate it, and just move on.

Thanks for contributing this!

content/server-server-api.md
richvdh: (we discussed in #matrix-spec: turns out those requirements for v1 and v2 rooms are still relevant.)

PR #2343, opened 2026-03-26
Try to fix npm release

turt2live:
https://github.com/matrix-org/matrix-spec/actions/runs/23603061407/job/68738177454 is the failure

There may be more we need to do, but we can find out at the next spec release.

### Pull Request Checklist

Preview: https://pr2343--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/internal/newsfragments/2343.clarification
+++ b/changelogs/internal/newsfragments/2343.clarification
@@ -1,0 +1,1 @@
+Update and fix GitHub Actions.


changelogs/internal/newsfragments/2343.clarification
```diff
@@ -0,0 +1 @@
+Update and fix GitHub Actions.
```
turt2live, line 1: This is vague because I expect there will be other PRs in the release cycle which do more GHA updates/fixes.

Issue #274, opened 2018-03-27
Add `replaces_state` to state events in the /sync response


manuroe:

richvdh:
Fixed by #2345  🎉

PR #2345, opened 2026-03-29
Specify `replaces_state` in `ClientEvent`

timedoutuk:
Related issue: https://github.com/matrix-org/matrix-spec/issues/274

### Pull Request Checklist

Preview: https://pr2345--matrix-spec-previews.netlify.app


tulir (approved):
Considering this has been in Synapse since the beginning and there are clients using it (e.g. [element web devtools](https://github.com/element-hq/element-web/blob/e0cf78b5b87cfb103b345e7e3ff29a34fb5747a8/apps/web/src/components/views/dialogs/devtools/RoomState.tsx#L54)), I'd say this doesn't need an MSC, but should get the opinion of at least one other SCT member.

data/api/client-server/definitions/client_event_without_room_id.yaml
```diff
@@ -113,6 +114,11 @@ properties:
             this.
         title: EventContent
         type: object
+      replaces_state:
+        description: |
+          The event ID of the state event replaced by this event. This field is generated
+          by the local homeserver, and is only returned if the event is a state event.
```
richvdh, line 120: It would be nice to specify whether or not it appears even if the client lacks permission to see the previous event (cf `prev_content`).

richvdh, line 120: could do with an `x-addedInMatrixVersion` property too.

richvdh (changes requested):

data/api/client-server/definitions/client_event_without_room_id.yaml
timedoutuk: The deliberate omission was meant to infer that it appeared regardless, in contrast to `prev_content`. I'll make that explicit instead

richvdh (approved):
LGTM. Thanks!

PR #2351, opened 2026-03-31
Replace `authorization` with `authorisation` throughout the spec

richvdh:
The style guide requires en_GB, which spells authorisation with an s.

Preview: https://pr2351--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -263,7 +263,7 @@ matching the threepid was found.
 : The request or entity was too large.
 
 `M_UNAUTHORIZED`
-: The request was not correctly authorized. Usually due to login failures.
+: The request was not correctly authorised. Usually due to login failures.
 
 `M_UNSUPPORTED_ROOM_VERSION`
 : The client's request to create a room used a room version that the
@@ -543,7 +543,7 @@ With the legacy API, a client can register a new account with the
 With the OAuth 2.0 API, a client can't register a new account directly. The end
 user must do that directly in the homeserver's web UI. However, the client can
 signal to the homeserver that the user wishes to create a new account with the
-[`prompt=create`](#user-registration) parameter during authorization.
+[`prompt=create`](#user-registration) parameter during authorisation.
 
 {{% boxes/note %}}
 {{% added-in v="1.17" %}}
@@ -559,7 +559,7 @@ endpoint. To invalidate the access token the client must call the [`/logout`](#p
 endpoint.
 
 With the OAuth 2.0 API, a client can obtain an access token by using one of the
-[grant types](#grant-types) supported by the homeserver and authorizing the
+[grant types](#grant-types) supported by the homeserver and authorising the
 proper [scope](#scope), as demonstrated in the [login flows](#login-flows). To
 invalidate the access token the client must use [token revocation](#token-revocation).
 
@@ -1736,7 +1736,7 @@ or best practice recommendations.
 
 The main change for end users with this API is that all the account management
 occurs in their browser on the homeserver's web UI. This is where they will
-register a new account or log into an existing account and authorize a client to
+register a new account or log into an existing account and authorise a client to
 access their Matrix account. This means that the homeserver has complete control
 over the requirements to create a new account and is not limited by the steps
 defined in this specification. It also means that less trust is given to clients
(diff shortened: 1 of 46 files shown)


turt2live (approved):
lgtm, assuming this doesn't cause problems in OAuth areas.

data/schemas/oauth2-client-metadata.yaml
```diff
@@ -111,7 +111,7 @@ properties:
       Array of the OAuth 2.0 grant types that the client may use.
 
       This MUST include:
-      - the `authorization_code` value to use the authorization code grant,
+      - the `authorisation_code` value to use the authorisation code grant,
```
hughns: This value is defined externally by [RFC 8414 OAuth 2.0 Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414) so cannot be changed.


```suggestion
      - the `authorization_code` value to use the authorisation code grant,
```

content/client-server-api/_index.md
```diff
@@ -2380,76 +2380,76 @@ This definition matches:
...
+authorisation grant lets clients on devices with limited input capabilities
+obtain an access token by having the user complete authorisation on a separate
 device with a web browser.
 
-This grant requires the client to know the following [authorization server
+This grant requires the client to know the following [authorisation server
 metadata](#server-metadata-discovery):
 
-- `device_authorization_endpoint`
+- `device_authorisation_endpoint`
```
hughns: This is specified in an RFC so cannot be changed here.

```suggestion
- `device_authorization_endpoint`
```

hughns (changes requested):
Two things:

Firstly, there are some clear errors in here which need to be addressed where constants have been renamed. See comments inline, but I could well have missed some more.

Secondly, for the OAuth related items sometimes/often we are referring to concepts or entities that are defined in external specs (e.g. RFCs). In these cases it doesn't feel "right" to have renamed them.

Examples of things that are external concepts:

- "Authorization Code": https://datatracker.ietf.org/doc/html/rfc6749#section-1.3.1
- "Authorization Code Grant": https://datatracker.ietf.org/doc/html/rfc6749#section-4.1
- "Device Authorization Grant": https://datatracker.ietf.org/doc/html/rfc8628
- "Authorization Request": https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1
- "Authorization Response": https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2
- "OAuth 2.0 Authorization Server Metadata": https://datatracker.ietf.org/doc/html/rfc8414
- "Device Authorization Request": https://datatracker.ietf.org/doc/html/rfc8628#section-3.1
- "Device Authorization Response": https://datatracker.ietf.org/doc/html/rfc8628#section-3.2

Ideas:

- we treat these external things as proper names that cannot be changed
- we make it clearer when we are referring to the proper name or not. e.g. instead of the heading "Authorization Request" perhaps "Sending the Authorization Request" with a note that this is an Authorization Request as defined by XYZ.

richvdh:
> Firstly, there are some clear errors in here which need to be addressed where constants have been renamed. See comments inline, but I could well have missed some more.

Good spots, thank you. I tried really hard to avoid changing these, so I'm sorry I missed them :/.

> Secondly, for the OAuth related items sometimes/often we are referring to concepts or entities that are defined in external specs (e.g. RFCs). In these cases it doesn't feel "right" to have renamed them.

Right, so this is the whole debate here. https://github.com/matrix-org/matrix-spec-proposals/pull/4421#discussion_r2817679368 is the relevant thread.

> we treat these external things as proper names that cannot be changed

I'm really not in favour of trying to use en_US terms when we are talking about OAuth things, and en_GB the rest of the time. I did try that when the OAuth stuff first landed in the spec, but figuring out where exactly to draw the line was a nightmare. And you get sentences like "... obtain an authorisation code via the authorization code grant", which feels odd.

Further, the existence of https://auth0.com/fr/intro-to-iam/what-is-oauth-2 makes me think that one certainly *can* i18nise these terms.

> we make it clearer when we are referring to the proper name or not. e.g. instead of the heading "Authorization Request" perhaps "Sending the Authorization Request" with a note that this is an Authorization Request as defined by XYZ.

I'm not really sure how practical this is. The problem is that these terms are repeated many times, and giving a reference each time they do is painful.

----

My only opinion here is that attempting a half-way house is unachievable; we should commit to one of en_GB or en_US. I don't really have a strong opinion on *which*, but I am a bit fed up with our current attempts to sit on the fence.

We got pushback against switching en_US, so this is my attempt to try en_GB. If you think this PR is worse, then I would welcome comments on MSC4421 :)

richvdh:
> > Firstly, there are some clear errors in here which need to be addressed where constants have been renamed. See comments inline, but I could well have missed some more.
> 
> Good spots, thank you. I tried really hard to avoid changing these, so I'm sorry I missed them :/.

I ran this to double-check for further changes within identifiers:

```
$ git diff origin/main...HEAD | grep '^\+' | grep '`[^ `]*authoris'
$
```

richvdh:
> If you think this PR is worse, then I would welcome comments on [MSC4421](https://github.com/matrix-org/matrix-spec-proposals/issues/4421) :)

MSC4421 is in final call for closure.

hughns (approved):
The OAuth 2 API related changes now look sensible to me.

richvdh:
MSC4421 has been rejected, so I'm merging this.

PR #2366, opened 2026-04-28
Clarify our stance on en_GB vs en_US

richvdh:
... per the closure of MSC4421

Preview: https://pr2366--matrix-spec-previews.netlify.app

### DIFF
--- a/meta/documentation_style.rst
+++ b/meta/documentation_style.rst
@@ -19,7 +19,7 @@ nested titles (h6, or 6 `#` characters) and instead re-evaluate the document str
 Correct capitalisation for long section names
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Headings should be in `sentence case <https://apastyle.apa.org/style-grammar-guidelines/capitalization/sentence-case>`,
+Headings should be in `sentence case <https://apastyle.apa.org/style-grammar-guidelines/capitalization/sentence-case>`_,
 as represented by this document.
 
 TODOs
@@ -59,7 +59,7 @@ General
     Of course, identifiers used within the protocol itself must use the
     spellings defined by the protocol. So, for example, the
     `authorization_endpoint` property in the `OAuth 2.0
-    metadata <https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1auth_metadata>`
+    metadata <https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1auth_metadata>`_
     uses the same spelling as defined in RFC8414.
 
 * The word "homeserver" is spelt thus (rather than "home server", "Homeserver",
--- a/meta/documentation_style.rst
+++ b/meta/documentation_style.rst
@@ -47,7 +47,20 @@ General
 * Try to write clearly and unambiguously. Remember that many readers will not
   have English as their first language.
 
-* Prefer British English (colour, -ise) to American English.
+* The Matrix spec uses British English rather than American English. For example, the
+  words "colour" and "authorise" use British spellings. The
+  `join_authorised_via_users_server` property in `m.room.member` events is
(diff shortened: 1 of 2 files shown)


meta/documentation_style.rst
```diff
@@ -47,7 +47,20 @@ General
 * Try to write clearly and unambiguously. Remember that many readers will not
   have English as their first language.
 
-* Prefer British English (colour, -ise) to American English.
+* The Matrix spec uses British English rather than American English. For example, the
+  words "colour" and "authorise" use British spellings. The
+  `join_authorised_via_users_server` property in `m.room.member` events is
```
richvdh: this was the best example I could find of a protocol identifier using British spelling. Better examples welcome.

anoadragon453:
```suggestion
  ``join_authorised_via_users_server`` property in ``m.room.member`` events is
```

Reminder that this document is in RST, which is ironic as it calls for documentation to be written in Markdown.

meta/documentation_style.rst
```diff
@@ -47,7 +47,20 @@ General
...
+  spelt accordingly.
+
+  * This extends to terms defined in other specifications: for example, in the
+    context of OAuth 2.0, we refer to an "authorisation code grant", even though
+    RFC6749 uses the spelling "authorization".
+
+    Of course, identifiers used within the protocol itself must use the
+    spellings defined by the protocol. So, for example, the
+    `authorization_endpoint` property in the `OAuth 2.0
```
anoadragon453:
```suggestion
    ``authorization_endpoint`` property in the `OAuth 2.0
```

meta/documentation_style.rst
```diff
@@ -47,7 +47,20 @@ General
...
+    metadata <https://spec.matrix.org/v1.18/client-server-api/#get_matrixclientv1auth_metadata>`_
+    uses the same spelling as defined in RFC8414.
```
anoadragon453: May be worth linkifying the RFC, so one can have a look inside and see the US spellings for themselves.

meta/documentation_style.rst
anoadragon453: I tasked GPT-5.5 with the problem and ironically this was the only example it could find in the entire spec. The rest used US English:

Error codes:

* `M_UNAUTHORIZED`
* `M_UNRECOGNIZED`

HTML attribute in formatted message bodies:

* `data-mx-color`
* `data-mx-bg-color`

...and that's the full list.

It's still quite small either way. We can still easily shift to British English going forward.

anoadragon453 (changes requested):
Thank you for taking this one on!

meta/documentation_style.rst
richvdh: ghahaha. We should really convert it to MD.

meta/documentation_style.rst
richvdh:
```suggestion
    uses the same spelling as defined in `RFC8414 <https://www.rfc-editor.org/rfc/rfc8414#section-2>`_.
```

meta/documentation_style.rst
richvdh: Well, it missed [`m.unauthorised`](https://spec.matrix.org/v1.18/client-server-api/#mroom_keywithheld). But nm.

clokep: Edit: I read your comment as the opposite -- following US identifiers even though Matrix spec is GB. So ignore this...

~~Maybe not the best example, but the `Referer` header is the canonical example of following what the spec says, even though the word is actually misspelled. Not really a US/GB example though.~~

~~We also use the `Authorization` header, not 'Authorisation` from [RFC7235](https://datatracker.ietf.org/doc/html/rfc7235).~~

clokep (approved):

anoadragon453 (approved):
Thank you again for getting this over the line!

PR #2368, opened 2026-04-30
appendices: update canonical JSON grammar

famfo:
### Pull Request Checklist

Preview: https://pr2368--matrix-spec-previews.netlify.app

### DIFF
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -145,10 +145,11 @@ object    = "{" [ member *( "," member ) ] "}"
 member    = string ":" value
 array     = "[" [ value *( "," value ) ] "]"
 number    = [ "-" ] int
-int       = %x30 / ( %x31-39 *DIGIT )   ; 0 or 1-9 *DIGIT (0-9)
-string    = DQUOTE *char DQUOTE         ; quoted chars
+int       = %x30 / ( %x31-39 *DIGIT )   ; Integer without leading zeros
+string    = DQUOTE *char DQUOTE         ; Quoted chars
 char      = unescaped / "\" escaped
-unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
+unescaped = %x20-21 / %x23-5B / %x5D-10FFFF ; All UTF-8 codepoints except ASCII control
+                                            ; characters, " and \
 escaped   = %x62 ; b    backspace       U+0008
           / %x74 ; t    tab             U+0009
           / %x6E ; n    line feed       U+000A
@@ -156,9 +157,10 @@ escaped   = %x62 ; b    backspace       U+0008
           / %x72 ; r    carriage return U+000D
           / %x22 ; "    quotation mark  U+0022
           / %x5C ; \    reverse solidus U+005C
-          / %s"u000" (%x30-37 / %x62 / %x65-66) ; all codepoints from u000X excluding the ones
-                                                ; explicitly escaped above (0-7, b, e, f)
-          / %s"u000" (%x30-39 / %x61-66)        ; u001X (0-9, a-f)
+          / %s"u000" (%x30-37 / %x62 / %x65-66) ; All ASCII control characters which have not been
+                                                ; escaped otherwise (for example \n).
+                                                ; u000X, where X is [0-7, b, e, f]
+          / %s"u001" (%x30-39 / %x61-66)        ; u001X, where X is [0-9, a-f]
 ```
 
 #### Examples
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -131,32 +131,35 @@ def canonical_json(value):
 
 #### Grammar
 
-Adapted from the grammar in <http://tools.ietf.org/html/rfc7159>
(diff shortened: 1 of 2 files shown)


content/appendices.md
```diff
@@ -131,32 +131,35 @@ def canonical_json(value):
...
+null      = %s"null"
+true      = %s"true"
+object    = "{" [ member *( "," member ) ] "}"
+member    = string ":" value
+array     = "[" [ value *( "," value ) ] "]"
+number    = [ "-" ] int
+int       = %x30 / ( %x31-39 *DIGIT )   ; 0 or 1-9 *DIGIT (0-9)
+string    = DQUOTE *char DQUOTE         ; quoted chars
+char      = unescaped / "\" escaped
+unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
```
zecakeh: This could be clarified too as all UTF-8 codepoints except control characters, `"` and `\`

content/appendices.md
```diff
@@ -131,32 +131,35 @@ def canonical_json(value):
...
+```
+value     = false / null / true / object / array / number / string
+false     = %s"false"
+null      = %s"null"
+true      = %s"true"
+object    = "{" [ member *( "," member ) ] "}"
+member    = string ":" value
+array     = "[" [ value *( "," value ) ] "]"
+number    = [ "-" ] int
+int       = %x30 / ( %x31-39 *DIGIT )   ; 0 or 1-9 *DIGIT (0-9)
```
zecakeh: This could be clarified as an integer without leading zeroes?

content/appendices.md
```diff
@@ -131,32 +131,35 @@ def canonical_json(value):
...
+unescaped = %x20-21 / %x23-5B / %x5D-10FFFF
+escaped   = %x62 ; b    backspace       U+0008
+          / %x74 ; t    tab             U+0009
+          / %x6E ; n    line feed       U+000A
+          / %x66 ; f    form feed       U+000C
+          / %x72 ; r    carriage return U+000D
+          / %x22 ; "    quotation mark  U+0022
+          / %x5C ; \    reverse solidus U+005C
+          / %s"u000" (%x30-37 / %x62 / %x65-66) ; all codepoints from u000X excluding the ones
+                                                ; explicitly escaped above (0-7, b, e, f)
```
zecakeh: The comment is not exactly true, a few others codepoints are excluded. I also find it confusing that the values in parentheses are the values of `X`. I would expect a list in parentheses to describe the actual list of allowed characters like in other comments, rather than the allowed codepoint suffixes.

Maybe we should say explicitly that those are the possible values of `X`?

content/appendices.md
```diff
@@ -131,32 +131,35 @@ def canonical_json(value):
...
+          / %s"u000" (%x30-39 / %x61-66)        ; u001X (0-9, a-f)
```
zecakeh:
```suggestion
          / %s"u001" (%x30-39 / %x61-66)        ; u001X (0-9, a-f)
```

zecakeh: I have the same issue that it's not clear in the comment that the values in parentheses are the values of `X`. These are actually all the codepoints so we could just say that?

zecakeh (changes requested):

content/appendices.md
famfo: Hm, maybe a comment over u000x and u001x saying "All ASCII control characters which have not otherwise been escaped (for example \n)" in addition to a possible list for X could explain the rationale a bit better.

richvdh:
```suggestion
          / %s"u000" (%x30-37 / %x62 / %x65-66) ; u000X, where X is [0-7, b, e, f]
```

possibly?

content/appendices.md
richvdh:
```suggestion
          / %s"u000" (%x30-39 / %x61-66)        ; u001X, where X is [0-9, a-f]
```

content/appendices.md
famfo: This is missing any context as for why again, I like the list though.

zecakeh (approved):
This looks good to me now.

content/appendices.md
```diff
@@ -131,32 +131,37 @@ def canonical_json(value):
...
+                                            ; characters, " and \
+escaped   = %x62 ; b    backspace       U+0008
+          / %x74 ; t    tab             U+0009
+          / %x6E ; n    line feed       U+000A
+          / %x66 ; f    form feed       U+000C
+          / %x72 ; r    carriage return U+000D
+          / %x22 ; "    quotation mark  U+0022
+          / %x5C ; \    reverse solidus U+005C
+          / %s"u000" (%x30-37 / %x62 / %x65-66) ; All ASCII control characters which have not been
+                                                ; escaped otherwise (for example \n).
```
richvdh:
```suggestion
          / %s"u000" (%x30-37 / %x62 / %x65-66) ; All ASCII control characters which do not have
                                                ; dedicated escape sequences (for example \n).
```

richvdh (approved):

PR #2383, opened 2026-05-21
replace "Olm message" where not referring to the Olm message format

gewitternacht:
### Pull Request Checklist
The term "Olm message" is used in the Olm specification, referring to the specific byte-encoded message format.
The term is also used in the Client-Server spec, but "event encrypted using Olm" (a wording also already used there) seems like a better fit in those places to avoid confusion.

Signed-off-by: Johanna Stuber <johannas@element.io>

Preview: https://pr2383--matrix-spec-previews.netlify.app


anoadragon453 (approved):
Clearer indeed. Thank you!

PR #2396, opened 2026-06-10
Clarify that clients must avoid producing ambiguous matrix.to URIs

HarHarLinks:
The spec already says this, yet clients tend to ignore what is clearly a requirement here, so this is an attempt to clarify. Note the notice further down about still trying to interpret unencoded (incorrect) URIs at best effort is left intact.

See also https://github.com/matrix-org/matrix.to/pull/404.

Updated the section about matrix.to with RFC 2119 lingo while at it.

### Pull Request Checklist

Preview: https://pr2396--matrix-spec-previews.netlify.app


content/appendices.md
```diff
@@ -887,20 +887,21 @@ https://matrix.to/#/<identifier>/<extra parameter>?<additional arguments>
...
+Clients SHOULD NOT rely on matrix.to URIs falling back to a web server
+if accessed and instead SHOULD perform some sort of action within the
 client. For example, if the user were to click on a matrix.to URI for a
-room alias, the client may open a view for the user to participate in
+room alias, the client MAY open a view for the user to participate in
 the room.
 
 The components of the matrix.to URI (`<identifier>` and
 `<extra parameter>`) MUST be percent-encoded as per RFC 3986.
+Clients MUST NOT produce incorrectly encoded URIs to avoid ambiguous interpretation.
```
anoadragon453:
```suggestion
Clients MUST NOT produce incorrectly encoded URIs to avoid ambiguous
interpretation.
```

anoadragon453: What point is this making other than making the previous statement harder to skim over? Assuming that "incorrectly encoded" means "not percent-encoded".

anoadragon453 (changes requested):
Thanks for fixing all the lowercase verbiage!

content/appendices.md
HarHarLinks: historically, major implementations have ignored this:
- https://github.com/matrix-org/matrix.to/issues/251
- https://github.com/element-hq/element-web/issues/32611

HarHarLinks: "incorrectly encoded" means "not encoded according to spec" :shrug:

richvdh:
@anoadragon453 and @HarHarLinks: whose court is this in?

anoadragon453:
@richvdh I believe my own, to respond to the above questions from @HarHarLinks.

Edit: which I've now done.

content/appendices.md
anoadragon453: I see. Another approach could be to provide an example of *why* clients shouldn't do this. The linked matrix.to issue is perfect.

matrix.to thought that they could skirt the rules to make URLs more legible, and you showed a clear case of *why* this resulted in pain downstream.

So perhaps we could say:

```
The components of the matrix.to URI (`<identifier>` and
`<extra parameter>`) MUST be percent-encoded as per RFC 3986.
Failure to do so will result in downstream software misinterpreting
the links as invalid/not turning them into clickable links in UI.
```

HarHarLinks: Sounds sounds great to me!

```suggestion
Failure to do so will result in downstream software misinterpreting
the links as invalid/not turning them into clickable links in UI.
```

anoadragon453 (approved):
Thanks! (and thanks @richvdh for the poke :)

PR #2398, opened 2026-06-12
Use OpenAPI definition for `EncryptedFile`

richvdh:
... instead of the ad-hoc table we have currently.

Since it uses url-safe unpadded base64, I specced that too.

Preview: https://pr2398--matrix-spec-previews.netlify.app

### DIFF
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -51,7 +51,7 @@ interoperability.
 
 URL-safe unpadded Base64 is identical to standard unpadded Base64, except that
 it uses `-` (minus) as the 62nd character in the alphabet, and `_` (underscore)
-as the 63rd.
+as the 63rd. This matches [RFC4648’s definition of URL-safe base64](https://tools.ietf.org/html/rfc4648#section-5).
 
 ## Binary data
 
--- a/content/appendices.md
+++ b/content/appendices.md
@@ -12,7 +12,7 @@ Specifically, where RFC 4648 requires that encoded data be padded to a
 multiple of four characters using `=` characters, unpadded Base64 omits
 this padding.
 
-For reference, RFC 4648 uses the following alphabet for Base 64:
+For reference, RFC 4648 uses the following alphabet for Base64:
 
     Value Encoding  Value Encoding  Value Encoding  Value Encoding
         0 A            17 R            34 i            51 z
@@ -46,6 +46,12 @@ Examples of strings encoded using unpadded Base64:
 When decoding Base64, implementations SHOULD accept input with or
 without padding characters wherever possible, to ensure maximum
 interoperability.
+
+### URL-safe unpadded Base64
+
+URL-safe unpadded Base64 is identical to standard unpadded Base64, except that
+it uses `-` (minus) as the 62nd character in the alphabet, and `_` (underscore)
+as the 63rd.
 
 ## Binary data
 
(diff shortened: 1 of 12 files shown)


data/api/client-server/definitions/encrypted_file.yaml
```diff
@@ -0,0 +1,84 @@
...
+        example: true
+    required:
+      - kty
+      - key_ops
+      - alg
+      - k
+      - ext
+  iv:
+    type: string
+    description: The 128-bit unique counter block used by AES-CTR, encoded as [unpadded Base64](/appendices/#unpadded-base64).
```
zecakeh, line 64: We have a custom string format for this:


```suggestion
    format: mx-unpadded-base64
    description: The 128-bit unique counter block used by AES-CTR, encoded as [unpadded Base64](/appendices/#unpadded-base64).
```

data/api/client-server/definitions/encrypted_file.yaml
```diff
@@ -0,0 +1,84 @@
...
+    example: "w+sE15fzSc0AAAAAAAAAAA"
+  hashes:
+    type: object
+    title: EncryptedFileHashes
+    description:
+      A map from an algorithm name to a hash of the ciphertext. Clients MUST support the SHA-256 hash, which uses the key `sha256`.
+    properties:
+      sha256:
+        type: string
+        description: The hash of the ciphertext. encoded as [unpadded Base64](/appendices/#unpadded-base64).
```
zecakeh, line 75: Same here.


```suggestion
        format: mx-unpadded-base64
        description: The hash of the ciphertext. encoded as [unpadded Base64](/appendices/#unpadded-base64).
```

data/api/client-server/definitions/encrypted_file.yaml
```diff
@@ -0,0 +1,84 @@
...
+          type: string
+        description: Key operations. Must at least contain `encrypt` and `decrypt`.
+        example: ["encrypt", "decrypt"]
+      alg:
+        type: string
+        description: Algorithm. Must be `A256CTR`.
+        example: A256CTR
+      k:
+        type: string
+        description: The key, encoded as [URL-safe unpadded Base64](/appendices/#url-safe-unpadded-base64).
```
zecakeh, line 49: We could add a custom string format for this, now that we have a section to link to.

zecakeh (commented):
This looks good to me, I just have nits about the string formats.

richvdh:
@zecakeh thanks for the review; you raise good points. Please take another look?

content/appendices.md
```diff
@@ -47,6 +47,12 @@ When decoding Base64, implementations SHOULD accept input with or
 without padding characters wherever possible, to ensure maximum
 interoperability.
 
+### URL-safe unpadded Base64
+
+URL-safe unpadded Base64 is identical to standard unpadded Base64, except that
+it uses `-` (minus) as the 62nd character in the alphabet, and `_` (underscore)
+as the 63rd.
```
zecakeh: Maybe we could a link to [RFC4648’s definition of URL-safe base64](https://tools.ietf.org/html/rfc4648#section-5), like in the definition of [the format event IDs in the room versions](https://spec.matrix.org/v1.18/rooms/v12/#event-ids)?

zecakeh (approved):
I have a last suggestion, but this looks ready to merge if you don't want to apply it.

content/appendices.md
richvdh: Good idea.

```suggestion
as the 63rd. This matches [RFC4648’s definition of URL-safe base64](https://tools.ietf.org/html/rfc4648#section-5).
```

PR #2399, opened 2026-06-15
Specify history sharing

richvdh:
Spec for MSC4268.

~~Based on #2398.~~

### Pull Request Checklist

Preview: https://pr2399--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1564,11 +1564,11 @@ before inviting him.
 Only room keys which were marked as "shareable" by the creator of the
 encryption session should be shared with new users.
 
-When Alice wants to send a message in the room she shares with Bob, she first
+When a user wants to send a message in an encrypted room, their client first
 checks the [history visibility](#room-history-visibility) state of the
-room. If it is `shared` or `world_readable`, then when she sends the Megolm
-keys to Bob via an [`m.room_key`](#mroom_key) message, she sets
-`shared_history` to `true`.
+room. If it is `shared` or `world_readable`, then when the client sends the Megolm
+keys to room members via [`m.room_key`](#mroom_key) messages, it SHOULD set
+`shared_history` to `true`. Otherwise, it SHOULD set `shared_history` to `false`.
 
 If the history visibility changes in a way that would affect the
 `shared_history` flag (i.e., it changes from `joined` or `invited` to `shared`
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -355,7 +355,7 @@ key, and encrypts the file using AES-CTR. The counter is 64 bits long,
 starting at 0 and prefixed by a random 64-bit Initialization Vector (IV),
 which together form a 128-bit unique counter block.
 
-Clients MUST generate both the AES key and IV using a cryptographically 
+Clients MUST generate both the AES key and IV using a cryptographically
 secure random source and MUST NOT use the same key or IV multiple
 times. The latter 64 bits of the 128-bit counter block MUST start at zero.
 
@@ -385,31 +385,11 @@ Key](https://tools.ietf.org/html/rfc7517#appendix-A.3) format, with a
 ###### Extensions to `m.room.message` msgtypes
 
 This module adds `file` and `thumbnail_file` properties, of type
-`EncryptedFile`, to `m.room.message` msgtypes that reference files, such
-as [m.file](#mfile) and [m.image](#mimage), replacing the `url` and `thumbnail_url`
-properties.
+[`EncryptedFile`](#definition-encryptedfile), to `m.room.message` msgtypes that
+reference files, such as [m.file](#mfile) and [m.image](#mimage), replacing the
+`url` and `thumbnail_url` properties.
 
-`EncryptedFile`
-
-| Parameter | Type             | Description                                                                                    |
-|-----------|------------------|------------------------------------------------------------------------------------------------|
-| url       | string           | **Required.** The URL to the file.                                                             |
-| key       | JWK              | **Required.** A [JSON Web Key](https://tools.ietf.org/html/rfc7517#appendix-A.3) object.       |
-| iv        | string           | **Required.** The 128-bit unique counter block used by AES-CTR, encoded as unpadded base64.    |
-| hashes    | {string: string} | **Required.** A map from an algorithm name to a hash of the ciphertext, encoded as unpadded base64. Clients MUST support the SHA-256 hash, which uses the key `sha256`. |
-| v         | string           | **Required.** Version of the encrypted attachment's protocol. Must be `v2`.                    |
-
-`JWK`
-
-| Parameter | Type     | Description                                                                                                              |
-| --------- |----------|--------------------------------------------------------------------------------------------------------------------------|
-| kty       | string   | **Required.** Key type. Must be `oct`.                                                                                   |
-| key_ops   | [string] | **Required.** Key operations. Must at least contain `encrypt` and `decrypt`.                                             |
-| alg       | string   | **Required.** Algorithm. Must be `A256CTR`.                                                                              |
-| k         | string   | **Required.** The key, encoded as urlsafe unpadded base64.                                                               |
-| ext       | boolean  | **Required.** Extractable. Must be `true`. This is a [W3C extension](https://w3c.github.io/webcrypto/#iana-section-jwk). |
-
-Example:
+Example `m.room.message` event containing an encrypted image:
 
 ```json
 {
@@ -469,6 +449,8 @@ Example:
   }
 }
 ```
+
+{{% definition path="api/client-server/definitions/encrypted_file" %}}
 
 #### Device verification
 
@@ -764,7 +746,7 @@ The process between Alice and Bob verifying each other would be:
 15. Assuming they match, Alice and Bob's devices each calculate Message
     Authentication Codes (MACs) for:
     * {{% changed-in v="1.18" %}} Each of the keys that they wish the other user
-      to verify (usually their device ed25519 key and their master signing key, 
+      to verify (usually their device ed25519 key and their master signing key,
       see below). The master signing key SHOULD be included when two different
       users are verifying each other. Verifying individual devices of other
       users is deprecated.
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1982,6 +1982,32 @@ When a client is updating a Megolm session in its store, the client MUST ensure:
   user, or from a `m.room_key` event.
 * that the new session key has a lower message index than the existing session key.
 
+When encrypting outgoing messages in a room using Megolm, clients MUST rotate
+their outgoing Megolm session (i.e. discard the existing session, and create
+and share a new session before sending more room messages) whenever any of the
+following happens:
+
+ * The existing session has been in use for longer than the period specified in
+   `rotation_period_ms` in the [`m.room.encryption`](#mroomencryption) room
+   state event, or an appropriate default.
+
+ * The existing session has been used to encrypt as many messages as specified in
+   `rotation_period_msgs` in the [`m.room.encryption`](#mroomencryption) room
+   state event, or an appropriate default.
+
+ * A user or device that was previously participating in the room, and may have
+   received a copy of the decryption keys for the session, is seen to leave the
+   room.
+
+   {{% changed-in v="1.19" %}} Since any user that received an invite to the
+   room may have received a copy of the decryption keys for the session via
+   [history sharing](#sharing-keys-between-users), clients MUST observe changes
+   in state in the room, and whenever they see a user leaving the room, assume
+   that the departed user may have access to any existing Megolm session, and
+   rotate the session. Note that, in a `limited` [sync](#syncing), clients must
+   treat any membership event with a membership other than `join` as an
+   indication that the affected user may have joined and left the room.
+
 #### Protocol definitions
 
 ##### Events
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1967,7 +1967,7 @@ In order to enable end-to-end encryption in a room, clients can send an
 When creating a Megolm session in a room, clients must share the
 corresponding session key using Olm with the intended recipients, so
 that they can decrypt future messages encrypted using this session. An
-`m.room_key` event is used to do this. Clients must also handle
+[`m.room_key`](#mroom_key) event is used to do this. Clients must also handle
 `m.room_key` events sent by other devices in order to decrypt their
 messages.
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1575,7 +1575,7 @@ If the history visibility changes in a way that would affect the
 or `world_readable`, or vice versa), then clients MUST rotate their outbound
 Megolm session before sending more messages.
 
-Clients SHOULD show a visual indication to users that their encrypted messages
+Clients SHOULD show an indication to users that their encrypted messages
 may be shared with future room members in this way.
 
 Recipients SHOULD keep a record of the `shared_history` flag for each
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1624,7 +1624,7 @@ Otherwise, before inviting Bob to a room, Alice's client constructs and sends a 
    containing the sessions she is aware of in the room. Alice MUST include
    only [shareable encryption sessions](#shareable-encryption-sessions) in the
    `room_keys` section of the structure; other sessions SHOULD be listed in the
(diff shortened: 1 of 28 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+keys to Bob via an [`m.room_key`](#mroom_key) message, she sets
+`shared_history` to `true`.
+
+If the history visibility changes in a way that would affect the
+`shared_history` flag (i.e., it changes from `joined` or `invited` to `shared`
+or `world_readable`, or vice versa), then clients MUST rotate their outbound
+megolm session before sending more messages.
+
+Clients SHOULD show a visual indication to users that their encrypted messages
+may be shared with future room members in this way.
```
Johennes: Possibly splitting hairs but do we need to generalise this slightly to also cover clients with a command-line interface or no interface at all (bots)? I'm always unsure if and how to add UX requirements to the spec.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+{{% event event="m.room_key_bundle" %}}
+
+##### Receiving a key bundle event
+
+When Bob's client receives an `m.room_key_bundle` event from Alice, there are two possibilities:
+
+ * If Bob has recently accepted an invite to the room from Alice, the client
+   should immediately download and decrypt the key bundle and start processing
+   it. Note, however, that this process must be resilient to Bob's client being
+   restarted before the download/import completes.
```
Johennes, line 1656: Should this be normative?

```suggestion
 * If Bob has recently accepted an invite to the room from Alice, the client
   SHOULD immediately download and decrypt the key bundle and start processing
   it. Note, however, that this process must be resilient to Bob's client being
   restarted before the download/import completes.
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+
+   The definition of "recently" is left up to clients. (They should consider
+   balancing the needs of (a) a user that closes their client just after
+   joining a room but before the bundle is imported, against (b) the overhead
+   of attempting to download a key bundle on every startup. 24 hours is a
+   recommended time limit.)
+
+ * Otherwise, Bob's client should store the details of the key bundle but not
+   download it immediately. If he later accepts an invite to the room from
+   Alice, his client downloads and processes the bundle at that point.
```
Johennes: Same question as above.

```suggestion
 * Otherwise, Bob's client SHOULD store the details of the key bundle but not
   download it immediately. If he later accepts an invite to the room from
   Alice, his client downloads and processes the bundle at that point.
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+
+Alice MAY choose not to share any room history (even messages sent when the
+history visibity setting would allow sharing) if the current history
+visibility setting does not allow sharing (i.e. if `history_visibility` is
+set to `invited` or `joined`).
+
+Otherwise, before inviting Bob to a room, Alice constructs and sends a key bundle as follows:
+
+1. Alice SHOULD ensure that she has downloaded all keys relevant to the room
+   from [server-side key backup](#server-side-key-backups), if she is using it.
```
Johennes: I'm wondering if we should rephrase this section so that the requirements and actions are worded with respect to Alice's client rather than herself? The section below about Bob does it this way and it feels a little cleaner to me, if a bit more wordy.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+If the history visibility changes in a way that would affect the
+`shared_history` flag (i.e., it changes from `joined` or `invited` to `shared`
+or `world_readable`, or vice versa), then clients MUST rotate their outbound
+megolm session before sending more messages.
+
+Clients SHOULD show a visual indication to users that their encrypted messages
+may be shared with future room members in this way.
+
+Recipients SHOULD keep a record of the `shared_history` flag for each
+encryption session. The state of the flag is also saved in the
```
Johennes, line 1582: Likely a stupid question but why do we RECOMMEND keeping a local record here?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+When Alice wants to send a message in the room she shares with Bob, she first
+checks the [history visibility](#room-history-visibility) state of the
+room. If it is `shared` or `world_readable`, then when she sends the Megolm
+keys to Bob via an [`m.room_key`](#mroom_key) message, she sets
+`shared_history` to `true`.
+
+If the history visibility changes in a way that would affect the
+`shared_history` flag (i.e., it changes from `joined` or `invited` to `shared`
+or `world_readable`, or vice versa), then clients MUST rotate their outbound
+megolm session before sending more messages.
```
Johennes: Looks like it's capitalised everywhere else.

```suggestion
Megolm session before sending more messages.
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1863,6 +1982,32 @@ When a client is updating a Megolm session in its store, the client MUST ensure:
...
+
+ * The existing session has been used to encrypt as many messages as specified in
+   `rotation_period_msgs` in the [`m.room.encryption`](#mroomencryption) room
+   state event, or an appropriate default.
+
+ * A user or device that was previously participating in the room, and may have
+   received a copy of the decryption keys for the session, is seen to leave the
+   room.
+
+   {{% changed-in v="1.19" %}} Since any user that received an invite to the
```
Johennes, line 2002: Should this not also list the new requirement to rotate the session when history visibility changes?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1565,6 +1547,143 @@ objects described as follows:
...
+Otherwise, before inviting Bob to a room, Alice constructs and sends a key bundle as follows:
+
+1. Alice SHOULD ensure that she has downloaded all keys relevant to the room
+   from [server-side key backup](#server-side-key-backups), if she is using it.
+
+2. Alice constructs a [`RoomKeyBundle`](#definition-roomkeybundle) structure,
+   containing the sessions she is aware of in the room. Alice SHOULD include
+   only [shareable encryption sessions](#shareable-encryption-sessions) in the
+   `room_keys` section of the structure; other sessions should be listed un the
+   with `withheld` section.
```
Johennes: I probably missed it when scanning the MSC but I was surprised that this is a SHOULD. Does that not undermine the `shared_history` setting and allow for sharing keys that should be deemed unshareable?

content/client-server-api/modules/end_to_end_encryption.md
richvdh: Yeah, I'll just drop the word `visual` here, which I hope will cover it. Of course there are also accessibility concerns with specifying a specifically *visual* indication.

```suggestion
Clients SHOULD show an indication to users that their encrypted messages
may be shared with future room members in this way.
```

content/client-server-api/modules/end_to_end_encryption.md
richvdh: Yup.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: Not a stupid question.

I dunno really. I try to reserve MUST for circumstances where the protocol's going to stop working if implementations don't do it, as opposed to where the user's going to get a poor experience (if you want to make a client that gives your user a crappy experience, that's on you). It gets more complicated when we're talking about giving *other* users a poor experience, as in this case, but still I don't feel like it crosses the line into needing a MUST. Hard to justify objectively though.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: yup, agreed.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: yup, good point. Let's change this.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: good spot.

richvdh (commented):
@Johennes thanks for the review; I think I have addressed all your comments.

content/client-server-api/modules/end_to_end_encryption.md
Johennes: That sounds sensible, thanks for elaborating.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1547,6 +1547,143 @@ objects described as follows:
...
+Otherwise, before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
+
+1. Alice's client SHOULD ensure that it has downloaded all keys relevant to the room
+   from [server-side key backup](#server-side-key-backups), if she is using it.
+
+2. Alice's client constructs a [`RoomKeyBundle`](#definition-roomkeybundle) structure,
+   containing the sessions she is aware of in the room. Alice MUST include
+   only [shareable encryption sessions](#shareable-encryption-sessions) in the
+   `room_keys` section of the structure; other sessions SHOULD be listed in the
+   with `withheld` section.
```
zecakeh:
```suggestion
   `withheld` section.
```

data/api/client-server/definitions/room_key_bundle.yaml
```diff
@@ -0,0 +1,104 @@
...
+      description: |-
+        The format of a session key, when shared as part of a `RoomKeyBundle`.
+      properties:
+        algorithm:
+          type: string
+          description: |-
+            The end-to-end message encryption algorithm that the key is for. Must be `m.megolm.v1.aes-sha2`.
+          example: "m.megolm.v1.aes-sha2"
+        room_id:
+          type: string
```
zecakeh, line 43:
```suggestion
          type: string
          format: mx-room-id
          pattern: "^!"
```

data/api/client-server/definitions/room_key_bundle.yaml
```diff
@@ -0,0 +1,104 @@
...
+          description: |-
+            The room where the session is used.
+          example: "!Cuyf34gef24t:localhost"
+        sender_claimed_keys:
+          type: object
+          additionalProperties:
+            type: string
```
zecakeh, line 52:
```suggestion
            type: string
            format: mx-unpadded-base64
```

data/api/client-server/definitions/room_key_bundle.yaml
```diff
@@ -0,0 +1,104 @@
...
+          description: |-
+            A map from algorithm name (`ed25519`) to the Ed25519 signing key of
+            the device which initiated the session originally, according to the
+            creator of this key bundle.
+          example: { "ed25519": "aj40p+aw64yPIdsxoog8jhPu9i7l7NcFRecuOQblE3Y" }
+        sender_key:
+          type: string
```
zecakeh, line 60:
```suggestion
          type: string
          format: mx-unpadded-base64
```

data/api/client-server/definitions/room_key_bundle.yaml
```diff
@@ -0,0 +1,104 @@
...
+        session_id:
+          type: string
+          description: |-
+            The Megolm session ID.
+          example: "X3lUlvLELLYxeTx4yOVu6UDpasGEVO0Jbu+QFnm0cKQ"
+        session_key:
+          type: string
+          description: |-
+            Unpadded base64-encoded session key in [session-export
+            format](https://gitlab.matrix.org/matrix-org/olm/blob/master/docs/megolm.md#session-export-format).
```
zecakeh: This can be a link to spec.matrix.org:


```suggestion
            format](/olm-megolm/megolm/#session-export-format).
```

data/api/client-server/definitions/room_key_bundle.yaml
```diff
@@ -0,0 +1,104 @@
...
+            initiated the session originally, according to the creator of this
+            key bundle.
+          example: "RF3s+E7RkTQTGF2d8Deol0FkQvgII2aJDf3/Jp5mxVU"
+        session_id:
+          type: string
+          description: |-
+            The Megolm session ID.
+          example: "X3lUlvLELLYxeTx4yOVu6UDpasGEVO0Jbu+QFnm0cKQ"
+        session_key:
+          type: string
```
zecakeh, line 73:
```suggestion
          type: string
          format: mx-unpadded-base64
```

zecakeh (dismissed):
This looks good to me, I just found a grammar error, and string formats can be clarified in the definitions. Note that I only did the definitions of `room_key_bundle.yaml` but `room_key_withheld_content.yaml` and `m.room_key_bundle.yaml` could also use string formats on the `room_id` and `sender_key` properties.

richvdh:
Oops, I forgot to add a definition of the new withheld code. Now added.

data/event-schemas/schema/m.room_key.yaml
```diff
@@ -25,6 +25,19 @@ properties:
       session_key:
         type: string
         description: The key to be exchanged.
+      shared_history:
+        x-addedInMatrixVersion: "1.19"
+        type: boolean
+        description: |
+          `true` indicates that the creator of this encryption considers that
```
uhoreg:
```suggestion
          `true` indicates that the creator of this encryption key considers that
```

?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1547,6 +1547,143 @@ objects described as follows:
...
+Alice does this by constructing an [encrypted key
+bundle](#construction-and-sharing-of-the-key-bundle) and sharing it with Bob
+before inviting him.
+
+##### Shareable encryption sessions
+
+Only room keys which were marked as "shareable" by the creator of the
+encryption session should be shared with new users.
+
+When Alice wants to send a message in the room she shares with Bob, she first
```
uhoreg: I found it mildly confusing that three paragraphs above, Bob is the person that Alice is inviting to the room, but here, Bob is already in the room.  Maybe change Bob here (and three lines below) to Carol?  Or maybe omit the name?

uhoreg (approved):
a couple nits, but otherwise lgtm

zecakeh (approved):

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1547,6 +1547,143 @@ objects described as follows:
...
+checks the [history visibility](#room-history-visibility) state of the
+room. If it is `shared` or `world_readable`, then when she sends the Megolm
+keys to Bob via an [`m.room_key`](#mroom_key) message, she sets
+`shared_history` to `true`.
```
```suggestion
When a user wants to send a message in an encrypted room, their client first
checks the [history visibility](#room-history-visibility) state of the
room. If it is `shared` or `world_readable`, then when the client sends the Megolm
keys to room members via [`m.room_key`](#mroom_key) messages, it SHOULD set
`shared_history` to `true`. Otherwise, it SHOULD set `shared_history` to `false`.
```

content/client-server-api/modules/end_to_end_encryption.md
richvdh: Yeah, I can rephrase this without a specific name and it will be clearer. Thanks.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1547,6 +1547,143 @@ objects described as follows:
 
 {{% definition path="api/client-server/definitions/megolm_export_session_data" %}}
 
+#### Sharing keys between users
+
+{{% added-in v="1.19" %}}
+
+When Alice invites Bob to an encrypted room, she might want Bob to have access
+to messages that were previously sent in that room, subject to the [history
+visibility](#room-history-visibility) setting of the room.
```
HarHarLinks, line 1556: @richvdh why would this not be at least SHOULD behavior? Clearly the decision of whether history is supposed to be shared in a room is supposed to be controlled by the history visibility setting?

turt2live: I suspect it's because it's dealing with an example. It'd be strange to say "You SHOULD want to share history".

HarHarLinks: Not "SHOULD want to", but "You SHOULD share historical keys".

The later paragraph

> Alice’s client MAY choose not to share any room history (even messages sent when the history visibity setting would allow sharing) if the current history visibility setting does not allow sharing (i.e. if history_visibility is set to invited or joined).

is written from the opposite perspective: "MAY withhold" IF history is not visible.

Why would sharing keys to messages that SHOULD NOT be visible to another user be allowed?

Regardless, the "otherwise" following it seems to imply a MUST share historical keys (i.e. "MAY NOT withhold"), hence I drafted https://github.com/matrix-org/matrix-spec/pull/2410.

turt2live: The later sentence is referencing a client, not an example person. Clarity is always welcome.

PR #1542, opened 2023-05-25
Update the changelog entry checker script to check newsfragment file extension

anoadragon453:
This PR updates the script that the newsfragment check CI step uses in order to verify that changelog entries have one of the [known newsfragment file extensions](https://github.com/matrix-org/matrix-spec/blob/main/CONTRIBUTING.rst#adding-to-the-changelog).

Inspired by failing to write a valid extension in https://github.com/matrix-org/matrix-spec/pull/1465 :)

```
matrix-spec on  appservice-ping-spec [$!?] took 5.7s …
➜ ./scripts/check-newsfragments 
+++ Checking newsfragments for PR #
ERROR: newsfragment changelogs/client_server/newsfragments/1465.clarifications does not have one of the required file extensions for a changelog entry: .new, .feature, .clarification, .breaking, .deprecation

matrix-spec on  appservice-ping-spec [$!?] …
➜ mv changelogs/client_server/newsfragments/1465.clarification{s,}

matrix-spec on  appservice-ping-spec [$✘!?] …
➜ ./scripts/check-newsfragments                                   
+++ Checking newsfragments for PR #
```

Preview: https://pr1542--matrix-spec-previews.netlify.app

PR #2402, opened 2026-06-16
Fix extension check in check-newsfragments

richvdh:
This was added in #1542, but has never worked correctly: it failed to set the exit code if an incorrectly-named newsfragment was found.

Preview: https://pr2402--matrix-spec-previews.netlify.app


turt2live (approved):

changelogs/internal/newsfragments/2402.clarification
turt2live, line 1: this should be 2402.clarification?

richvdh: Apparently 🤦

PR #2404, opened 2026-06-22
Spec for MSC4402: Consistent redirects for .well-known-files

derhagen:
### Pull Request Checklist

Spec writing for https://github.com/matrix-org/matrix-spec-proposals/pull/4402

Signed-off-by: Hagen Echzell <hagene@uio.no>

Preview: https://pr2404--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -429,7 +429,7 @@ Instead, they can be reached via HTTPS on the [server name](/appendices/#server-
 
 Servers hosting the `.well-known` JSON file SHOULD offer CORS headers,
 as per the [CORS](#web-browser-clients) section in this specification.
-Servers SHOULD also ensure that each 30x redirect, if any, offers such CORS headers.
+{{% added-in v="1.19" %}} Servers SHOULD also ensure that each 30x redirect, if any, offers such CORS headers.
 {{% /boxes/note %}}
 
 The flow for auto-discovery is as follows:
@@ -437,8 +437,8 @@ The flow for auto-discovery is as follows:
 1.  Extract the [server name](/appendices/#server-name) from the user's Matrix ID by splitting the
     Matrix ID at the first colon.
 2.  Extract the hostname from the server name as described by the [grammar](/appendices/#server-name).
-3.  Make a GET request to `https://hostname/.well-known/matrix/client`. 30x redirects should be followed,
-    however redirection loops should be avoided.
+3.  Make a GET request to `https://hostname/.well-known/matrix/client`.
+    {{% added-in v="1.19" %}} 30x redirects should be followed, however redirection loops should be avoided.
     1.  If the returned status code is 404, then `IGNORE`.
     2.  If the returned status code is not 200, or the response body is
         empty, then `FAIL_PROMPT`.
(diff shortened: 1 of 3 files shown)


content/client-server-api/_index.md
```diff
@@ -429,14 +429,16 @@ Instead, they can be reached via HTTPS on the [server name](/appendices/#server-
...
 {{% /boxes/note %}}
 
 The flow for auto-discovery is as follows:
 
 1.  Extract the [server name](/appendices/#server-name) from the user's Matrix ID by splitting the
     Matrix ID at the first colon.
 2.  Extract the hostname from the server name as described by the [grammar](/appendices/#server-name).
-3.  Make a GET request to `https://hostname/.well-known/matrix/client`.
+3.  Make a GET request to `https://hostname/.well-known/matrix/client`. 30x redirects should be followed,
+    however redirection loops should be avoided.
```
Johennes: I know the server-server API uses the same wording but should this be normative? In other words, is "should" here a "SHOULD" or a "MUST"? Same question for the usage of "should" in the YAML file.

derhagen: I'd propose to keep this consistent with the server-server API wording for the scope of this PR to get it done as soon as possible.

Do you think changing from SHOULD to MUST requires a separate MSC? If so, I think we should open one that changes both S-S API as well as C-S API, as I think MUST makes more sense.

If there is no MSC required, I still think this PR should be merged first, and the wording adjusted for both APIs at the same time, afterwards.

Johennes: I think it would only require an MSC if the choice is controversial. MSCs are technically not required to use the RFC 2119 keywords. The translation often only happens when writing the spec PR.

Are there any reasons why a server might not want to follow redirects (other than when it detected a loop)?

derhagen: I don't see any reason to not follow redirects, except for loop detection effort. One could also limit the number of redirects to some small integer in order to simplify the required logic and avoid being able to DoS clients/servers with infinite redirection chains.

Yet, I think we should get this PR merged first, and change C-S and S-S wording subsequently, if discussion shows this to be uncontroversial.

uhoreg: Changing the "should" to a "SHOULD" wouldn't require an MSC.  Most "should"s currently in the spec should probably be "SHOULD"s.

Changing to a "MUST" would require an MSC.

content/client-server-api/_index.md
```diff
@@ -429,14 +429,16 @@ Instead, they can be reached via HTTPS on the [server name](/appendices/#server-
...
+Servers SHOULD also ensure that each 30x redirect, if any, offers such CORS headers.
 {{% /boxes/note %}}
 
 The flow for auto-discovery is as follows:
 
 1.  Extract the [server name](/appendices/#server-name) from the user's Matrix ID by splitting the
     Matrix ID at the first colon.
 2.  Extract the hostname from the server name as described by the [grammar](/appendices/#server-name).
-3.  Make a GET request to `https://hostname/.well-known/matrix/client`.
+3.  Make a GET request to `https://hostname/.well-known/matrix/client`. 30x redirects should be followed,
```
uhoreg: Changes should be noted, indicating the version of the spec that the changes were made in


```suggestion
3.  Make a GET request to `https://hostname/.well-known/matrix/client`. {{% added-in v="1.19" %}} 30x redirects should be followed,
```

(similarly with the yaml file)

uhoreg (approved):
Thanks!

PR #2407, opened 2026-07-09
Add further normative language in mutual rooms server behaviour

Johennes:
This adds further RFC2119 language to the server behaviour section of the mutual rooms endpoint. The "should" in the current wording struck me as ambiguous.

Either it is a requirement to include the pagination token, then we should use "MUST". Or it is recommended but not required (meaning the server can just truncate). Then we should use "SHOULD".

I assume we want a MUST but https://github.com/matrix-org/matrix-spec-proposals/pull/2666 doesn't seem to make this explicit (maybe @Half-Shot could clarify the intention as the author).

### Pull Request Checklist

Preview: https://pr2407--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/mutual_rooms.md
+++ b/content/client-server-api/modules/mutual_rooms.md
@@ -7,7 +7,7 @@
 #### Server behaviour
 
 The server MAY decide that the response to this endpoint is too large, and only return a
-subset of the results. In this case, the server MUST populate the optional field `next_batch`
+subset of the results. In this case, the server populates the optional field `next_batch`
 with an [opaque identifier](/appendices/#opaque-identifiers). The client can then supply
 the identifier as the `from` query parameter in a subsequent request, along with the original
 `user_id`, to fetch the next batch of responses. This will continue until the server no longer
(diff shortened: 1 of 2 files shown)


content/client-server-api/modules/mutual_rooms.md
```diff
@@ -6,9 +6,9 @@
 
 #### Server behaviour
 
-The server may decide that the response to this endpoint is too large, and only return a
-subset of the results. In this case, the server should populate the optional field `next_batch`
-with an [opaque identifier](/appendices/#opaque-identifiers). The client may then supply
+The server MAY decide that the response to this endpoint is too large, and only return a
+subset of the results. In this case, the server MUST populate the optional field `next_batch`
```
uhoreg: I don't think we need the RFC keyword here, since it's talking about what happens if the server decides to do something, but I agree that the original "should" is a bit confusing.

```suggestion
subset of the results. In this case, the server populates the optional field `next_batch`
```

Johennes: Sounds good, yes.

uhoreg (approved):
Thanks!

PR #2410, opened 2026-07-10
Clarify historical key sharing

HarHarLinks:
See https://github.com/matrix-org/matrix-spec/pull/2399/files#r3560365639

We understand there is an introductory paragraph to the use case, but we think it is setting the rest of the section up for confusion given
- that there is no clear requirement level for historical key sharing in general stated (only different aspects about how to create the key bundles etc)
- how fuzzy the spec is in general towards proper RFC 2119 usage.

Our reading of the spec is that for some reason there is an exemption for rooms that are currently not using shareable history, but by default ("otherwise") shareable keys hence may not not [sic!] be shared, i.e. MUST be shared. We think stating that outright clarifies this requirement a lot.

Further the way the exemption was defined was hard to grasp for a group of us, hence an attempt of a slightly more verbose approach.

Finally the error codes for withholding keys were missing a reference to when to choose which one.

### Pull Request Checklist

Preview: https://pr2410--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1551,7 +1551,7 @@ objects described as follows:
 
 {{% added-in v="1.19" %}}
 
-When Alice invites Bob to an encrypted room, she might want Bob to have access
+When Alice invites Bob to an encrypted room, she likely wants Bob to have access
 to messages that were previously sent in that room, subject to the [history
 visibility](#room-history-visibility) setting of the room.
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1551,9 +1551,12 @@ objects described as follows:
 
 {{% added-in v="1.19" %}}
 
-When Alice invites Bob to an encrypted room, she might want Bob to have access
-to messages that were previously sent in that room, subject to the [history
-visibility](#room-history-visibility) setting of the room.
+When Alice invites Bob to an encrypted room, that room might be set up for
+Bob to have access to messages that were previously sent in that room,
+subject to the [history visibility](#room-history-visibility) setting of the room.
+Because the key material is not available on the server,
+Alice MUST share the decryption keys of messages visible according to
+the history visibility setting so Bob can read the message history.
 
 Alice does this by constructing an [encrypted key
 bundle](#construction-and-sharing-of-the-key-bundle) and sharing it with Bob
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1623,9 +1623,8 @@ In such cases where the current history visibility setting does not allow sharin
 MAY choose not to share *any* room history, even messages sent when the
 history visibility setting would allow sharing.
 
-In all other cases, Alice's client SHOULD share all available shareable keys.
-
-Before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
+In all other cases, before inviting Bob to a room, Alice's client SHOULD share all available shareable keys.
+It does this by constructing and sending a key bundle as follows:
 
 1. Alice's client SHOULD ensure that it has downloaded all keys relevant to the room
    from [server-side key backup](#server-side-key-backups), if she is using it.
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1614,12 +1614,12 @@ Marking keys as [shareable](#shareable-encryption-sessions) essentially serves a
 precomputation of which keys Bob needs to decrypt all messages he can see, as
 defined by the room's history visibility.
 
-History visibility mechanics prevent changing the visibility of events in
-hindsight. It is thus possible for a room timeline to have "gappy" visibility,
-i.e. events sent during `shared` visibility in the past are visible, but events
-more recently sent during `joined` visibility and before joining are not visible.
+History visibility mechanics prevent changing the visibility of events
+retrospectively. It is thus possible for a room timeline to have "gappy" visibility.
+For example, events sent in the past during `shared` visibility are visible to new
+members, even though events sent more recently (during `joined` visibility) are not visible.
 In such cases where the current history visibility setting does not allow sharing
-(i.e. if `history_visibility` is set to `invited` or `joined`), Alice's client
+(i.e. if `history_visibility` is set to `invited` or `joined`), client implementations
 MAY choose not to share *any* room history, even messages sent when the
 history visibility setting would allow sharing.
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1623,7 +1623,7 @@ In such cases where the current history visibility setting does not allow sharin
 MAY choose not to share *any* room history, even messages sent when the
 history visibility setting would allow sharing.
 
-In all other cases, Alice's client MUST share all available shareable keys.
+In all other cases, Alice's client SHOULD share all available shareable keys.
 
 Before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1551,12 +1551,9 @@ objects described as follows:
 
 {{% added-in v="1.19" %}}
 
-When Alice invites Bob to an encrypted room, that room might be set up for
-Bob to have access to messages that were previously sent in that room,
-subject to the [history visibility](#room-history-visibility) setting of the room.
-Because the key material is not available on the server,
-Alice MUST share the decryption keys of messages visible according to
-the history visibility setting so Bob can read the message history.
+When Alice invites Bob to an encrypted room, she might want Bob to have access
+to messages that were previously sent in that room, subject to the [history
+visibility](#room-history-visibility) setting of the room.
 
 Alice does this by constructing an [encrypted key
 bundle](#construction-and-sharing-of-the-key-bundle) and sharing it with Bob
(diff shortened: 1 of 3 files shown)


HarHarLinks:
- [x] normative text explaining `m.blacklisted, m.unverified, m.unauthorised, m.unavailable, m.no_olm, m.history_not_shared`
- [x] rework the paragraph "Alice’s client MAY choose not to share ..."

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1610,12 +1610,22 @@ room.
...
+i.e. events sent during `shared` visibility in the past are visible, but events
+more recently sent during `joined` visibility and before joining are not visible.
+In such cases where the current history visibility setting does not allow sharing
+(i.e. if `history_visibility` is set to `invited` or `joined`), Alice's client
+MAY choose not to share *any* room history, even messages sent when the
+history visibility setting would allow sharing.
+
+In all other cases, Alice's client MUST share all available shareable keys.
+
+Before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
```
richvdh: I agree this needs clarifying; however, I don't think I agree with this particular interpretation, especially

> In all other cases, Alice's client MUST share all available shareable keys.

This basically means that clients MUST implement history sharing. That wasn't really the intention when I specced this: I was aiming rather more for a SHOULD semantic.

@erikjohnston's take on this was:

 * a) a MUST is a bit pointless [as clients devs aren't going to implement it if they don't want to] 
 * b) there doesn't seem to be a reason for it to be MUST? Clients shouldn't be forced to share keys if they don't want to, and I think there are valid use cases for not

---

> Our reading of the spec is that for some reason there is an exemption for rooms that are currently not using shareable history, but by default ("otherwise") shareable keys hence may not not [sic!] be shared, i.e. MUST be shared. 

Yeah, that wasn't the intention of that clause. Rather, I was aiming for something closer to "even if you *usually* do history sharing, it might be a good idea *not* to do it in this case". 

As you noted elsewhere, this isn't discussed in the MSC. We found it beneficial during our implementation (there was some discussion in https://github.com/element-hq/element-meta/issues/3101). If you interpret the MSC as "clients can do this when they see fit, and not otherwise", then that fits fine. It seemed helpful to try and share that insight with future client developers, hence adding the MAY clause.

---

In short: can we reword this to use SHOULD semantics rather than MUST?

HarHarLinks: > a MUST is a bit pointless [as clients devs aren't going to implement it if they don't want to]

isn't this always the case though? SHOULD is allowing implementers to diverge from a recommendation if they have "really good reasons to", while MUST asserts that otherwise they are not compliant with the spec, i.e. breaking expectations (-> see below).

> valid use cases for not

assuming this doesn't mean the MAY exception that is still in this PR as of now, i would like to hear some?

> I was aiming rather more for a SHOULD semantic

why?

what i think is going to happen here by not strongly asserting a single behaviour, is a divergence of the configured history visibility everyone already agrees upon(!) from clients doing "whatever" anyway. essentially, this is provoking UTD situations when history visibility says one thing and e2ee history sharing allows another. this means participants in a room will have different experiences based on not just who invites but with what client implementation. even with the (unspecced? hence not to be expected across the ecosystem) behaviour of element (iirc) to hide undecryptable historical messages as "no access to previous messages", there will be inconsistent behaviour for what changing the history visibility over the course of history results in between e2ee and unencrypted rooms (cf. https://github.com/element-hq/element-meta/issues/3101).

speaking of https://github.com/element-hq/element-meta/issues/3101, the same is true for the MAY clause creating that exception. with my current reading of the spec as MUST, this creates the only exception. if process allows I would move for removal of that exception from the spec entirely through this clarification, and discuss in an MSC whether to not add it and keep in sync with history visibility, instead adjust history visibility to achieve this bevahiour consistently also with unencrypted rooms. However I suspect it would need an MSC to remove at this point since it was somehow released as spec 1.19 without being agreed upon in an MSC, in which case I would MSC to adjust the history visibility in general to match this behaviour and get rid of this exception.

I would appreciate a pointer from the SCT here about how to proceed.

---

In short:

1. Maybe the SCT needs to agree if they interpreted MUST or SHOULD during MSC review?
  I can then depending on the outcome MSC for a MUST if I feel like there were not enough arguments against.
2. With the SCT's guidance about which route to take, I would MSC to unify history visibility behaviour with https://github.com/element-hq/element-meta/issues/3101 in a way that makes a stronger recommendation about the behaviour than "MAY" to achieve a more consistent behaviour across Matrix.

richvdh: Kim, that's quite a wall of text. Even I'm struggling to parse it, and I have most of the context here. I don't think it's fair to ask 9 people to jump into it and read and understand it: even if they were inclined to do so, I don't think it's the best use of their time.

You've had my review, and I really don't think you're going to get much disagreement from the rest of the SCT. I'd suggest you accept that it's a SHOULD for now, and if you feel strongly that it needs to be a MUST, make that case in a separate MSC.

HarHarLinks: Sorry, I am not trying to argue for argument's sake, and perhaps I didn't understand your politely phrased suggestion as a decision. I'm asking or arguing to learn more because I would have concluded something else which I tried to explain by my arguments. I still think the arguments provided in favour of SHOULD are rather thin. What I failed to write explicitly is that I will of course adjust this PR if it is the decision of the SCT which you have made clear now is the one you gave.

Lacking feedback about my other process question, I accept leaving the MAY exception in for the purpose of this clarification as well.

I suggest the SCT take note of this case and include it in a future review of its process towards preventing significant deviations from MSCs proposals slip into the spec in spec MSCs as has happened here, e.g. by taking responsibility for removing them when noticed.

richvdh: > significant deviations from MSCs proposals slip into the spec in spec MSCs as has happened here

I think that's a matter of opinion.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1610,12 +1610,22 @@ room.
 
 ##### Construction and sharing of the key bundle
 
-Alice's client MAY choose not to share any room history (even messages sent when the
-history visibity setting would allow sharing) if the current history
-visibility setting does not allow sharing (i.e. if `history_visibility` is
-set to `invited` or `joined`).
+Marking keys as [shareable](#shareable-encryption-sessions) essentially serves as
+precomputation of which keys Bob needs to decrypt all messages he can see, as
+defined by the room's history visibility.
```
richvdh, line 1615: I'm not really sure this is a helpful way of thinking about things? When Charlie sends a message in a room and marks it as shareable, he may not even know who Bob is, and no computation seems to be required. I really don't think that Charlie is "precomputing" anything?

[Also, it's a long time since Bob was mentioned, so we need to reintroduce him here. Admittedly that's a problem to some extent with the existing text, but at least we have Alice inviting Bob in the next paragraph at the moment.]

I suggest removing this paragraph.

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1610,12 +1610,22 @@ room.
...
 
-Otherwise, before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
+History visibility mechanics prevent changing the visibility of events in
+hindsight. It is thus possible for a room timeline to have "gappy" visibility,
+i.e. events sent during `shared` visibility in the past are visible, but events
+more recently sent during `joined` visibility and before joining are not visible.
```
```suggestion
History visibility mechanics prevent changing the visibility of events
retrospectively. It is thus possible for a room timeline to have "gappy" visibility.
For example, events sent in the past during `shared` visibility are visible to new
members, even though events sent more recently (during `joined` visibility) are not visible.
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1610,12 +1610,22 @@ room.
...
+In such cases where the current history visibility setting does not allow sharing
+(i.e. if `history_visibility` is set to `invited` or `joined`), Alice's client
```
```suggestion
(i.e. if `history_visibility` is set to `invited` or `joined`), client implementations
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1610,12 +1610,22 @@ room.
...
+MAY choose not to share *any* room history, even messages sent when the
+history visibility setting would allow sharing.
+
+In all other cases, Alice's client SHOULD share all available shareable keys.
+
+Before inviting Bob to a room, Alice's client constructs and sends a key bundle as follows:
```
```suggestion
In all other cases, before inviting Bob to a room, Alice's client SHOULD share all available shareable keys.
It does this constructs and sends a key bundle as follows:
```

richvdh (changes requested):

content/client-server-api/modules/end_to_end_encryption.md
HarHarLinks: brain fails to parse the the last sentence, perhaps:
```suggestion
In all other cases, before inviting Bob to a room, Alice's client SHOULD share all available shareable keys.
It does this by constructing and sending a key bundle as follows:
```

HarHarLinks:
https://github.com/matrix-org/matrix-spec/blob/bb623a587e2512d9d13b11812740b84590c3eec6/content/client-server-api/modules/end_to_end_encryption.md?plain=1#L1564-L1565

Would you agree this should instead say

```
 Only room keys which were marked as "shareable" by the creator of the 
 encryption session MUST be shared with new users. 
```

(or less normative `must`)

given that further down defines

https://github.com/matrix-org/matrix-spec/blob/bb623a587e2512d9d13b11812740b84590c3eec6/content/client-server-api/modules/end_to_end_encryption.md?plain=1#L1634-L1635

content/client-server-api/modules/end_to_end_encryption.md
richvdh: sorry, yes

richvdh:
>  Only room keys which were marked as "shareable" by the creator of the encryption session MUST be shared with new users. 

That materially changes the emphasis: it implies that keys *not* marked as shareable can (and indeed probably should) be shared, which is obviously not what we mean.

If you wanted to strengthen this, it would need to say:

```
Room keys which were not marked as "shareable" by the creator of the 
encryption session MUST NOT be shared with new users. 
```

... however, that's an odd thing to say *before* describing sharing keys in general. 

I think the current text is adequate, given (as you note) it's described more formally later on.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: bump

PR #2413, opened 2026-07-13
improve event format description of m.room.encrypted events

gewitternacht:
The current description for the format of the content of `m.room.encrypted` Olm/Megolm events (see https://spec.matrix.org/v1.19/client-server-api/#molmv1curve25519-aes-sha2 and https://spec.matrix.org/v1.19/client-server-api/#mmegolmv1aes-sha2) is rather confusing, which I'm trying to improve by fixing the following issues (as discussed with @richvdh and @andybalaam):
- we can only really define the format of `content`, any properties like `type` alongside `content` are already fixed by how events look like (which is why they are now mentioned in the accompanying text)
- importantly, events also have a `sender` property alongside `content`, which wasn't present in the JSON snippet before, but is referenced further below
- the old placeholders for `body` contained "payload", which is a very overloaded term
- the description of `ciphertext` wasn't really correct
- for Megolm, a reference to the message format was missing

### Pull Request Checklist

Preview: https://pr2413--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1742,11 +1742,11 @@ The content of an event encrypted using Olm has the following format:
 Note that when the event is received from the server, it will have a `type`
 (with a value of `m.room.encrypted`) and `sender` property alongside the
 `content` property. In `content`, `ciphertext` is a mapping from a device
-Curve25519 key to an object with a `type` and a `body`. Here, `type` is an
-integer indicating the type of the message body:
+Curve25519 key to an object with a `type` and a `body`. Here, `body` is a
+Base64-encoded [Olm message](/olm-megolm/olm/#the-olm-message-format), and
+`type` is an integer indicating the type of the message:
 0 for the initial [pre-key message](/olm-megolm/olm/#pre-key-messages),
-1 for [normal messages](/olm-megolm/olm/#normal-messages);
-`body` is a Base64-encoded [Olm message body](/olm-megolm/olm/#the-olm-message-format).
+1 for [normal messages](/olm-megolm/olm/#normal-messages).
 
 Olm sessions will generate messages with a type of 0 until they receive
 a message. Once a session has decrypted a message it will produce
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1911,7 +1911,9 @@ The content of an [`m.room.encrypted`](#mroomencrypted) event using Megolm has t
 ```
 Note that when the event is received from the server, it will have additional
 properties alongside the `content` property, including a `type` (with a value
-of `m.room.encrypted`) and a `sender` property. In `content`, `ciphertext`
+of `m.room.encrypted`) and a `sender` property. (See [Room event format](#room-event-format).)
+
+Within `content`, `ciphertext`
 is a Base64-encoded [Megolm message](/olm-megolm/megolm/#message-format),
 whose plaintext body is of the form:
 
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1745,7 +1745,7 @@ Note that when the event is received from the server, it will have a `type`
 Curve25519 key to an object with a `type` and a `body`. Here, `body` is a
 Base64-encoded [Olm message](/olm-megolm/olm/#the-olm-message-format), and
 `type` is an integer indicating the type of the message:
-0 for the initial [pre-key message](/olm-megolm/olm/#pre-key-messages),
+0 for the initial [pre-key messages](/olm-megolm/olm/#pre-key-messages),
 1 for [normal messages](/olm-megolm/olm/#normal-messages).
 
 Olm sessions will generate messages with a type of 0 until they receive
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1723,28 +1723,30 @@ Devices that support Olm must include "m.olm.v1.curve25519-aes-sha2" in
 their list of supported messaging algorithms, must list a Curve25519
 device key, and must publish Curve25519 one-time keys.
 
-An event encrypted using Olm has the following format:
+The content of an event encrypted using Olm has the following format:
 
 ```json
 {
-  "type": "m.room.encrypted",
   "content": {
     "algorithm": "m.olm.v1.curve25519-aes-sha2",
     "sender_key": "<sender_curve25519_key>",
     "ciphertext": {
       "<device_curve25519_key>": {
         "type": 0,
-        "body": "<encrypted_payload_base_64>"
+        "body": "<base64_encoded_olm_message>"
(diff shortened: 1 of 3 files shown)


content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1742,11 +1742,11 @@ The content of an event encrypted using Olm has the following format:
 Note that when the event is received from the server, it will have a `type`
```
andybalaam, line 1742: The commit description appears to be missing a word. Did you do what I always do and use backticks in a `git commit -m` command?

andybalaam (approved):
Looks good to me, with one comment.

content/client-server-api/modules/end_to_end_encryption.md
gewitternacht: Oh... yes I did. It was supposed to say "reorder explanation of `body` and `type`, avoid 'message body'".
Anything I can or should do about that now?

tulir: All PRs get squash merged, so the individual commits inside aren't important. The PR title is used as the final commit message

richvdh:
(I've added some links to the existing spec to the PR description, for my own reference, and the benefit for future archaeologists)

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1723,28 +1723,30 @@ Devices that support Olm must include "m.olm.v1.curve25519-aes-sha2" in
 their list of supported messaging algorithms, must list a Curve25519
 device key, and must publish Curve25519 one-time keys.
 
-An event encrypted using Olm has the following format:
+The content of an event encrypted using Olm has the following format:
```
richvdh: How about something like:


```suggestion
The content of an [`m.room.encrypted`](#mroomencrypted) event using Olm has the following format:
```

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1894,23 +1896,24 @@ This uses:
 Devices that support Megolm must support Olm, and include
 "m.megolm.v1.aes-sha2" in their list of supported messaging algorithms.
 
-An event encrypted using Megolm has the following format:
+The content of an event encrypted using Megolm has the following format:
```
richvdh: again, link to the `m.room.encrypted` event type definition?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1723,28 +1723,30 @@ Devices that support Olm must include "m.olm.v1.curve25519-aes-sha2" in
...
 }
 ```
-
-`ciphertext` is a mapping from device Curve25519 key to an encrypted
-payload for that device. `body` is a Base64-encoded [Olm message body](/olm-megolm/olm/#the-olm-message-format).
-`type` is an integer indicating the type of the message body: 0 for the
-initial [pre-key message](/olm-megolm/olm/#pre-key-messages), 1 for [normal messages](/olm-megolm/olm/#normal-messages).
+Note that when the event is received from the server, it will have a `type`
+(with a value of `m.room.encrypted`) and `sender` property alongside the
+`content` property. In `content`, `ciphertext` is a mapping from a device
```
richvdh, line 1744:
```suggestion
`content` property. (See [Room event format](#room-event-format).)

Within `content`, `ciphertext` is a mapping from a device
```

maybe?

content/client-server-api/modules/end_to_end_encryption.md
```diff
@@ -1894,23 +1896,24 @@ This uses:
...
+    "ciphertext": "<base64_encoded_megolm_message>"
   }
 }
 ```
-
-The encrypted payload can contain any message event. The plaintext is of
-the form:
+Note that when the event is received from the server, it will have additional
+properties alongside the `content` property, including a `type` (with a value
+of `m.room.encrypted`) and a `sender` property. In `content`, `ciphertext`
```
```suggestion
of `m.room.encrypted`) and a `sender` property. (See [Room event format](#room-event-format).)

Within `content`, `ciphertext`
```

maybe?

content/client-server-api/modules/event_replacements.md
```diff
@@ -91,13 +91,13 @@ For example, a replacement for an encrypted event might look like this:
...
         "session_id": "<outbound_group_session_id>",
-        "ciphertext": "<encrypted_payload_base_64>"
+        "ciphertext": "<base64_encoded_megolm_message>"
     }
     // irrelevant fields not shown
 }
 ```
 
-... and, once decrypted, the payload might look like this:
+... and, once decrypted, the event might look like this:
```
richvdh: Whilst in practice there might be a 1:1 correspondence between the plaintext of the megolm message and the event it represents, it might not be obvious that that's the case, and I think it's important to note that the data below is the former, not the latter. Can we find some other words to note that this is what is actually encrypted via Megolm?

```suggestion
... and the plaintext body of the Megolm message might look like this:
```

maybe?

richvdh (commented):
Thanks for this! A couple of comments.

content/client-server-api/modules/end_to_end_encryption.md
gewitternacht: That's a very good idea! Applied in a692b7be8392a21477e269eb06f490f119a91bc3.

content/client-server-api/modules/end_to_end_encryption.md
gewitternacht: Done in a692b7be8392a21477e269eb06f490f119a91bc3.

content/client-server-api/modules/end_to_end_encryption.md
gewitternacht: For reference, this links to the [ClientEvent](https://spec.matrix.org/latest/client-server-api/#room-event-format) format. I'm not sure: Does that format also apply to to-device messages, which to my understanding are the main (or in practice, only) usage of Olm-encrypted messages?

content/client-server-api/modules/end_to_end_encryption.md
gewitternacht: Similarly to above: will Megolm-encrypted events always use the [ClientEvent](https://spec.matrix.org/latest/client-server-api/#room-event-format) format, or could there be instances where the properties alongside `content` are different ones?

content/client-server-api/modules/event_replacements.md
gewitternacht: I can't think of anything else that sounds good, so will apply your suggestion 👍

content/client-server-api/modules/end_to_end_encryption.md
richvdh: Er, no. That doesn't apply to to-device events. I think I put this comment in the wrong place, and then failed to delete it when I moved it further down. Sorry for the noise.

content/client-server-api/modules/end_to_end_encryption.md
richvdh: I'm not sure if the spec makes it explicit, but Megolm is only designed for use on room events, so in this case, yes such events *will* always use the ClientEvent format.

richvdh (approved):
Thanks!

PR #2414, opened 2026-07-13
rename payload -> body in Olm and Megolm specification

gewitternacht:
The term "payload" is (very confusingly) used for two different things in regards to Olm/Megolm:
1. The [`OlmPayload`](https://spec.matrix.org/v1.19/client-server-api/#molmv1curve25519-aes-sha2) struct in the Client-Server API – which is the plaintext corresponding to the "Cipher-Text" that is part of ...
2. the [`Payload Bytes`](https://spec.matrix.org/v1.19/olm-megolm/olm/#normal-messages) in the Olm specification.

To reduce confusion, I suggest renaming the `Payload` in the Olm and Megolm specification to `Body`.

### Pull Request Checklist

Preview: https://pr2414--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2414.clarification
+++ b/changelogs/client_server/newsfragments/2414.clarification
@@ -1,0 +1,1 @@
+Rename `Payload` in the Olm/Megolm specification to `Body` to avoid confusion with the distict `OlmPayload`.
--- a/changelogs/client_server/newsfragments/2414.clarification
+++ b/changelogs/client_server/newsfragments/2414.clarification
@@ -1,1 +1,1 @@
-Rename `Payload` in the Olm/Megolm specification to `Body` to avoid confusion with the distict `OlmPayload`.
+Rename `Payload` in the Olm/Megolm specification to `Body` to avoid confusion with the distinct `OlmPayload`.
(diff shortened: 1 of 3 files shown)


turt2live:
Wouldn't this then conflict with `body` that we use in `m.room.message` events?

gewitternacht:
Yes, and I'm also remembering now that there's a `body` in the [`m.room.encrypted`](https://spec.matrix.org/v1.19/client-server-api/#molmv1curve25519-aes-sha2) envelope of Olm messages. I think our reasoning when discussing this with @richvdh and @andybalaam was that `body` is a way more general/usual term that you would expect to be used for a bunch of different things. But yeah, I agree, probably not the best renaming after all.

Maybe there's a third alternative that's better suited (but I can't think of any right now)?

An alternative suggestion of Rich was to rename `OlmPayload` to `OlmPlaintext` – however, I'm not sure how easily a "struct" like this can be renamed, would you say that's a possibility?

turt2live:
The struct names are largely invented, so can typically be renamed without major issue.

gewitternacht:
There's now https://github.com/matrix-org/matrix-spec/pull/2418 as an alternative, which I'd actually prefer now.

gewitternacht:
https://github.com/matrix-org/matrix-spec/pull/2418 was merged, so I'm closing this.

PR #2418, opened 2026-07-15
Rename OlmPayload -> OlmPlaintext to avoid confusion with Olm message Payload Bytes

gewitternacht:
Alternative to https://github.com/matrix-org/matrix-spec/pull/2414. As explained there:

The term "payload" is (very confusingly) used for two different things in regards to Olm/Megolm:
1. The [`OlmPayload`](https://spec.matrix.org/v1.19/client-server-api/#definition-olmpayload) struct in the Client-Server API – which is the plaintext corresponding to the "Cipher-Text" that is part of ...
2. the [`Payload Bytes`](https://spec.matrix.org/v1.19/olm-megolm/olm/#normal-messages) in the Olm specification.

To reduce confusion, I suggest renaming the `OlmPayload` to `OlmPlaintext`.

I tried to catch any references to the term "payload" in this context, but note that there's also https://github.com/matrix-org/matrix-spec/pull/2413 which already removes some occurrences.

I also renamed the `olm_payload.yaml` file to `olm_plaintext.yaml`, is that okay/desired, or might that break something (which I might not have seen when searching for "payload" throughout the spec)?

### Pull Request Checklist

Preview: https://pr2418--matrix-spec-previews.netlify.app

### DIFF
--- a/changelogs/client_server/newsfragments/2418.clarification
+++ b/changelogs/client_server/newsfragments/2418.clarification
@@ -1,0 +1,1 @@
+Rename OlmPayload to OlmPlaintext to avoid confusion with Olm message Payload Bytes.
--- a/content/client-server-api/modules/end_to_end_encryption.md
+++ b/content/client-server-api/modules/end_to_end_encryption.md
@@ -1762,12 +1762,9 @@ Messages with type 1 can only be decrypted with an existing session. If
 there is no matching session, the client must treat this as an invalid
 message.
 
-The plaintext payload is of the form:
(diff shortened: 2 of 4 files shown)


richvdh (approved):
Looks great, thanks!

richvdh:
> I also renamed the olm_payload.yaml file to olm_plaintext.yaml, is that okay/desired, or might that break something (which I might not have seen when searching for "payload" throughout the spec)?

It should be fine; the CI should catch anything we've forgotten.

PR #2427, opened 2026-07-31
Add `prev_sender` to client event spec

### LINKED ISSUE #2416: Spec `prev_sender` in CS API
Somehow we've never specified `unsigned.prev_sender` - a field on state events which tells you who the sender was of the state event that was replaced by the current state event.

As far as I can see, what happened is:
 * @richvdh first specced it in 2015 for CS events: 29bd4d45ee9155bd80e61ce50ab34745be1fd8a4 and 6653362f315d1379d1dda248dfd299c8c3c8530d
 * @turt2live then moved it to a SS-specific `unsigned` block in 2018: 5027a9a59a4ca257f6b71f78f6f0fe3811940400
 * @richvdh then removed it from the SS block in [#3522](https://github.com/matrix-org/matrix-spec-proposals/pull/3522)
 * `prev_content` got re-specced based in [#3524](https://github.com/matrix-org/matrix-spec-proposals/pull/3524) based on https://github.com/matrix-org/matrix-spec-proposals/issues/877
 * `replaces_state` eventually got re-specced a few months ago thanks to @timedoutuk in https://github.com/matrix-org/matrix-spec/pull/2345 closing https://github.com/matrix-org/matrix-spec/issues/274

...but somehow `prev_sender` got missed from all this (which confuses me greatly: isn't it in the same bucket as `replaces_state`?).

In practice, `prev_sender` is in use in the wild:
 * matrix-js-sdk uses it to determine DMedness of things
 * matrix-js-sdk also uses it to 'play state backwards' for sync v2
 * It's also useful for seeing who answered a knock
 * Synapse implements it
 * Continuwuity (picking an indie server at random) implements it too

Therefore, surely it should be specced as an omission?

timedoutuk:
Per https://github.com/matrix-org/matrix-spec/issues/2416, adds `prev_sender` to the client event `unsigned` object. Following the same rules as `replaces_state`.

### Pull Request Checklist

Preview: https://pr2427--matrix-spec-previews.netlify.app


richvdh:
Let the record show that I debated whether `x-addedInMatrixVersion` was correct, given this has been implemented since time immemorial, or whether the changelog could be clearer, but in the end: I think this is fine, and it matches what we did for #2345.

richvdh (approved):
Thanks!

richvdh:
Fixes: #2416

PR #2432, opened 2026-08-06
Link to OAuth RFCs for error response definitions

hughns:
The Client-Server [API standards](https://spec.matrix.org/v1.19/client-server-api/#api-standards) says at that they don't apply to OAuth 2.0 (and other unspecified) endpoints:

> These standards only apply to the APIs defined in the Matrix specification. APIs used by this specification but defined in other specifications, like the [OAuth 2.0 API](https://spec.matrix.org/v1.19/client-server-api/#oauth-20-api), use their own rules.

But this detail can be lost when you are reading about the OAuth 2.0 flows lower down on the page. Particularly in the case of errors where we don't always given examples of the error responses.

In this PR I have attempt to make the situation better by:

- adding explicit notes next to where OAuth 2.0 API responses are documented and link to where the error responses are actually defined
- clarifying within the "standard error responses" section that it doesn't apply to OAuth 2.0

### Pull Request Checklist

Preview: https://pr2432--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -56,8 +56,9 @@ requirements for server responses.
 
 ### Standard error response
 
-Any errors which occur at the Matrix API level MUST return a "standard
-error response". This is a JSON object which looks like:
+Any errors which occur at the Matrix API level (for requests that are within
+scope of these [API Standards](#api-standards)) MUST return a "standard error
+response". This is a JSON object which looks like:
 
 ```json
 {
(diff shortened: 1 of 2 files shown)


content/client-server-api/_index.md
```diff
@@ -56,8 +56,9 @@ requirements for server responses.
 
 ### Standard error response
 
-Any errors which occur at the Matrix API level MUST return a "standard
-error response". This is a JSON object which looks like:
+Any errors which occur at the Matrix API level (for requests that are within
```
hughns, line 59: This new wording feels inelegant but can hopefully be iterated on.

The intent was to make it clear that this statement is excepted higher up the page.

An alternative could be:

> Any errors which occur at the Matrix API level (e.g. not including the OAuth 2.0 API) MUST return a "standard error response".

richvdh (approved):
Looks good to me. Thanks.

PR #2429, opened 2026-07-31
Restore legacy well-known URI anchor

lntutor:
Fixes #2425.

Restores the singular `well-known-uri` anchor referenced by the IANA registration while preserving the current plural heading anchor.

### Pull Request Checklist

* [x] Pull request includes a changelog file
* [x] Pull request includes a sign off
* [x] Pull request is classified as "other changes"

PR #2433, opened 2026-08-11
Fix links broken in an earlier commit

richvdh:
Followup to #2429, which broke some existing links

Preview: https://pr2433--matrix-spec-previews.netlify.app

### DIFF
--- a/content/client-server-api/_index.md
+++ b/content/client-server-api/_index.md
@@ -416,7 +416,9 @@ valid data was obtained, but no server is available to serve the client.
 No further guess should be attempted and the user should make a
 conscientious decision what to do next.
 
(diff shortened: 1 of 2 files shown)


content/client-server-api/_index.md
```diff
@@ -416,7 +416,9 @@ valid data was obtained, but no server is available to serve the client.
 No further guess should be attempted and the user should make a
 conscientious decision what to do next.
 
-### Well-known URIs {#well-known-uri}
+### <a id="well-known-uri" /> Well-known URIs
```
richvdh, line 419: the purpose here is to end up with two anchors, which both point at the heading.

anoadragon453 (approved):

